PR #73, opened 2026-02-02
Sampleworks refactoring - separate guidance from sampling, and improve ModelWrapper protocols

k-chrispens:
This is a huge PR, apologies in advance.

Main things to note:

1. We now have 3 ModelWrapper types:
   - StructureModelWrapper (AF2-like models, direct prediction of structure)
   - FlowModelWrapper (diffusion and flow models)
   - EnergyBasedModelWrapper (name should be explanatory)
2. We can now write new samplers, rather than just using the default AF3 EDM-style sampler (which I've implemented in the new Sampler and TrajectorySampler protocols). Changing the sampler could improve sampling particularly for the ensemble sampling tasks were interested in.
3. Guidance is now separated into a two separate protocols:
   - TrajectoryScalerProtocol, which deals with methods that reweight and resample (SMC-type methods, or methods that require running multiple trajectories)
      - PureGuidance and FKSteering have been migrated to this new protocol, and use a TrajectorySampler and StepScaler to accomplish guidance. FKSteering has a regression in speed in this refactor, which I will open an issue for (having the sampler separated makes it more challenging to vmap over the reward and run batched/vectorized sampling)
   - StepScalerProtocol, which alters the actual step in the sampler using some reward function.
      - NoScalingScaler, which is a no-op scaler (useful for debugging the guidance in sampling)
      - DataSpaceDPSScaler, which is what the previous "tweedie-mode" was
      - NoiseSpaceDPSScaler, which is what the previous "training-free guidance" mode was, and backprops through the model to $x_t$
4. The structure processing is still messy, but is in its own function now, `process_structure_to_trajectory_input`, which hopefully will be easier to debug for shape mismatches (WHICH ARE STILL HAPPENING IN PROTENIX!!!)
5. I have included some improved tests that have been helping to catch many more problems. Coverage for scalers and the processing is still lower than I'd like and I've been testing that by hand mostly, but the ModelWrappers are well tested by it.

## Summary by CodeRabbit

* **New Features**
  * New EDM-based sampler for trajectory generation and improved sampling control.
  * Structured sampler, scaler and reward protocols for consistent interfaces.
  * Per-step guidance scalers (data-space and noise-space) and a no-op scaler.
  * Framework interop utilities for JAX ↔ PyTorch conversions.
  * Unified guidance APIs returning structured outputs for sampling workflows.
* **Tests**
  * Large suite of unit and integration tests and new test mocks for samplers, scalers, rewards, and wrappers.

### DIFF
--- a/src/sampleworks/core/samplers/protocol.py
+++ b/src/sampleworks/core/samplers/protocol.py
@@ -107,7 +107,7 @@ class StepParams:
     ) -> StepParams:
         """Return new StepParams with updated metadata.
         Will merge with existing metadata if present."""
-        merged_metadata = self.metadata if self.metadata is not None else {}
+        merged_metadata = dict(self.metadata) if self.metadata is not None else {}
         merged_metadata.update(metadata)
         return StepParams(
             step_index=self.step_index,
--- a/src/sampleworks/models/protocol.py
+++ b/src/sampleworks/models/protocol.py
@@ -138,7 +138,7 @@ class FlowModelWrapper(Protocol[C, FlowOrEnergyBasedModelOutputT]):
 
         Parameters
         ----------
-        x_t: FlowOrEquilibriumModelOutputT
+        x_t: FlowOrEnergyBasedModelOutputT
             Noisy structure at timestep :math:`t`.
         t: Float[Array, "*batch"]
             Current timestep/noise level.
@@ -147,7 +147,7 @@ class FlowModelWrapper(Protocol[C, FlowOrEnergyBasedModelOutputT]):
 
         Returns
         -------
-        FlowOrEquilibriumModelOutputT
+        FlowOrEnergyBasedModelOutputT
             Predicted clean sample or predicted noise.
         """
         ...
@@ -174,7 +174,7 @@ class FlowModelWrapper(Protocol[C, FlowOrEnergyBasedModelOutputT]):
 
         Returns
         -------
-        FlowOrEquilibriumModelOutputT
+        FlowOrEnergyBasedModelOutputT
             Noisy output.
         """
         ...
--- a/src/sampleworks/utils/guidance_script_utils.py
+++ b/src/sampleworks/utils/guidance_script_utils.py
@@ -237,6 +237,33 @@ def save_everything(
     scaler_type: str,
     final_state: torch.Tensor | None = None,
 ) -> None:
+    """Save everything: refined structure/ensemble CIF, trajectories, and losses.
+
+    When `final_state` is provided, its coordinates are written into the
+    `refined_structure` atom array (respecting the occupancy/NaN validity mask) before
+    saving.  Both the denoised and next-step trajectories are saved as
+    multi-model CIF files (subsampled every 10 steps via ``save_trajectory``).
+
+    Parameters
+    ----------
+    output_dir : str | Path
+        Directory to write all output files into. Created if it doesn't exist.
+    losses : list[Any]
+        Per-step loss values (may contain ``None`` entries for unguided steps).
+    refined_structure : dict
+        Atomworks structure dict whose ``"asym_unit"`` is used as the template
+        for saving.
+    traj_denoised : list[Any]
+        Denoised-prediction trajectory tensors, one per diffusion step.
+    traj_next_step : list[Any]
+        Next-step (noisy) trajectory tensors, one per diffusion step.
+    scaler_type : str
+        Scaler/guidance identifier, forwarded to ``save_trajectory`` for
+        scaler-specific handling. # TODO: handle more gracefully
+    final_state : torch.Tensor | None
+        Final coordinates with shape ``(ensemble, atoms, 3)``.  If ``None``,
+        the `refined_structure`'s existing coordinates are saved as-is.
+    """
     output_dir = Path(output_dir)
     output_dir.mkdir(parents=True, exist_ok=True)
 
@@ -244,15 +271,14 @@ def save_everything(
     from biotite.structure.io.pdbx import CIFFile, set_structure
 
     atom_array = refined_structure["asym_unit"]
-    if isinstance(atom_array, AtomArrayStack):
-        atom_array = atom_array[0]
+    atom_array_for_masking = atom_array[0] if isinstance(atom_array, AtomArrayStack) else atom_array
 
     if final_state is not None:
         ensemble_size = final_state.shape[0]
-        occupancy_mask = atom_array.occupancy > 0  # pyright: ignore[reportOptionalOperand]
-        nan_mask = ~np.any(np.isnan(atom_array.coord), axis=-1)  # pyright: ignore[reportArgumentType, reportCallIssue]
-        reward_param_mask = occupancy_mask & nan_mask
-        ensemble_array = stack([atom_array.copy() for _ in range(ensemble_size)])
+        reward_param_mask = atom_array_for_masking.occupancy > 0  # pyright: ignore[reportOptionalOperand]
+        reward_param_mask &= ~np.any(np.isnan(atom_array_for_masking.coord), axis=-1)  # pyright: ignore[reportArgumentType, reportCallIssue]
+
+        ensemble_array = stack([atom_array_for_masking.copy() for _ in range(ensemble_size)])
         ensemble_array.coord[:, reward_param_mask] = final_state.detach().cpu().numpy()  # pyright: ignore[reportOptionalSubscript]
         atom_array = ensemble_array
 
@@ -461,11 +487,11 @@ def get_job_result(
     start_time = epoch_seconds(started_at)
     end_time = epoch_seconds(ended_at)
     result = JobResult(
-        protein=getattr(args, "protein", "unknown"),
-        model=getattr(args, "model", "unknown"),
-        method=getattr(args, "method", "None"),
-        scaler=getattr(args, "guidance_type", "unknown"),
-        ensemble_size=getattr(args, "ensemble_size", 1),
+        protein=args.protein,
+        model=args.model,
+        method=getattr(args, "method", None),
+        scaler=args.guidance_type,
+        ensemble_size=args.ensemble_size,
         gradient_weight=getattr(args, "guidance_weight", -1.0),
         gd_steps=getattr(args, "num_gd_steps", -1),
         status=status,
--- a/src/sampleworks/models/rf3/wrapper.py
+++ b/src/sampleworks/models/rf3/wrapper.py
@@ -338,7 +338,8 @@ class RF3Wrapper:
         Parameters
         ----------
         features: dict[str, Any]
-            Model features dict (raw features, not GenerativeModelInput).
+            Model features dict that is computed internally in featurize
+            (raw features, not GenerativeModelInput).
         grad_needed: bool, optional
             Whether gradients are needed for this pass, by default False.
         **kwargs: dict, optional
--- a/src/sampleworks/utils/framework_utils.py
+++ b/src/sampleworks/utils/framework_utils.py
@@ -75,6 +75,10 @@ def ensure_torch(*arg_names: str, device: torch.device | None = None):
             ...
     """
 
+    # NOTE: Callable[P, T] -> Callable[P, T] is technically inexact. It
+    # accepts jax.Array where fn's signature expects torch.Tensor. ParamSpec can't
+    # express this broadened input in current Python typing, so we keep this for
+    # IDE autocomplete/return type propagation and accept the input-type imprecision.
     def decorator(fn: Callable[P, T]) -> Callable[P, T]:
         @wraps(fn)
         def wrapper(*args, **kwargs):
@@ -102,6 +106,7 @@ def ensure_jax(*arg_names: str):
     Decorator that converts specified PyTorch tensor arguments to JAX arrays.
     """
 
+    # NOTE: Same ParamSpec caveat as ensure_torch.
     def decorator(fn: Callable[P, T]) -> Callable[P, T]:
         @wraps(fn)
         def wrapper(*args, **kwargs):


src/sampleworks/core/samplers/protocol.py
```diff
@@ -0,0 +1,276 @@
...
+
+    # Guidance parameters (optional, for guided sampling)
+    reward: RewardFunctionProtocol | None = None
+    reward_inputs: RewardInputs | None = None
+
+    # Optional other metadata (could include velocity for momentum, etc.)
+    metadata: dict[str, Any] | None = None
+
+    @property
+    def t_effective(self) -> Float[Array, " batch"]:
```
k-chrispens: redundant at this point?

src/sampleworks/core/rewards/protocol.py
```diff
@@ -0,0 +1,96 @@
+from dataclasses import dataclass
+from typing import Protocol, runtime_checkable
+
+import numpy as np
+import torch
+from jaxtyping import Float
+
+
+@dataclass
+class RewardInputs:
```
marcuscollins: should it have a `@classmethod` to construct itself from an AtomArray?

src/sampleworks/core/rewards/protocol.py
```diff
@@ -0,0 +1,96 @@
...
+        occupancies
+            Per-atom occupancies, shape [batch, n_atoms]
+        unique_combinations
+            Optional pre-computed unique (element, b_factor) pairs
+        inverse_indices
+            Optional pre-computed inverse indices for vmap compatibility
+
+        Returns
+        -------
+        Float[torch.Tensor, ""]
```
marcuscollins: dimensions?

src/sampleworks/core/samplers/protocol.py
```diff
@@ -0,0 +1,276 @@
...
+        if self.total_steps is None:
+            return False
+        return self.step_index >= self.total_steps - 1
+
+    def with_reward(
+        self,
+        reward: RewardFunctionProtocol,
+        reward_inputs: RewardInputs,
+    ) -> StepContext:
+        """Return a new context with reward information for guided sampling."""
```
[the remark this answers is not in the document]
marcuscollins: In particular they'd need `__enter__` and `__exit__` methods

src/sampleworks/core/scalers/fk_steering.py
```diff
@@ -1,495 +1,465 @@
...
+from sampleworks.utils.framework_utils import Array
 
 
 class FKSteering:
-    """
-    Feynman-Kac Steering scaler.
+    """Feynman-Kac Steering trajectory scaler.
+
+    The key difference from the standard FKSteering implementation is that "particles"
+    are ensembles of structures
```
marcuscollins: What else would they be? Just one trajectory/structure?

src/sampleworks/models/protocol.py
```diff
@@ -0,0 +1,233 @@
...
+    Attributes:
+        x_init: Initial structure coordinates, shape (*batch, atoms, 3).
+                This can be a reference structure (e.g. for alignment during sampling)
+                or a noisy sample from a prior distribution. This should have the proper
+                shape expected for the given ensemble being sampled, e.g. (4, atoms, 3) for
+                ensemble size 4.
+        conditioning: Model-specific conditioning features, or None.
+    """
+
+    # TODO: make x_init more general (not just Float),
```
marcuscollins: What else could it be?

src/sampleworks/eval/structure_utils.py
```diff
@@ -1,13 +1,102 @@
...
+    mask: np.ndarray
+    occupancies: torch.Tensor
+    b_factors: torch.Tensor
+    elements: torch.Tensor
+
+
+# TODO: this function could maybe be an atomworks transform/use those?
+def process_structure_to_trajectory_input(
+    structure: dict,
+    coords_from_prior: torch.Tensor,
```
marcuscollins: meaning a (partially?) noised structure? Or really just a sample from some Normal or other basic distribution?

src/sampleworks/eval/structure_utils.py
```diff
@@ -1,13 +1,102 @@
...
+    features: GenerativeModelInput,
+    ensemble_size: int,
+) -> SampleworksProcessedStructure:
+    if features.conditioning:
```
marcuscollins: I think I would rewrite this:
```
if (features.conditioning and hasattr(features.conditioning, "true_atom_array") and
         features.conditioning.true_atom_array):
    atom_array = features.conditioning.true_atom_array
elif "asym_unit" in structure and isinstance(structure["asym_unit"], AtomArrayStack):
    atom_array = structure["asym_unit"][0]
else:
    raise ...
```

"flat is better than nested"

src/sampleworks/core/samplers/edm.py
```diff
@@ -0,0 +1,426 @@
...
+
+@dataclass(frozen=True, slots=True)
+class EDMSamplerConfig:
+    """Config to initialize an EDM sampler plus common step defaults."""
+
+    # Schedule parameters (AF3 defaults)
+    sigma_data: float = 16.0
+    s_max: float = 160.0
+    s_min: float = 4e-4
+    p: float = 7.0
```
marcuscollins: I remember what this is only because I read the paper again yesterday--it would be helpful to readers to add a docstring here.

src/sampleworks/core/samplers/edm.py
```diff
@@ -0,0 +1,426 @@
...
+            "gamma": self.gamma,
+        }
+
+
+@dataclass(frozen=True, slots=True)
+class EDMSamplerConfig:
+    """Config to initialize an EDM sampler plus common step defaults."""
+
+    # Schedule parameters (AF3 defaults)
+    sigma_data: float = 16.0
```
marcuscollins: We don't compute this from the data?

src/sampleworks/core/samplers/edm.py
```diff
@@ -0,0 +1,426 @@
...
+    scale_guidance_to_diffusion: bool = True
+    device: str | torch.device = "cpu"
+
+    # Default per-step scale (AF3 default)
+    step_scale: float = 1.5
+
+    def create_sampler(self) -> AF3EDMSampler:
+        """Create EDM sampler instance from this config."""
+        return AF3EDMSampler(
+            sigma_data=self.sigma_data,
```
marcuscollins: why not just have the AF3EDMSampler require an EDMSamplerConfig object as input to its __init__ method?

src/sampleworks/core/samplers/edm.py
```diff
@@ -0,0 +1,426 @@
...
+        Returns
+        -------
+        tuple[torch.Tensor, torch.Tensor | None, torch.Tensor]
+            (modified_delta, loss, proposal_shift)
+        """
+        scaler_metadata: dict[str, object] = {"x_t": noisy_state}
+        if context.metadata:
+            scaler_metadata = {**context.metadata, "x_t": noisy_state}
+
+        scaler_context = context.with_metadata(scaler_metadata)
```
marcuscollins: Should your context object perhaps just have a method to update/merge metadata? Or, if it is always a `dict` (instead of `None`) you could just do `context.metadata.update({"x_t": noisy_state})`. Much simpler.
```

File: src/sampleworks/core/samplers/edm.py [hunk 13]
```diff
@@ -0,0 +1,426 @@
...
+
+        guidance_direction, loss = scaler.scale(
+            x_hat_0_working_frame, scaler_context, model=model_wrapper
```
marcuscollins: Is `x_hat_0_working_frame` actually always \hat{x}_0? I.e., is it always time = 0? Is it for alignment?

src/sampleworks/core/samplers/edm.py
```diff
@@ -0,0 +1,426 @@
...
+        align_transform = None
+
+        if self.align_to_input and features is not None:
+            align_mask = (
+                torch.as_tensor(context.metadata.get("mask"))
+                if context.metadata and context.metadata.get("mask") is not None
+                else None
+            )
+            x_hat_0_working_frame, align_transform = align_to_reference_frame(
+                torch.as_tensor(x_hat_0),
```
marcuscollins: add a comment which one is being aligned to the other.

src/sampleworks/core/scalers/fk_steering.py
```diff
@@ -1,495 +1,465 @@
...
+        Returns
+        -------
+        GuidanceOutput
+            Output containing final state, trajectory, losses, and metadata.
+        """
 
-            - num_particles: int, optional
-                Number of particles (replicas) for FK steering (default: 10)
+        features = model.featurize(structure)
+        coords = cast(
```
marcuscollins: Is there any way to avoid all these `cast` calls? I know it doesn't really change the code, but it definitely makes it harder to read for me. Presumably `model.initialize_from_prior` is already annotated as returning a torch.Tensor?

src/sampleworks/core/scalers/fk_steering.py
```diff
@@ -1,495 +1,465 @@
...
+            structure=structure,
+            coords_from_prior=coords[: self.ensemble_size],
+            features=features,
+            ensemble_size=self.ensemble_size,
+        )
 
-            - fk_lambda: float, optional
-                Weighting factor for resampling (default: 1.0)
+        reward_inputs = RewardInputs(
+            elements=cast(torch.Tensor, processed.elements),
```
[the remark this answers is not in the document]
marcuscollins: Alternatively, why not make `process_structure_to_trajectory_input` return `RewardInputs`?

src/sampleworks/core/scalers/fk_steering.py
```diff
@@ -1,495 +1,465 @@
...
-                if align_transform is not None
-                else maybe_augmented_coords
-            )
-            eps_in_working_frame = (
-                apply_forward_transform(eps, align_transform, rotation_only=True)
-                if align_transform is not None
-                else eps
+        try:
+            is_resampling_step = step % self.resampling_interval == 0 and float(noise_scale) > 0  # pyright: ignore[reportArgumentType]
+        except ValueError:
```
marcuscollins: does this happen dynamically, or is it something that will happen when the first StepContext is retrieved and passed to fk steering? I wonder if you could handle this condition elsewhere.

src/sampleworks/core/scalers/fk_steering.py
```diff
@@ -1,495 +1,465 @@
...
-                scaled_guidance_update = (
-                    apply_forward_transform(
-                        scaled_guidance_update.reshape(num_particles * ensemble_size, -1, 3),
-                        align_transform,
-                        rotation_only=True,
-                    ).reshape(num_particles, ensemble_size, -1, 3)
-                    if align_transform is not None
-                    else scaled_guidance_update
-                )
+            raise
```
marcuscollins: Won't this just raise a generic Exception? I thought you had to use `except Exception as e:... raise e`?

src/sampleworks/core/scalers/fk_steering.py
```diff
@@ -1,495 +1,465 @@
...
-                # energy_traj shape: (num_particles, steps)
-                energy_traj = torch.cat((energy_traj, energy.unsqueeze(1)), dim=1)
-
-                if i == partial_diffusion_step:
-                    log_G = -1 * energy
+        Returns
+        -------
+        AtomArrayStack | object
+            AtomArrayStack with coordinates from lowest-loss particle,
+            or mock object in test contexts.
```
marcuscollins: probably don't need to say this.

src/sampleworks/core/scalers/fk_steering.py
```diff
@@ -1,495 +1,465 @@
...
-                        allow_gradients=False,
-                    ).reshape(num_particles, ensemble_size, -1, 3)
-
-                dt = sigma_t - t_hat
-                denoised_over_sigma = (noisy_coords - denoised_working_frame) / t_hat
-
-                coords_next = noisy_coords + step_scale * dt * denoised_over_sigma
-
-                coords = coords_next.detach().clone()
+                    atom_array = structure["asym_unit"][0]
```
marcuscollins: I feel like either one of my comments disappeared or this exact code has appeared elsewhere (in which case there's an opportunity to refactor/condense code).  I would suggest reworking the logic here so that you only have to write `atom_array = structure["asym_unit"][0]` once.

marcuscollins: Also, I think you're taking just the first structure because this structure is actually the input, and you're just building a container for the actual ensemble results, is that correct? If so, I would rename the `structure` variable to indicate that it is an original input structure, and add a comment that at this point this is still a dummy atom_array.

src/sampleworks/core/scalers/fk_steering.py
```diff
@@ -1,495 +1,465 @@
...
+            atom_array = structure["asym_unit"][0]
 
-            trajectory_next_step.append(coords.clone().cpu())
+        occupancy_mask = atom_array.occupancy > 0
+        nan_mask = ~np.any(np.isnan(atom_array.coord), axis=-1)
+        reward_param_mask = occupancy_mask & nan_mask
 
-        # Stack atom array to match ensemble size
-        final_atom_array = stack([atom_array] * ensemble_size)
+        final_atom_array = stack([atom_array] * self.ensemble_size)
```
marcuscollins: I'm a little lost--you haven't actually copied in the final coordinates yet, right? Some comments would be helpful here. 

Also, you'll have to do something similar for PureGuidance, right? Should this be in some utility or parent class method?

src/sampleworks/core/scalers/fk_steering.py
```diff
@@ -1,495 +1,465 @@
...
-        final_atom_array.coord[..., reward_param_mask, :] = (  # type: ignore[reportOptionalSubscript] coords will be subscriptable
-            coords[min_energy_index].cpu().numpy()
-        )
+        if loss_history:
+            min_loss_index = torch.argmin(loss_history[-1])
+        else:
+            min_loss_index = 0
+
+        coords_4d = coords.reshape(effective_particles, self.ensemble_size, -1, 3)
+        if final_atom_array.coord is not None:
```
marcuscollins: What if it is None? Do we just return nothing? That seems weird.

src/sampleworks/core/scalers/pure_guidance.py
```diff
@@ -1,364 +1,149 @@
...
+                f"Partial diffusion starting from step {self.starting_step} of {self.num_steps}."
             )
-            maybe_augmented_coords = cast(
-                torch.Tensor,
-                apply_forward_transform(coords, transform, rotation_only=False)
-                if transform is not None
-                else coords,
+            starting_context = sampler.get_context_for_step(self.starting_step - 1, schedule)
+            # coords will be a noisy version of input coords at this t
+            coords = processed_structure.input_coords + coords * torch.as_tensor(
```
marcuscollins: `coords` at this point comes from the model initialization, right? So it's a sample from some noise distribution? I guess I haven't thought this through at this detail before, but this is different than the usual movie that's shown where everything starts out in a ball near the origin and explodes out to a sane structure (which, not that I think of it, doesn't actually make sense in terms of the processes that are applied in training).

src/sampleworks/core/scalers/score_scalers.py
```diff
@@ -0,0 +1,147 @@
...
+
+    def guidance_strength(self, context: StepContext) -> Float[Tensor, " batch"]:
+        t = context.t_effective
+        return torch.zeros_like(torch.as_tensor(t))
+
+
+class DataSpaceDPSScaler:
+    """Step scaler that operates in the data space.
+
+    Computes gradients only on the denoised prediction (x̂₀), avoiding backprop
```
marcuscollins: What's theta in the subscript?

src/sampleworks/core/scalers/score_scalers.py
```diff
@@ -0,0 +1,147 @@
...
+        *,
+        model: FlowModelWrapper | None = None,
+    ) -> tuple[Float[Tensor, "*batch atoms 3"], Float[Tensor, " batch"]]:
+        if context.reward is None or context.reward_inputs is None:
+            raise ValueError(
+                "StepContext missing reward/reward_inputs. "
+                "Use context.with_reward() before calling scale()."
+            )
+
+        x0_for_grad = state.detach().requires_grad_(True)
```
marcuscollins: That's a strange statement. Doesn't "detach" usually mean to remove something from the gradient graph?

src/sampleworks/core/scalers/score_scalers.py
```diff
@@ -0,0 +1,147 @@
...
+from jaxtyping import Float
+from torch import Tensor
+
+
+if TYPE_CHECKING:
+    from sampleworks.core.samplers.protocol import StepContext
+    from sampleworks.models.protocol import FlowModelWrapper
+
+
+class NoScalingScaler:
```
marcuscollins: Do any of these get used anywhere? I haven't seen them. If they are unused, do you have plans to test them, or are they speculative? If speculative, maybe remove?

src/sampleworks/models/boltz/wrapper.py
```diff
@@ -205,7 +310,7 @@ def __init__(
                 Boltz2.load_from_checkpoint(
                     checkpoint_path,
                     strict=True,
-                    predict_args=asdict(predict_args),
+                    predict_args=asdict(PredictArgs()),
```
marcuscollins: Why make this non-configurable?

src/sampleworks/models/boltz/wrapper.py
```diff
@@ -304,140 +397,155 @@ def _setup_data_module(
...
-            Boltz-2 input features (from dataloader).
+        GenerativeModelInput[BoltzConditioning]
+            Model input with x_init and Pairformer conditioning.
         """
-        # If featurize is called again, we should clear cached representations
-        # to avoid using stale data
         self.cached_representations.clear()
 
-        # Side effect: creates Boltz input YAML file in out_dir
+        config = structure.get("_boltz_config", BoltzConfig())
```
marcuscollins: Does this code not create a YAML file in out_dir anymore? I think it is a good idea to note that in comments.

src/sampleworks/models/boltz/wrapper.py
```diff
@@ -28,17 +28,115 @@
...
+
+    """
+
+    out_dir: str | Path | None = None
+    num_workers: int = 8
+    ensemble_size: int = 1
+    recycling_steps: int = 3
+
+
+def annotate_structure_for_boltz(
```
marcuscollins: this method actually does more than just annotate the structure--it generates a lot of files (I think) and sets up the data loader. I would give it a better name.

src/sampleworks/models/boltz/wrapper.py
```diff
@@ -304,140 +397,155 @@ def _setup_data_module(
...
-            relative_position_encoding = self.model.rel_pos(features)
+        mask: Tensor = features["token_pad_mask"]
+        pair_mask = mask[:, :, None] * mask[:, None, :]
+        s_inputs = self.model.input_embedder(features)
 
-            z_init = z_init + relative_position_encoding
-            z_init = z_init + self.model.token_bonds(features["token_bonds"].float())
+        s_init = self.model.s_init(s_inputs)
+        z_init = (
+            self.model.z_init_1(s_inputs)[:, :, None] + self.model.z_init_2(s_inputs)[:, None, :]
```
marcuscollins: Can you write this line the original way? It was easier to read. I personally don't like splitting lines with square brackets, even though it is technically correct.

src/sampleworks/models/boltz/wrapper.py
```diff
@@ -304,140 +397,155 @@ def _setup_data_module(
...
-                predict_args.
+        # Use atom count from Boltz featurization (via conditioning) to ensure consistency
+        # between x_init shape and atom_pad_mask used in step(). Note: x_init=None is
+        # a temporary placeholder; initialize_from_prior derives shape from conditioning.
+        temp_features = GenerativeModelInput(x_init=None, conditioning=conditioning)  # pyright: ignore[reportArgumentType]
+        x_init = self.initialize_from_prior(batch_size=ensemble_size, features=temp_features)
 
+        return GenerativeModelInput(x_init=x_init, conditioning=conditioning)
+
+    def _pairformer_pass(
```
marcuscollins: Is there something different about your pairformer pass that doesn't allow us to just use whatever is in the Boltz model? (Maybe there isn't one method to do this and only this?) It seems like there should be.

src/sampleworks/models/boltz/wrapper.py
```diff
@@ -457,269 +565,123 @@ def step(self, features: dict[str, Any], grad_needed: bool = False, **kwargs) ->
...
-        s_inputs = features.get("s_inputs", None)
-        relative_position_encoding = features.get("relative_position_encoding", None)
-        # These are the input features to the conditioning networks
-        feats = features.get("feats", None)
-
-        if any(x is None for x in [s, z, s_inputs, relative_position_encoding, features]):
-            raise ValueError("Missing required features for denoise_step")
+        # Ensure t has batch dimension and is on correct device
+        if isinstance(t, (int, float)):
+            t = torch.tensor([t], device=self.device, dtype=torch.float32)
```
marcuscollins: is the batch size just one? You don't need to tile this?

src/sampleworks/models/boltz/wrapper.py
```diff
@@ -457,269 +565,123 @@ def step(self, features: dict[str, Any], grad_needed: bool = False, **kwargs) ->
...
-        sigma_tm = self.noise_schedule["sigma_tm"][int(timestep)]
-        sigma_t = self.noise_schedule["sigma_t"][int(timestep)]
-        gamma = self.noise_schedule["gamma"][int(timestep)]
+        pad_len = atom_mask.shape[1] - x_t.shape[1]
+        if pad_len >= 0:
+            padded_x_t = pad_dim(x_t, dim=1, pad_len=pad_len)
+        else:
+            raise ValueError("pad_len is negative, cannot pad x_t")
+
+        padded_atom_coords_denoised = self.model.structure_module.preconditioned_network_forward(
```
marcuscollins: I've lost track--was z only needed when computing pairformer features?

src/sampleworks/models/boltz/wrapper.py
```diff
@@ -873,390 +819,260 @@ def _setup_data_module(
             constraints_dir=processed.constraints_dir,
         )
 
-    def featurize(self, structure: dict, **kwargs) -> dict[str, Any]:
-        """From an Atomworks structure, calculate model features.
+    def featurize(self, structure: dict) -> GenerativeModelInput[BoltzConditioning]:
```
marcuscollins: How much of this is identical to what's used for Boltz2? Can we condense?

src/sampleworks/models/protenix/structure_processing.py
```diff
@@ -728,6 +728,18 @@ def structure_to_protenix_json(structure: dict) -> dict[str, Any]:
     has_ligand_bonds = lig_polymer_bonds.size > 0 or lig_lig_bonds.size > 0
 
     if has_modifications or has_ligand_bonds:
+        entity_min_res_id: dict[str, int] = {}
+        for label_entity_id in np.unique(cast(np.ndarray, atom_array.label_entity_id)):
```
marcuscollins: does atom_array not have a method for this? I thought maybe it did.

src/sampleworks/models/protenix/structure_processing.py
```diff
@@ -728,6 +728,18 @@ def structure_to_protenix_json(structure: dict) -> dict[str, Any]:
...
+            for chain_id, info in chain_info.items():
```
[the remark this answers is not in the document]
marcuscollins: Or, likely, just use `for chain_id in chain_info:`

src/sampleworks/models/protenix/structure_processing.py
```diff
@@ -728,6 +728,18 @@ def structure_to_protenix_json(structure: dict) -> dict[str, Any]:
     has_ligand_bonds = lig_polymer_bonds.size > 0 or lig_lig_bonds.size > 0
 
     if has_modifications or has_ligand_bonds:
+        entity_min_res_id: dict[str, int] = {}
```
marcuscollins: Can you add a comment to explain the purpose of this block?

src/sampleworks/models/protenix/structure_processing.py
```diff
@@ -749,13 +761,17 @@ def structure_to_protenix_json(structure: dict) -> dict[str, Any]:
             for atoms in token_bonds[:, :2]:
                 bond_dict = {}
                 for i in range(2):
-                    position = cast(np.ndarray, atom_array.res_id)[atoms[i]]
+                    raw_res_id = int(cast(np.ndarray, atom_array.res_id)[atoms[i]])
```
marcuscollins: ditto, comments will help a lot here to understand the intent. I would make as many of these kinds of things utility methods as possible, to make it easier to test and understand.

src/sampleworks/utils/guidance_script_utils.py
```diff
@@ -226,15 +235,29 @@ def save_everything(
...
 ) -> None:
     output_dir = Path(output_dir)
     output_dir.mkdir(parents=True, exist_ok=True)
 
     logger.info("Saving results")
     from biotite.structure.io.pdbx import CIFFile, set_structure
 
+    atom_array = refined_structure["asym_unit"]
+    if isinstance(atom_array, AtomArrayStack):
+        atom_array = atom_array[0]
```
marcuscollins: I am confused--don't we want to save all members of the ensemble?

src/sampleworks/utils/guidance_script_utils.py
```diff
@@ -389,19 +461,19 @@ def get_job_result(
...
     end_time = epoch_seconds(ended_at)
     result = JobResult(
-        protein=args.protein,
-        model=args.model,
-        method=args.method if hasattr(args, "method") else "None",
-        scaler=args.guidance_type,
-        ensemble_size=args.ensemble_size,
-        gradient_weight=args.guidance_weight if hasattr(args, "guidance_weight") else -1.0,
-        gd_steps=args.num_gd_steps if hasattr(args, "num_gd_steps") else -1,
+        protein=getattr(args, "protein", "unknown"),
```
marcuscollins: Aren't many of these args required to be set? `args.model` has to have some proper value right, it can't be unknown. So we should error out if things like that aren't set.

src/sampleworks/core/rewards/protocol.py
```diff
@@ -0,0 +1,96 @@
...
+        ----------
+        coordinates
+            Atomic coordinates, shape [batch, n_atoms, 3]
+        elements
+            Atomic element indices, shape [batch, n_atoms]
+        b_factors
+            Per-atom B-factors, shape [batch, n_atoms]
+        occupancies
+            Per-atom occupancies, shape [batch, n_atoms]
+        unique_combinations
```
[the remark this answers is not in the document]
k-chrispens: This is required if the reward is vmappable. I'll add a comment addressing this.

src/sampleworks/core/rewards/protocol.py
k-chrispens: Good idea.

src/sampleworks/core/rewards/utils.py
```diff
@@ -0,0 +1,122 @@
...
+    -------
+    RewardInputs
+        Dataclass containing all inputs needed for reward function computation
+    """
+    occupancy_mask = atom_array.occupancy > 0  # pyright: ignore[reportOptionalOperand]
+    nan_mask = ~np.any(np.isnan(atom_array.coord), axis=-1)  # pyright: ignore[reportArgumentType, reportCallIssue]
+    reward_param_mask = occupancy_mask & nan_mask
+
+    elements_list = [
+        ATOMIC_NUM_TO_ELEMENT.index(e.title())
```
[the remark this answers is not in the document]
k-chrispens: Had claude search for all places where I use this pattern and replace with reverse lookup dict.

src/sampleworks/core/samplers/protocol.py
```diff
@@ -0,0 +1,276 @@
...
+    - store any per-step arrays needed to compute step context
+    - provide `as_dict()` for compatibility/serialization.
+    """
+
+    def as_dict(self) -> dict[str, Float[torch.Tensor, ...]]:  # pragma: no cover
+        raise NotImplementedError
+
+
+@dataclass(frozen=True, slots=True)
+class StepContext:
```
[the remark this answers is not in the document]
k-chrispens: yes, that is in the AF3EDMSampler schedule parameters (L109-110 in src/sampleworks/core/samplers/edm.py)

src/sampleworks/core/samplers/protocol.py
k-chrispens: That happens in the SamplerSchedule implementation (src/sampleworks/core/samplers/edm.py L185), so that's why I had marked it as redundant.

src/sampleworks/core/samplers/protocol.py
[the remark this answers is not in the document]
k-chrispens: I'll rename to StepParams

src/sampleworks/core/samplers/protocol.py
```diff
@@ -0,0 +1,276 @@
...
+@dataclass(frozen=True, slots=True)
+class SamplerStepOutput(Generic[StateT]):  # noqa: UP046
+    """Output from a single sampler step.
+
+    Encapsulates the updated state along with optional intermediate values
+    useful for trajectory tracking and debugging.
+    """
+
+    state: StateT
+    """Updated coordinates after the sampling step."""
```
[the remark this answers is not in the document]
k-chrispens: I'll put them in the docstring instead

src/sampleworks/core/samplers/protocol.py
```diff
@@ -0,0 +1,276 @@
...
+
+    def step(
+        self,
+        state: StateT,
+        model_wrapper: ModelWrapperT,
+        context: StepContext,
+        *,
+        scaler: StepScalerProtocol | None = None,
+        features: GenerativeModelInput | None = None,
+    ) -> SamplerStepOutput[StateT]:
```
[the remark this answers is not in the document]
k-chrispens: It is a type annotation here, so SamplerStepOutput[torch.Tensor] would mean that the outputs from that Sampler are Tensors wherever we see StateT

src/sampleworks/core/scalers/fk_steering.py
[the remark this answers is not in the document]
k-chrispens: I don't quite want to do that since the structure dictionary could be useful here for other things potentially? But maybe we should consolidate more.

src/sampleworks/core/scalers/fk_steering.py
```diff
@@ -1,495 +1,465 @@
...
+            b_factors=cast(torch.Tensor, processed.b_factors),
+            occupancies=cast(torch.Tensor, processed.occupancies),
+            input_coords=cast(torch.Tensor, processed.input_coords),
+            reward_param_mask=processed.mask,
+            mask_like=torch.as_tensor(processed.mask, dtype=coords.dtype, device=coords.device),
+        )
 
-            - num_gd_steps: int, optional
-                Number of gradient descent steps on x0 (default: 0)
+        schedule = sampler.compute_schedule(self.num_steps)
```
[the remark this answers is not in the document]
k-chrispens: Yeah, I did kinda YOLO this refactor. I got a bit excited about how it would improve things but it would have been better to think through more for sure 😅

As far as the sampler having this method, I think it is important because it gets us that initial state concisely. Maybe we could have some other thing manage it, but that adds another piece of complexity, though maybe it would be worth it

src/sampleworks/core/scalers/fk_steering.py
k-chrispens: For PureGuidance we can just return the final state, but here we have to do the selection of the best loss. I fixed this in the updates I have

src/sampleworks/core/scalers/fk_steering.py
```diff
@@ -1,495 +1,465 @@
...
+        particle's loss gradient is computed only from its own state.
 
-            - alignment_reverse_diffusion: bool, optional
-                Whether to perform alignment of noisy coords to denoised coords
-                during reverse diffusion steps. This is relevant for doing Boltz-2-like
-                alignment during diffusion. (default: False)
+        Parameters
+        ----------
+        coords : torch.Tensor
+            Current coordinates with shape (particles * ensemble, atoms, 3).
```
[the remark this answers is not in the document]
k-chrispens: Mainly so that we can pass step_output.state directly in. The state still has things in 3d, but it is easier to handle the FK resampling in 4d so I return the relevant tensors for that as 4d. This is a bit clunky and will be improved ideally when this can be vmapped

src/sampleworks/core/samplers/edm.py
```diff
@@ -0,0 +1,426 @@
...
+
+@dataclass
+class AF3EDMSampler:
+    """EDM-style sampler from AF3-like models.
+
+    All constants are configurable via constructor for model-specific values.
+    Default values match AF3 parameterization.
+
+    This sampler implements the EDM (Karras et al.) style sampling
+    approach as used in AlphaFold3 and related models, which is the Euler
```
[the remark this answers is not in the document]
k-chrispens: EDM does, but all the AF3 models use the Euler solver. Mostly for compute reasons I think, because they do a short rollout of the model during training. Could be interesting to experiment though!

src/sampleworks/core/samplers/edm.py
k-chrispens: This is calculated based on prior assumptions about the data, so it's something we can play with empirically, but 16 is what AF3 uses

src/sampleworks/core/samplers/edm.py
```diff
@@ -0,0 +1,426 @@
...
+            dt=dt,
+            noise_scale=eps_scale,
+        )
+
+    def _apply_scaler_guidance(
+        self,
+        scaler: StepScalerProtocol,
+        x_hat_0_working_frame: Float[torch.Tensor, "*batch n 3"],
+        noisy_state: Float[torch.Tensor, "*batch n 3"],
+        delta: torch.Tensor,
```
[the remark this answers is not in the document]
k-chrispens: Delta is the denoising update $(x_t - \hat{x}_0) / t, we use this here to optionally scale the guidance to the size of the denoising update (Bronstein paper does this)

src/sampleworks/core/samplers/edm.py
k-chrispens: I'll update the context.with_metadata to keep previous values if not overwritten and document that - it will use context.metadata.update under the hood.

src/sampleworks/core/scalers/pure_guidance.py
```diff
@@ -1,364 +1,149 @@
...
-        # TODO: jank way to get atomic numbers, fix this in real space density
-        elements = [
-            ATOMIC_NUM_TO_ELEMENT.index(e.title()) for e in atom_array.element[reward_param_mask]
-        ]
-        elements = einx.rearrange("n -> b n", torch.Tensor(elements), b=ensemble_size)
-        b_factors = einx.rearrange(
-            "n -> b n",
-            torch.Tensor(atom_array.b_factor[reward_param_mask]),
-            b=ensemble_size,
+        processed_structure = process_structure_to_trajectory_input(
```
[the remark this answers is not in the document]
k-chrispens: It shouldn't... but it does right now (gets there with features.conditioning.__name__). Hopefully now that this is refactored out now though it will be easier to get there.

src/sampleworks/core/scalers/step_scalers.py
```diff
@@ -0,0 +1,147 @@
...
+        zeros = torch.zeros_like(state)
+        loss = torch.zeros(state.shape[0], device=state.device)
+        return zeros, loss
+
+    def guidance_strength(self, context: StepContext) -> Float[Tensor, " batch"]:
+        t = context.t_effective
+        return torch.zeros_like(torch.as_tensor(t))
+
+
+class DataSpaceDPSScaler:
```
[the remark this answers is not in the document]
k-chrispens: Yep!

src/sampleworks/models/boltz/wrapper.py
```diff
@@ -457,269 +565,123 @@ def step(self, features: dict[str, Any], grad_needed: bool = False, **kwargs) ->
...
-            Timestep/noise level - for Boltz, this is the reverse timestep.
-            This means that it starts from 0 and goes up to (sampling_steps - 1), so it
-            is the "reverse" diffusion time.
-        ensemble_size: int, optional
-            Number of noisy samples to generate per input structure (default 1).
-        **kwargs: dict, optional
-            Additional keyword arguments needed for classes that implement this Protocol
+        batch_size : int
+            Number of noisy samples to generate.
+        features: GenerativeModelInput[BoltzConditioning] | None, optional
```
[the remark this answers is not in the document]
k-chrispens: I'll update the docstring

src/sampleworks/models/boltz/wrapper.py
```diff
@@ -457,269 +565,123 @@ def step(self, features: dict[str, Any], grad_needed: bool = False, **kwargs) ->
...
-            )
+        num_atoms = int(atom_mask.sum())
 
-        if noise_level == 0:
-            noisy_coords = self.noise_schedule["sigma_tm"][0] * torch.randn_like(coords)
-        else:
-            noisy_coords = coords + self.noise_schedule["sigma_tm"][
-                int(noise_level)
-            ] * torch.randn_like(coords)
+        x_init = torch.randn((batch_size, num_atoms, 3), device=self.device)
```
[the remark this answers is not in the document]
k-chrispens: Those are generated by the sampler, which will define the inference noise schedule

src/sampleworks/utils/guidance_script_utils.py
```diff
@@ -226,15 +235,29 @@ def save_everything(
     traj_denoised: list[Any],
     traj_next_step: list[Any],
     scaler_type: str,
+    final_state: torch.Tensor | None = None,
```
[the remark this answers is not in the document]
k-chrispens: refined_structure is a structure dict, final_state is just the final trajectory coords output, which we then need to combine with the structure info.

Issue #85, opened 2026-02-10
refactor save_everything in guidance_script_utils.py to better accomodate refactor


k-chrispens:
The current `save_everything` is in between where it was pre-refactor and post-refactor. We should standardize this saving utility to make more sense, and make sure it tracks the metadata we care about. RE: Marcus's comment in guidance_script_utils.py, PR #73

Issue #80, opened 2026-02-09
Add utility for matching batches across Tensors and Arrays (will be useful for particle methods, etc.)


k-chrispens:
Implement a utility (probably in framework_utils) to match the batch sizes across Tensors. Call it `match_batch` :)

Could be used in AF3EDMSampler guidance weight expansion, etc.

Issue #42, opened 2025-12-20
Use more and more interesting structures as test fixtures


k-chrispens:
Multimodel structures, large structures, structures with missing atoms and residues

PR #108, opened 2026-02-26
fix(patch): applies patch functionality in a cleaner, tested fashion on top of the refactor

### LINKED ISSUE #88: Apply patch updates more cleanly once refactor is complete
The patch in v0.0.1a1 is pretty hacky, but can be more cleanly applied post-refactor. The main perspective is that we want to always save the model output coordinates, even if those don't fully map onto the input structure (which might have missing atoms, etc.). We can use the new `filter_to_common_atoms` in that patch to get around this in alignment, etc, and then update our saving logic #85 #86 to account for this with sensible defaults.

k-chrispens:
I've created new AtomReconciler object that allows us to handle all these cases where the atom counts don't quite line up. This should make us much more robust to different models operating on different subsets of atoms.

Implementing this required making lots of targeted modifications to the existing scalers, samplers, etc. and also updates to the saving utilities. I also changed `filter_to_common_atoms` for this.

This also picked up my match_batch utility :)

The massive LOC change is because I added more examples to tests/resources, which was necessary for proper testing (partially addresses #42)

Closes #88 #80 #85

## Summary by CodeRabbit

* **New Features**
  * Model-aware atom reconciliation and propagation of model templates through processing, sampling, and saving.
  * Batch-matching utility for reliable ensemble sizing and device placement.
  * Atom-ID normalization helper for robust cross-structure matching.

* **Bug Fixes**
  * Validation for atom-count mismatches between states and reward inputs.
  * Ensured coordinate contiguity, consistent batching, and device placement during sampling and saving.
  * More reliable trajectory metadata when model/template mismatches occur.

* **Refactor**
  * Consolidated trajectory and metadata handling; unified final-state saving with model-aware masking.

* **Tests**
  * Large expansion of unit and integration tests and fixtures covering mismatch, reconciliation, alignment, and utilities.


src/sampleworks/core/rewards/protocol.py
```diff
@@ -68,6 +68,10 @@ def from_atom_array(
 
         total_batch_size = num_particles * ensemble_size if num_particles > 1 else ensemble_size
 
+        # ensure contiguous arrays for safe conversion to PyTorch tensors
```
marcuscollins: Is this for efficiency, or can something actually go wrong converting to torch?

src/sampleworks/utils/atom_reconciler.py
```diff
@@ -0,0 +1,265 @@
+"""Adapter between model and structure atom coordinate spaces.
+
+When a generative model's internal atom representation differs from the input
+structure (different atom count, different residue numbering), the
+:class:`AtomReconciler` translates coordinates between the two spaces.
+"""
+
+from __future__ import annotations
```
marcuscollins: Vratin was using this too... I had the distinct impression that it was out of date and that we could just classes by default now for annotations as long as they were imported. Is the goal to not have to import things?

src/sampleworks/core/samplers/protocol.py
```diff
@@ -88,38 +95,67 @@ def with_reward(
...
+        self,
+        reconciler: AtomReconciler,
+        alignment_reference: Float[Array, "*batch atoms 3"] | None = None,
+    ) -> StepParams:
+        """Return a new ``StepParams`` with atom reconciliation context.
+
+        Parameters
+        ----------
+        reconciler : AtomReconciler
+            Adapter that translates between model and structure atom spaces.
```
marcuscollins: "translates between model and structure atom spaces" shows up in a lot of places, and it's kind of confusing. Which is the model and which is the structure? Is it really a reference structure and the sampled structure being generated by Boltz, Protenix, RF3...?

src/sampleworks/eval/structure_utils.py
```diff
@@ -59,12 +66,43 @@ def to_reward_inputs(self, device: torch.device | str = "cpu") -> RewardInputs:
...
+            struct_b_factors = torch.as_tensor(
+                cast(np.ndarray, self.atom_array.b_factor)[
+                    reconciler.struct_indices.detach().cpu().numpy()
+                ],
+                device=reward_inputs.b_factors.device,
+                dtype=reward_inputs.b_factors.dtype,
+            )
+            model_indices = reconciler.model_indices.to(device=reward_inputs.b_factors.device)
+            updated_b_factors = reward_inputs.b_factors.clone()
+            updated_b_factors[..., model_indices] = struct_b_factors
```
marcuscollins: I wonder if you could make a method for updating a tensor with the model-aligned parts. It happens several places, might make sense to make a method to avoid errors.

src/sampleworks/eval/structure_utils.py
```diff
@@ -131,20 +172,59 @@ def process_structure_to_trajectory_input(
...
-            e=ensemble_size,
-        ),
+    # Build reconciler from model and structure atom arrays.
+    model_atom_array = (
+        cast(AtomArray | None, conditioning.model_atom_array)
+        if hasattr(conditioning, "model_atom_array")
+        else None
     )
+    if model_atom_array is not None:
+        reconciler = AtomReconciler.from_arrays(model_atom_array, atom_array)  # pyright: ignore[reportArgumentType]
```
marcuscollins: so model_atom_array is... the previous structure? Or always the reference structure? And atom_array is the updated structure? A comment would help here.

src/sampleworks/models/boltz/wrapper.py
```diff
@@ -34,9 +35,218 @@
...
+    ----------
+    npz_path
+        Path to the ``.npz`` in ``<out_dir>/processed/structures/``.
+
+    Returns
+    -------
+    AtomArray
+        One atom per model atom with chain_id, res_id, atom_name, res_name.
+    """
+    if npz_path.suffix != ".npz":
```
marcuscollins: I don't think either of these are really necessary. When you go to load the file, np.load will raise these errors, although I suppose the traceback will be a little shorter.

src/sampleworks/models/boltz/wrapper.py
```diff
@@ -34,9 +35,218 @@
...
+        raise ValueError(f"{npz_path}: residues.atom_idx must be sorted ascending")
+    if np.any(np.diff(chain_residue_starts) < 0):
+        raise ValueError(f"{npz_path}: chains.res_idx must be sorted ascending")
+
+    n_atoms = len(atoms)
+    atom_indices = np.arange(n_atoms, dtype=np.int64)
+
+    # Map each atom to residue to chain via searchsorted on start indices.
+    atom_res_idx = np.searchsorted(residue_atom_starts, atom_indices, side="right") - 1
+    if np.any((atom_res_idx < 0) | (atom_res_idx >= len(residues))):
```
marcuscollins: Can this actually happen?

src/sampleworks/models/boltz/wrapper.py
```diff
@@ -34,9 +35,218 @@
...
+            "skipping model_atom_array (mismatch reconciliation disabled)"
+        )
+        return None
+
+    if len(npz_files) > 1:
+        logger.warning(
+            f"Found {len(npz_files)} NPZ files in {structures_dir}; "
+            "skipping model_atom_array (mismatch reconciliation disabled)"
+        )
+        return None
```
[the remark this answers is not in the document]
k-chrispens: I don't know what would cause there to be multiple npzs, there should only ever be one in the structure dir, so I am going to keep as is so we are alerted if this changes upstream in boltz

src/sampleworks/utils/guidance_script_utils.py
```diff
@@ -76,6 +77,32 @@ def save_trajectory(
...
 
+def _write_coords_into_array(
+    array_copy: AtomArrayStack,
+    coords: np.ndarray,
+    reward_param_mask: np.ndarray,
+) -> None:
+    """**Mutates** ``array_copy.coord`` in-place with trajectory coordinates.
+
+    When the trajectory spans all atoms in the array (model trajectories during
+    a mismatch run), coords are assigned directly to ``.coord``. Otherwise the
```
marcuscollins: What is a "mismatch run"? 

This comment is difficult to understand.

src/sampleworks/utils/guidance_script_utils.py
```diff
@@ -269,24 +297,43 @@ def save_everything(
...
     output_dir.mkdir(parents=True, exist_ok=True)
 
     logger.info("Saving results")
     from biotite.structure.io.pdbx import CIFFile, set_structure
 
-    atom_array = refined_structure["asym_unit"]
-    atom_array_for_masking = atom_array[0] if isinstance(atom_array, AtomArrayStack) else atom_array
+    base_atom_array = ensure_atom_array_stack(refined_structure["asym_unit"])[0]
+
+    # Use model template for mismatch runs when available
```
marcuscollins: Terminology is weird here.

src/sampleworks/utils/guidance_script_utils.py
```diff
@@ -269,24 +297,43 @@ def save_everything(
...
+    atom_array_for_masking: AtomArray = (
+        model_atom_array if model_atom_array is not None else base_atom_array
+    )
+
+    # Build occupancy mask properly as model atom arrays may lack occupancy annotation
+    if (
+        hasattr(atom_array_for_masking, "occupancy")
+        and atom_array_for_masking.occupancy is not None
+    ):
+        reward_param_mask = atom_array_for_masking.occupancy > 0  # pyright: ignore[reportOptionalOperand]
```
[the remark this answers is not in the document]
marcuscollins: Also, migrate the `# pyright` annotations

src/sampleworks/utils/guidance_script_utils.py
```diff
@@ -269,24 +297,43 @@ def save_everything(
...
+        reward_param_mask &= ~np.any(np.isnan(atom_array_for_masking.coord), axis=-1)  # pyright: ignore[reportArgumentType, reportCallIssue]
+    else:
+        reward_param_mask = np.ones(len(atom_array_for_masking), dtype=bool)
```
marcuscollins: do you not need to check for NaNs here?

marcuscollins (approved):
Haven't reviewed tests yet. Overall LGTM. The main thing is that the logic isn't always obvious. Some careful renaming of variables and lots of (human written) comments will help. 

I am approving, but look over the comments before you merge and make issues or fix them in further commits.

src/sampleworks/core/rewards/protocol.py
k-chrispens: I got a memory error during testing, which I believe is because torch attempts to directly copy over the np data when you assign to tensor. If it isn't contiguous, downstream torch ops break. I had this same issue back when I was writing the density code

src/sampleworks/utils/framework_utils.py
```diff
@@ -127,3 +127,49 @@ def wrapper(*args, **kwargs):
...
+        raise ValueError("match_batch requires ndim >= 1")
+    b, n = array.shape[0], target_batch_size
+    if b == n:
+        return array
+    # torch.broadcast_to / torch.tile mirror jnp.broadcast_to / jnp.tile exactly
+    if is_torch_tensor(array):
+        _broadcast_to, _tile = torch.broadcast_to, torch.tile  # pyright: ignore[reportAttributeAccessIssue]
+    elif is_jax_array(array):
+        _broadcast_to, _tile = jnp.broadcast_to, jnp.tile  # pyright: ignore[reportAttributeAccessIssue]
+    else:
```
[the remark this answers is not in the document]
k-chrispens: Good point

src/sampleworks/utils/atom_reconciler.py
k-chrispens: I tried to remove it everywhere some time ago and pyright then complained about all my return annotations whenever I have a class method returning an instance of that class or operating on an instance of that class. ty still complains about it here.

Sounds like it will eventually be deprecated though, at python 3.13 EOL

src/sampleworks/utils/atom_reconciler.py
```diff
@@ -0,0 +1,265 @@
...
+        Copies common atoms from ``struct_coords`` into a clone of
+        ``model_template``. Differentiable through ``struct_coords``.
+
+        Parameters
+        ----------
+        struct_coords
+            Coordinates in structure space, shape ``(*batch, n_struct, 3)``.
+        model_template
+            Template in model space, shape ``(*batch, n_model, 3)``.
+            Non-common atom positions are preserved from this template.
```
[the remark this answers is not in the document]
k-chrispens: Yes, it will, but it will allow us to initialize coords from an incomplete initial structure. Sort of remains to be tested if this has a large effect, but hopefully the model figres it out! Gradients won't flow back to non-shared atoms through this op

src/sampleworks/core/samplers/protocol.py
k-chrispens: Yeah I will clean this up when I address #109. Structure space is the input structure, model space is the model's internal count of how the input structure should be represented as atoms

PR #114, opened 2026-02-27
chore(cleanup): remove checkpoint_utils.py

### LINKED ISSUE #70: Remove checkpoint_utils.py
checkpoint_utils.py introduces the get_checkpoint function, which isn't necessary and adds extra clutter when we could just take the checkpoint as an argument.

k-chrispens:
Closes #70 

## Summary by CodeRabbit

* **Refactor**
  * Replaced multiple model-specific checkpoint flags with a single unified --model-checkpoint CLI option; legacy per-model flags removed.

* **Bug Fixes**
  * Centralized checkpoint validation and normalization to produce clearer errors and consistent behavior when specifying model checkpoints.

* **Tests**
  * Added tests covering checkpoint argument handling, validation, and config population to ensure correct override and error cases.


marcuscollins (approved):
Easy.

marcuscollins (changes requested):

marcuscollins (changes requested):
Sorry, one more round of changes needed. I'm pretty sure for this to work, you will have to edit the `parse_args` method in `run_grid_search.py` to remove the model-specific checkpoint names (like `--boltz2-checkpoint`) and add only `--model-checkpoint`

marcuscollins (approved):
Some things to think about for the future: a lot of software developers (and, as a result, LLMs) will write a lot of short methods to do things, and lots and lots of validation. I think both can be overdone, and I'd say that happened here. The reason it is is a problem is that it obfuscates what the code actually does in multiple layers of function calls. We can talk about better ways to do it in the future.

Issue #76, opened 2026-02-05
Make grid search input file paths relative


marcuscollins:
Right now the input format read in at https://github.com/diff-use/sampleworks/blob/main/run_grid_search.py#L297 assumes the protein structure and density paths are absolute. We should make them relative to the input file directory, so find the `Path(args.proteins).parent` directory and pre-pend it to the paths read in from that file `args.proteins` so paths in that file are relative to its location. This will simplify deployment to different storage.

PR #131, opened 2026-03-05
feat(io): make structure and density paths in input protein csv file relative

DorisMai:
Address issue #76 

:exclamation: Warning: A patch in `qFit` source code is needed if we want to support both `Path` and `str` of `density` file. This PR adds this patch but also forces `density` to be `str` in `run_grid_search.py` for robustness.

## Summary by CodeRabbit

* **New Features**
  * Added CSV protein input parsing with automatic path resolution, validation of name/structure/density/resolution, and clearer error messages.

* **Refactor**
  * Accept both path and string inputs for structure/density and simplified protein data access.

* **Bug Fixes**
  * More robust handling of file-like inputs when deriving filenames; no user-facing API changes.

### DIFF
--- a/src/sampleworks/utils/protein_input.py
+++ b/src/sampleworks/utils/protein_input.py
@@ -1,4 +1,5 @@
 import csv
+import math
 from dataclasses import dataclass
 from pathlib import Path
 
@@ -27,9 +28,10 @@ class ProteinInput:
             raise FileNotFoundError(
                 f"Density file does not exist for protein '{self.name}': {self.density}"
             )
-        if self.resolution <= 0:
+        if not math.isfinite(self.resolution) or self.resolution <= 0:
             raise ValueError(
-                f"Resolution must be positive for protein '{self.name}', got {self.resolution}."
+                f"Resolution must be a positive finite number for protein '{self.name}', "
+                f"got {self.resolution}."
             )
 
     @classmethod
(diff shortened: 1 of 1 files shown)


run_grid_search.py
```diff
@@ -295,8 +296,8 @@ def generate_jobs(args: argparse.Namespace) -> list[JobConfig]:
     methods = [m.strip() for m in args.methods.split(",")]
 
     for protein in proteins:
-        structure = protein["structure"].strip()
-        density = protein["density"].strip()
+        structure = proteins_dir / protein["structure"].strip()
```
marcuscollins: We should probably allow absolute and relative paths. I believe Path has a method to check if a path is already absolute, so please test that and decide whether it is necessary to complete the path or not.

marcuscollins (dismissed):

run_grid_search.py
DorisMai: should be addressed in the new data class construction `from_csv`

src/sampleworks/core/forward_models/xray/real_space_density_deps/qfit/volume.py
k-chrispens: just as a general comment: these files in `qfit/` shouldn't be modified unless really necessary since they are ported over from qfit upstream. Ideally we eventually get rid of these though

PR #140, opened 2026-03-05
fix(rf3): squashed RF3 bug that was blocking running on models with unresolved residues

### LINKED ISSUE #86: Make sure our structure outputs don't contain nan coordinates.
At present, our output (in the method `save_everything` https://github.com/diff-use/sampleworks/blob/64a1b963d96f9b2c14f86935e74c955a5bb73077/src/sampleworks/utils/guidance_script_utils.py#L236 appears to include hydrogens with nan coordinates. We should explicitly make sure these are removed to prevent problems with other programs (phenix, pepp'r, etc...) that read in CIF files.

k-chrispens:
This also removed the reward_param_mask as it isn't needed anymore if we are computing common atoms and saving the atoms that the model is operating on instead of having to reconcile and save the original structure. This should close #86 as well.

## Release Notes

* **New Features**
  * Enhanced input validation for atom arrays with checks for element presence, b-factor presence, finite coordinates, and valid occupancy values.
  * Improved handling of NaN coordinates with automatic centroid-based noise generation during model initialization.

* **Refactor**
  * Simplified coordinate preprocessing pipeline by removing unnecessary masking complexity.
  * Streamlined API by removing internal parameters from core utility functions.


marcuscollins (commented):
Just a couple questions to be sure of things. If you can answer them, I think it will be fine and I will approve.

marcuscollins (approved):
Questions answered, thanks. Approved.

Issue #122, opened 2026-03-02
Rename classes to avoid confusion with hub.diffuse.science


marcuscollins:
In particular, we have the class `sampleworks.eval.eval_dataclasses.Experiment` which corresponds to all the metadata for an output CIF file. Related to https://github.com/diff-use/sampleworks/issues/121

PR #142, opened 2026-03-07
refactor(eval): rename Experiment → Trial

DorisMai:
Resolves issue #122 . The name "Experiment" conflicted with the term used in hub.diffuse.science. To avoid confusion, the Experiment class is renamed to Trial, which holds all the metadata associated with a single run in grid search.

Renames across all eval scripts and data classes:
- Experiment → Trial, ExperimentList → TrialList
- exp_dir → trial_dir, parse_experiment_dir → parse_trial_dir
- all_experiments → all_trials, exp/experiment → trial
- process_one_experiment → process_one_trial
- process_exp_with_selection → process_trial_with_selection

## Summary by CodeRabbit

* **Refactor**
  * Terminology unified from “experiments” to “trials” across evaluation tools, user-facing logs, summaries and public APIs to present consistent wording to users.
* **Tests**
  * Added tests validating trial-related data handling and summary behavior: field access and defaults, nullable handling, copy semantics, list operations, and summary reporting.

### DIFF
--- a/scripts/eval/bond_geometry_eval.py
+++ b/scripts/eval/bond_geometry_eval.py
@@ -209,7 +209,7 @@ def main(args: argparse.Namespace):
     logger.info(f"Grid search directory: {grid_search_dir}")
 
     # Scan for trials (look for refined.cif files)
-    all_trials = scan_grid_search_results(grid_search_dir)
+    all_trials = scan_grid_search_results(grid_search_dir, target_filename=args.target_filename)
     logger.info(f"Found {len(all_trials)} trials with refined.cif files")
 
     if all_trials:
--- a/scripts/eval/lddt_evaluation_script.py
+++ b/scripts/eval/lddt_evaluation_script.py
@@ -208,7 +208,7 @@ def main(args: argparse.Namespace):
     logger.info(f"Proteins configured: {list(protein_configs.keys())}")
 
     # Scan for trials (look for refined.cif files)
-    all_trials = scan_grid_search_results(grid_search_dir)
+    all_trials = scan_grid_search_results(grid_search_dir, target_filename=args.target_filename)
     logger.info(f"Found {len(all_trials)} trials with refined.cif files")
 
     if all_trials:


scripts/eval/rscc_grid_search_script.py
```diff
@@ -180,20 +180,20 @@ def main(args: argparse.Namespace):
...
+                            f"Could not find reference structure for occupancy {_trial.occ_a}"
                         )
 
                     # 2. Load the reference structure with parse() to get only the first altloc
                     ref_structure = parse(ref_path, ccd_mirror_path=None)
                     ref_atom_array = get_asym_unit_from_structure(ref_structure)
-                    logger.info(f"Caching reference structure for {protein} occ_a={_exp.occ_a}")
-                    ref_full_structure_cache[(protein, _exp.occ_a)] = ref_atom_array
+                    logger.info(f"Caching reference structure for {protein} occ_a={_trial.occ_a}")
+                    ref_full_structure_cache[(protein, _trial.occ_a)] = ref_atom_array
```
marcuscollins: Could you get rid of these leading underscores while we're at it? They're a holdout from this originally being in a marimo notebook, and don't really make sense here.

DorisMai:
> LGTM, is there any reason to leave it as a draft?
> 
> Please make sure you've run all the tests. We'll update that soon, as we'll have the ability to run them when PRs are created/updated, but for now please make sure everything runs. As we do this PR, it might be good to add some simple tests for these two classes as well.

I left it as draft thinking it wouldn't notify people on weekend...nvm I've change it to ready for review.

I did ran tests in the boltz-dev environment and nothing failed, but quite some tests were skipped due to issues like to no model checkpoints.  I ll add some simple tests for these two classes.

DorisMai:
> @DorisMai to run all tests, I typically run the pixi task I configured `pixi run test-all | tee pytest.log`. That way it will run each model through, and it will stop if any tests fail with a particular model (though that should probably be fixed)

@k-chrispens I think the difficulty I am encountering now is with respect to model checkpoints. Are these expected to in a shared folder accessible to everyone? Or is this expected to be downloaded when set up environment / installing dependencies? It looks like the paths are hardcoded like `~/.boltz/boltz2_conf.ckpt` in `conftest.py`.

k-chrispens:
Yes, you will need to download. I added this to the new readme, which I am PRing shortly. You could also run boltz predict in the boltz environment on some random sequence and it will auto download.

Additionally I noticed that you did a merge commit on your other PR #131, which will block it from being merged (it breaks linear history). If you use the default option to update to the latest main it will do this, so you will have to press the arrow and use rebase instead. Here, it says that there are conflicts, which are also blocking things. You will need to do something like `git fetch && git rebase origin/main` to get the latest history. This should also help highlight how to fix the conflicts!

Issue #160, opened 2026-03-13
Fix logging of occupancies in rscc_grid_search_script.py


DorisMai:
_⚠️ Potential issue_ | _🔴 Critical_

**Pass an occupancy mapping into `get_base_map_path_for_occupancy()`.**

`occ` is a scalar float from `args.occupancies`, but `ProteinConfig.get_base_map_path_for_occupancy()` expects `dict[str, float]`. This preflight block now raises before the scan starts.

<details>
<summary>🐛 Minimal fix</summary>

```diff
     logger.debug("Testing base map path resolution:")
     for _, config in protein_configs.items():
         for occ in args.occupancies:
-            path = config.get_base_map_path_for_occupancy(occ)  # will warn if not found
+            altloc_occ = {"A": occ, "B": 1.0 - occ}
+            path = config.get_base_map_path_for_occupancy(altloc_occ)  # will warn if not found
             if path:
-                logger.debug(f"  {config.protein} occ={occ}: {path}")
+                logger.debug(f"  {config.protein} occ={altloc_occ}: {path}")
```
</details>

<details>
<summary>🤖 Prompt for AI Agents</summary>

```
Verify each finding against the current code and only fix it if needed.

In `@scripts/eval/rscc_grid_search_script.py` around lines 66 - 69, The loop is
passing a scalar occ from args.occupancies into
ProteinConfig.get_base_map_path_for_occupancy(), which expects a dict[str,
float]; wrap the scalar into a mapping before calling the method (e.g. call
config.get_base_map_path_for_occupancy({config.protein: occ})), updating the
call site that uses occ in the preflight loop (the for occ in args.occupancies
block) so the function receives the expected dict[str, float].
```

</details>

_Originally posted by @coderabbitai[bot] in https://github.com/diff-use/sampleworks/pull/142#discussion_r2915236150_

PR #169, opened 2026-03-14
fix(log): pass and log altloc_occ dict rather than float to get base map

DorisMai:
Address issue #160 

## Summary by CodeRabbit

* **Refactor**
  * Occupancy handling in grid-search evaluation updated to use explicit A/B occupancy mapping; logging updated to reflect the new mapping.

* **Tests**
  * Added tests covering base-map path resolution for varied occupancy scenarios: existing/missing maps, zero-occupancy pruning, all-zero and over-sum occupancies triggering warnings.


marcuscollins (approved):
Thanks for the extra tests! LGTM.

PR #182, opened 2026-03-20
feat(rf3): chiral gradient zeroing and tracking

k-chrispens:
Chiral gradients may be why we saw some geometry problems with RF3. This adds the ability to track them for guided and non-guided runs, as well as zero them out so that we can make sure we aren't breaking geometry with those.

## Summary by CodeRabbit

* **New Features**
  * Added options to disable chiral inputs and to track per-step chiral diagnostics for RF3 models.
  * New CLI flags to enable/disable these behaviors.
  * Featurization now resets per-run chiral state, captures original chiral inputs, and preserves model–atom mapping for reconciliation.
  * When tracking is enabled, per-step chiral gradient metrics are recorded and written to chiral_grad_stats.json.

* **Tests**
  * Added end-to-end tests covering chiral feature handling, disabling behavior, and diagnostics collection.

### DIFF
--- a/tests/models/test_rf3_chiral_features.py
+++ b/tests/models/test_rf3_chiral_features.py
@@ -3,7 +3,11 @@
 import pytest
 import torch
 from sampleworks.core.samplers.edm import AF3EDMSampler, EDMSamplerConfig
-from sampleworks.models.rf3.wrapper import annotate_structure_for_rf3
+from sampleworks.utils.imports import require_rf3, RF3_AVAILABLE
+
+
+if RF3_AVAILABLE:
+    from sampleworks.models.rf3.wrapper import annotate_structure_for_rf3
 
 
 def _make_sampler(device: torch.device) -> AF3EDMSampler:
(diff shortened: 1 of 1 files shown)


src/sampleworks/models/rf3/wrapper.py
```diff
@@ -488,6 +547,43 @@ def step(
...
 
+        # Track chiral gradient statistics using original features to report what the chiral
+        # gradient would have been for diagnostics
+        if (
+            self._track_chiral_features
+            and self._original_chiral_centers is not None
+            and self._original_chiral_dihedral_angles is not None
+            and self._original_chiral_centers.shape[0] > 0
+        ):
+            from rf3.loss.loss import calc_chiral_grads_flat_impl
```
marcuscollins: Slightly more than a nit: Is there a reason not to import this sooner? I personally don't like inline imports, because there can be a (admittedly probably small in this case) performance hit when you go to re-import.

PR #195, opened 2026-03-28
Fix problem with heterogeneous HETATM/ATOM in the same position in different altlocs

marcuscollins:
## Summary by CodeRabbit

* **Bug Fixes**
  * Detects and resolves mixed ATOM/HETATM entries at the same residue positions, removing conflicting HETATM atoms and returning a corrected structure path when changes occur; logs warnings identifying affected residues and modified names.

* **Integration**
  * Structure loading now pre-processes inputs with the resolver and removes any temporary resolved files after use.

* **Tests**
  * Added comprehensive tests for clean, mixed, standalone ligand, multi-mixed, and real-file cases, including warning assertions.

* **Chores**
  * Adjusted evaluation script output locations to write results into the configured results path.

### DIFF
--- a/scripts/eval/run_and_process_tortoize.py
+++ b/scripts/eval/run_and_process_tortoize.py
@@ -40,11 +40,11 @@ def main(args: argparse.Namespace) -> None:
 
     all_residue_results, all_protein_results = tuple(zip(*tortoize_results, strict=True))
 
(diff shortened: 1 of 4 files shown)


k-chrispens (approved):
Looks like a solid fix, the tests could take another pass and the coderabbit comment should be addressed, but will approve and let you take care of that

PR #203, opened 2026-04-07
fix(ci): run semantic-release as CLI instead of Docker action

Abdelsalam-Abbas:
## Summary
- Replaces the Docker-based `python-semantic-release` action with direct CLI invocation via `uvx`
- The Docker container didn't have pixi, so `build_command = "pixi lock"` failed with `command not found`
- Running on the host means `build_command` can use pixi (installed by `setup-pixi`)
- Adds `assets = ["pixi.lock"]` so the regenerated lockfile is staged in the version commit

## How it works
The semantic-release lifecycle: stamp version in pyproject.toml -> run `build_command` (`pixi lock`) -> stage version files + `assets` (`pixi.lock`) -> commit -> tag -> push. This means the lockfile update is part of the same commit and tag as the version bump.

## Test plan
- [ ] Merge and verify next release commit includes updated pixi.lock
- [ ] Verify `semantic-release publish` creates the GitHub Release correctly with `GH_TOKEN`

## Summary by CodeRabbit

* **Chores**
  * Updated release workflow and included additional artifacts in release packages.


marcuscollins (approved):
LGTM

PR #204, opened 2026-04-08
Speed up Docker build: free disk space, drop broken cache

Abdelsalam-Abbas:
## Summary
- Add `jlumbroso/free-disk-space` action to reclaim ~30 GB on the runner, preventing OOM/disk-full crashes
- Remove GHA cache (`type=gha`) which was counterproductive — the 10 GB limit caused unpredictable cache evictions on our ~20 GB image, leading to build times ranging 25-71 min
- Split pixi install into separate RUN layers per environment for better local cache reuse
- Move checkpoint COPY before pixi installs so the ~10 GB checkpoint layer stays cached when only dependencies change

Tested builds on this branch: **12-20 min** (runner variance) vs baseline **28 min average** (up to 71 min with cache thrashing). No more disk-full failures.

Closes https://app.clickup.com/t/86e0pc15n

## Test plan
- [x] Verified build succeeds with free-disk-space on `ubuntu-latest` (runs #24156031787, #24159607975)
- [x] Confirmed GHA cache removal doesn't regress build time (cache was net-negative)
- [x] Tested registry cache alternative — runner disk too small for cache export, not viable

## Summary by CodeRabbit

* **Chores**
  * Added automatic disk cleanup step on CI runners to prevent storage-related build failures.
  * Removed CI-level cache integration for Docker builds to simplify caching behaviour.
  * Reorganized image build steps and split dependency installs into separate layers to reduce unnecessary rebuilds.
  * Kept the final build output message showing the pushed image digest.


marcuscollins (approved):
@Abdelsalam-Abbas just left a couple comments/questions, but LGTM.

Abdelsalam-Abbas (commented):
Nit: the image digest echo was changed from "pushed" to "built", but since `push: true` is set, "pushed" is the correct wording. This should stay as-is (or be reverted to the original).

PR #205, opened 2026-04-09
Add configurable recycling steps and diffusion step count parameters

marcuscollins:
This PR adds two options to control diffusion: the total number of steps, and the number of recycling steps. These were used previously but with hard-coded values which are now surfaced to the user. They are passed to the respective models by use of the annotate* methods which append keyword config values to the structure dictionary. 

## Summary by CodeRabbit

* **New Features**
  * Added CLI options to set recycling steps and override diffusion-step count (default 200).

* **Bug Fixes**
  * Validation for recycling and diffusion-step values; invalid values now raise clear errors.
  * Backend dispatch generalized so RF3, Boltz, and Protenix backends receive and honor recycling/diffusion settings; unknown backends produce explicit errors.

* **Chores**
  * Cleaned up inline type-check comments and normalized None recycling to a sensible default for Boltz.

### DIFF
--- a/run_grid_search.py
+++ b/run_grid_search.py
@@ -450,7 +450,15 @@ def parse_args() -> argparse.Namespace:
         "--recycling-steps",
         type=int,
         default=None,
-        help="Number of recycling steps for model inference (if not specified, uses model default)",
+        help="Number of recycling steps for model inference. If not specified, "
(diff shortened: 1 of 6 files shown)


k-chrispens (approved):
Generally lgtm after addressing ty comments, there are a little comment things that might be useful though!

PR #207, opened 2026-04-09
feat: Add unified CLI for run_guidance and fix missing args bug

### LINKED ISSUE #153: Document how to use the `run_guidance` method in guidance_script_utils
This is now the main interface to actually run guidance trials, so we should create documentation on how to use it.

### LINKED ISSUE #154: Add a CLI interface for run_guidance() in guidance_script_utils.py
This function is the basis for an individual run on a single structure, and so we should expose this more accessibly.

### LINKED ISSUE #181: run_guidance when run with scripts exits non-zero because the protein, model, and guidance_type args are never added
The entrypoint to run a given guidance method with a specific model (not a grid search) uses run_guidance, which doesn't fully populate the argparse.Namespace unlike when GuidanceConfig is passed during a grid search. This should be addressed with #154 and #153.

Abdelsalam-Abbas:
## Summary

- **Fixes #181**: `run_guidance` scripts crash with `AttributeError: 'Namespace' object has no attribute 'protein'` because `protein`, `model`, and `guidance_type` were never added to the argparse parser. The error occurs in `get_job_result()` after guidance completes (or fails), making it impossible to get a clean exit from standalone scripts.
- **Fixes #154**: Adds `sampleworks-guidance` as a unified CLI entrypoint (registered via `console_scripts`) that replaces the need for 8 separate model+guidance scripts.
- **Fixes #153**: Documents the new CLI in the README Quick Start section with a reference table of required arguments.

### What changed

- Added `GuidanceConfig.from_cli()` classmethod that handles two-pass argument parsing: first resolves `--model` and `--guidance-type`, then builds the full parser with model-specific and guidance-type-specific args dynamically.
- Created `src/sampleworks/cli/guidance.py` as the unified entrypoint.
- Simplified all 8 standalone scripts to thin wrappers calling `from_cli()` with pre-set model/guidance_type.
- Removed the 8 `parse_*_args()` functions (replaced by `from_cli()`).
- Added `--protein` as a required argument (must match naming used in grid search/evaluation).
- Registered `sampleworks-guidance` console_scripts in `pyproject.toml`.
- Propagate `run_guidance` exit code in all 8 wrapper scripts via `sys.exit(main())` so CI and automation see non-zero exits on failure (previously always exited 0).
- Stop registering `--model`/`--guidance-type` on the argparse parser when preset by wrapper scripts, so conflicting CLI input (e.g. `rf3_fk_steering.py --model boltz2`) is rejected instead of silently ignored.

## Test plan

- [x] 22 new tests in `tests/cli/test_guidance_cli.py` covering unified CLI, legacy script pattern, validation errors, and defaults
- [x] All 678 existing tests pass (164 skipped)
- [x] Manually reproduced #181 on main (`AttributeError: 'Namespace' object has no attribute 'protein'`, EXIT CODE: 1)
- [x] Verified fix on branch: same script exits cleanly (EXIT CODE: 0) with `GuidanceConfig` properly populating `protein`, `model`, `guidance_type`
- [ ] Run GPU end-to-end test with a valid CCD/checkpoint setup to verify full guidance completion

## Summary by CodeRabbit

* **New Features**
  * Added a unified CLI command for guidance workflows with explicit model, guidance type, protein, and experimental-data inputs; registered as a console entrypoint.

* **Refactor**
  * Consolidated legacy script entrypoints into a single CLI-based approach with centralized configuration parsing.

* **Documentation**
  * Updated Quick Start to use the unified command and added a CLI reference with required arguments and discoverable options.

* **Tests**
  * Added comprehensive CLI parsing tests covering modes, validation, and edge cases.


src/sampleworks/utils/guidance_script_arguments.py
```diff
@@ -112,9 +112,44 @@ def validate_model_checkpoint(
     return str(checkpoint_path)
 
 
+_MODEL_ARG_ADDERS: dict[str, Any] = {
+    "boltz1": lambda p: add_boltz1_specific_args(p),
```
marcuscollins: Do you have a reason to use lambdas here? Why not use, say,
```
"boltz1": add_boltz1_specific_args,
```
You'd still call it exactly the same way: `_MODEL_ARG_ADDERS[model](parser)`

src/sampleworks/utils/guidance_script_arguments.py
```diff
@@ -112,9 +112,44 @@ def validate_model_checkpoint(
...
+    "boltz2": lambda p: add_boltz2_specific_args(p),
+    "protenix": lambda p: add_protenix_specific_args(p),
+    "rf3": lambda p: add_rf3_specific_args(p),
+}
+
+_GUIDANCE_ARG_ADDERS: dict[str, Any] = {
+    "pure_guidance": lambda p: add_pure_guidance_args(p),
```
marcuscollins: Same question about lambdas.

src/sampleworks/utils/guidance_script_arguments.py
```diff
@@ -143,6 +178,108 @@ def add_argument(self, name: str, default: Any = None, **kwargs):
...
+            if not guidance_preset:
+                pre.add_argument(
+                    "--guidance-type",
+                    type=str,
+                    required=True,
+                    choices=guidance_choices,
+                    help="Guidance method",
+                )
+            pre_args, _ = pre.parse_known_args(argv)
+            model = model or pre_args.model
```
marcuscollins, line 217: I feel like this should raise an error. If someone is running, say, the boltz2_pure_guidance.py script and includes the --model flag, at least if the specified value isn't boltz2, that ought to fail.

src/sampleworks/utils/guidance_script_arguments.py
```diff
@@ -143,6 +178,108 @@ def add_argument(self, name: str, default: Any = None, **kwargs):
...
+            guidance_type = guidance_type or pre_args.guidance_type
```
marcuscollins, line 218: Likewise for guidance_type--if the value is set already (`guidance_type is not None`) and the user specifies a value that disagrees with whatever `guidance_type` is, it should raise a RuntimeError or ValueError. I shouldn't be able to use the `boltz1_fk_steering.py` script to run the rf3 model or pure guidance.

src/sampleworks/utils/guidance_script_arguments.py
marcuscollins: Could you  use your _MODEL_ARG_ADDERS here? And the _GUIDANCE_ARG_ADDERS above?

scripts/boltz1_fk_steering.py
```diff
@@ -1,19 +1,21 @@
...
+        getattr(config, "model_checkpoint", None),
+        config.model,
     )
-    run_guidance(args, GuidanceType.FK_STEERING, model_wrapper, device)
+    result = run_guidance(config, config.guidance_type, model_wrapper, device)
+    return result.exit_code
 
 
 if __name__ == "__main__":
-    guidance_args = parse_boltz1_fk_steering_args()
```
marcuscollins, line 18: What happens now if I run `pixi run -e boltz python boltz1_fk_steering.py --help`? Will I get the usual command documentation? I would like to make sure that still works.

scripts/boltz1_fk_steering.py
```diff
@@ -1,19 +1,21 @@
...
 
 
-def main(args):
+def main():
+    config = GuidanceConfig.from_cli(model="boltz1", guidance_type="fk_steering")
     device, model_wrapper = get_model_and_device(
-        "", args.model_checkpoint, StructurePredictor.BOLTZ_1
+        config.device,
+        getattr(config, "model_checkpoint", None),
+        config.model,
```
marcuscollins: This goes with some comments below: I don't think it makes sense to run a script `boltz1_fk_steering.py` and use any other model than Boltz1 or any steering method besides FK_STEERING.

src/sampleworks/cli/guidance.py
```diff
@@ -0,0 +1,24 @@
+"""Unified CLI for running diffusion guidance with sampleworks."""
```
marcuscollins, line 1: We should ask @k-chrispens what he thinks, but I think you could get rid of all the model- and guidance-type-specific scripts (like `boltz2_pure_guidance.py`) and just use this one.

tests/cli/test_guidance_cli.py
```diff
@@ -0,0 +1,209 @@
...
+    "--density",
+    "test.ccp4",
+    "--resolution",
+    "1.8",
+    "--output-dir",
+    "output",
+]
+
+
+class TestFromCliUnified:
```
marcuscollins, line 26: Please add a test/tests to check that argument parsing fails/raises an error (I assume it should) if someone uses a model-specific argument (like --method for Boltz or --msa-path for RF3) with the wrong type of model (e.g. they use --method with Protenix)

marcuscollins (changes requested):

src/sampleworks/cli/guidance.py
k-chrispens: I agree with this, I had actually thought there was already an issue for this but ther's only the CLI one. we should just use a single CLI script for all the "one-off" guidance runs

src/sampleworks/utils/guidance_script_arguments.py
Abdelsalam-Abbas: Done, passing a conflicting --model now raises a clear error:
`boltz2_pure_guidance.py: error: This script is fixed to --model boltz2. Use sampleworks-guidance for other models.`

src/sampleworks/utils/guidance_script_arguments.py
Abdelsalam-Abbas: Good call, replaced with direct function refs.

src/sampleworks/utils/guidance_script_arguments.py
Abdelsalam-Abbas: Good call, replaced with direct function refs.

src/sampleworks/utils/guidance_script_arguments.py
Abdelsalam-Abbas: Handled alongside `--model` in the same commit. Both are validated when preset.

src/sampleworks/utils/guidance_script_arguments.py
Abdelsalam-Abbas: Done

scripts/boltz1_fk_steering.py
Abdelsalam-Abbas: Verified.
```
usage: boltz1_fk_steering.py [-h] --protein PROTEIN --structure STRUCTURE


  -h, --help            show this help message and exit
  --protein PROTEIN     Protein identifier (must match naming used in grid
  --structure STRUCTURE
  --density DENSITY     Input density map
  --output-dir OUTPUT_DIR
  --log-path LOG_PATH   Log file path (default: output-dir/run.log)
  --partial-diffusion-step PARTIAL_DIFFUSION_STEP
  --loss-order {1,2}    L1 or L2 loss
  --resolution RESOLUTION
  --device DEVICE       Device (cuda/cpu, auto-detect)
  --gradient-normalization
  --em                  Use EM scattering factors
  --guidance-start GUIDANCE_START
  --augmentation        Enable data augmentation
  --align-to-input      Enable alignment to input
  --ensemble-size ENSEMBLE_SIZE
  --model-checkpoint MODEL_CHECKPOINT
  --num-particles NUM_PARTICLES
  --fk-resampling-interval FK_RESAMPLING_INTERVAL
  --fk-lambda FK_LAMBDA
  --num-gd-steps NUM_GD_STEPS
  --guidance-weight GUIDANCE_WEIGHT
  --guidance-interval GUIDANCE_INTERVAL
```
 --help shows all boltz1+fk_steering args. --model/--guidance-type are hidden from help but still validated if passed.

scripts/boltz1_fk_steering.py
Abdelsalam-Abbas: Done

src/sampleworks/cli/guidance.py
Abdelsalam-Abbas: Agree they're now thin wrappers. Let's get @k-chrispens  input.

tests/cli/test_guidance_cli.py
Abdelsalam-Abbas: added more tests

src/sampleworks/cli/guidance.py
Abdelsalam-Abbas: Done. and updated docs

k-chrispens (commented):
I think the one remaining change to make is to expose the new arguments that Marcus enabled in #205 to the CLI, then it's good to go! It works on my dev node when I run things. I'll let @marcuscollins come back to approve if he things his comments were addressed fully

Abdelsalam-Abbas:
> I think the one remaining change to make is to expose the new arguments that Marcus enabled in #205 to the CLI, then it's good to go! It works on my dev node when I run things. I'll let @marcuscollins come back to approve if he things his comments were addressed fully

exposed both arguments ` --recycling-steps` and `--num-diffusion-steps`

marcuscollins (approved):
LGTM. Thanks for the changes @Abdelsalam-Abbas !

PR #217, opened 2026-04-11
Add occupancy validation with proper error handling

ManjunathByadagi:
### Changes
- Added validation to ensure occupancy values are numeric
- Ensured values are within valid range (0.0 to 1.0)
- Added clear error messages for invalid inputs

### Why
Previously, invalid occupancy values were not handled properly.
This update ensures proper validation and prevents incorrect data processing.

### Example

**Valid:**
0.5, 1.0

**Invalid:**
-1, 2 → will raise error

## Summary by CodeRabbit

* **Refactor**
  * Streamlined CIF patching utility for a simpler, more maintainable processing flow.

* **Bug Fixes**
  * Improved validation and handling of atomic coordinates and occupancy/B_iso values to reduce malformed output and surface clearer error messages.

* **Other**
  * Patched files are saved with a "-patched.cif" suffix.

### DIFF
--- a/scripts/patch_output_cif_files.py
+++ b/scripts/patch_output_cif_files.py
@@ -15,7 +15,6 @@ from loguru import logger
 from sampleworks.utils.atom_array_utils import remove_atoms_with_any_nan_coords
 from sampleworks.utils.cif_utils import add_category_to_cif
 
-
 SAMPLEWORKS_CACHE = Path("~/.sampleworks/rcsb").expanduser()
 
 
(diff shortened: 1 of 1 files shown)


ManjunathByadagi:
Thanks for the review! I've addressed the comments and updated the PR. Please let me know if anything else is needed.

k-chrispens:
Hi! I know we haven't really added contribution guidelines yet, but something we plan to require generally is that new contributor PRs are attached to an existing issue (though exceptions will apply for new ModelWrappers, etc.). For example, we haven't included an issue for this because the existing patch script is very hacky and will likely be replaced entirely in the future, so this type of QC adds review burden on us for changes that are likely not to be useful in the long run.

I encourage you to take a look at the existing issues and try out one of those instead. Closing this PR.

PR #220, opened 2026-04-13
test(msa): add MSAManager unit tests (#178)

### LINKED ISSUE #178: Add tests for MSAManager
This class is currently untested, but is a critical piece of infrastructure. 

We should test the following (presumably with a mock for the actual API)
1. it calls `_compute_msa` when _not all_ the required files are present.
2. `_compute_msa` calls the correct API
3. `_compute_msa` correctly writes out the .a3m and .csv files and they contain the same output (see `_validate_msa_cache_contents` in msa.py for some good checks, on branch `fix-path-serialization-bug` if it hasn't yet been merged.
4. It correctly creates a cache directory, or uses one that is specified (test both)
5. the hash function returns the correct values for the same ordered set of sequences

Abdelsalam-Abbas:
## Summary
Adds [tests/utils/test_msa.py](tests/utils/test_msa.py) covering the five tests requested in #178:

1. `_hash_arguments` is deterministic for identical inputs and sensitive to sequence content, pairing strategy, and value order.
2. Explicit `msa_cache_dir` is used (both `Path` and `str` forms); default `None` creates `~/.sampleworks/msa` (with `Path.home` monkeypatched to `tmp_path` so the real home is untouched).
3. `get_msa` calls `_compute_msa` when cache files are missing, and skips it on the second call when they exist. Cache-hit/api-call counters are asserted.
4. `_compute_msa` forwards the expected arguments to `run_mmseqs2` for both the paired and unpaired branches (2-sequence input so both fire), including `auth_headers` built from `api_key_header` / `api_key_value`.
5. `_compute_msa` writes matching `.csv` and `.a3m` files: the second column of the csv equals the odd (sequence) lines of the a3m.

## Notes
- Mocks patch `sampleworks.utils.msa.run_mmseqs2` and `sampleworks.utils.msa._compute_msa` (not the source modules) so the actual call sites in `msa.py` are intercepted — `run_mmseqs2` is imported at module load, and `_compute_msa` is module-level.
- Test 3 uses a `side_effect` that writes a valid csv/a3m pair so the unconditional `_validate_msa_cache_contents` call inside `get_msa` passes, doubling as a smoke check that the cache layout `_compute_msa` writes matches what `get_msa` expects.
- Protenix branch is intentionally not tested per the issue; no new dependency on `PROTENIX_AVAILABLE`.

## Test plan
- [x] `pixi run -e boltz-dev python3 -m pytest tests/utils/test_msa.py -v` → 7 passed in 0.03s
- [x] `pixi run -e boltz-dev all-tests` → 748 passed, 509 skipped, no regressions

## Summary by CodeRabbit

* **Tests**
  * Added a comprehensive test suite for MSA utilities covering deterministic input hashing, cache behavior and directory initialization, correct triggering of MSA computation when cache entries are missing, and subsequent cache hits. Also validates integration with the external alignment tool (correct call patterns and flags), temp-directory usage, and output consistency between generated alignment files.

### DIFF
--- a/tests/utils/test_msa.py
+++ b/tests/utils/test_msa.py
@@ -1,0 +1,189 @@
+"""Tests for MSAManager and the MSA cache helpers."""
+
+from __future__ import annotations
+
+from pathlib import Path
+from unittest.mock import patch
+
+import pytest
+from sampleworks.utils import msa as msa_module
+from sampleworks.utils.guidance_constants import StructurePredictor
+from sampleworks.utils.msa import _compute_msa, MSAManager
+
+
+# ============================================================================
+# _hash_arguments (test 5)
+# ============================================================================
+
+
+def test_hash_arguments_is_deterministic_and_input_sensitive():
+    data = {"A": "MKTAY", "B": "GGGAA"}
+    base = MSAManager._hash_arguments(data, "greedy")
+
+    assert base == MSAManager._hash_arguments({"A": "MKTAY", "B": "GGGAA"}, "greedy")
+
+    # Different sequence -> different hash.
+    assert base != MSAManager._hash_arguments({"A": "MKTAY", "B": "GGGAC"}, "greedy")
+    # Different pairing strategy -> different hash.
+    assert base != MSAManager._hash_arguments(data, "complete")
+    # Different value order -> different hash (hash is over tuple(values)).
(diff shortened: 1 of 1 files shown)


k-chrispens (approved):
This looks correct to me, though I don't see where the original issue said to not add Protenix branch related tests. I believe the tests requested in the issue behave similarly for either branch, so will approve here (differs just what server is called and how things are linked up for Protenix).

PR #222, opened 2026-04-14
fix(rf3): bind Fabric to caller-specified GPU

### LINKED ISSUE #37: For ModelWrappers that use Fabric, etc., figure out how to properly handle device
The RF3 wrapper uses Lightning Fabric, which will set up the model on its own device. I haven't yet figured out how to select the device manually, but this is something that needs to be fixed at some point.

saada:
Closes #37.

RF3Wrapper didn't accept a `device` argument, so the grid-search dispatcher's `cuda:N` assignment was silently dropped. Lightning Fabric's default `devices=1` always resolves to GPU 0, which meant eight parallel workers all piled onto `cuda:0`.

The wrapper now takes a `device` and forwards the CUDA index to Fabric as `devices_per_node=[idx]`, matching the Boltz and Protenix wrappers.

### Reproduction & verification

Two RF3 wrappers launched in parallel inside the same container (different processes, no `CUDA_VISIBLE_DEVICES` hacks):

**Before** — both pinned to the same GPU:
```
[worker0] wrapper.device=cuda:0  current_device=0
[worker1] wrapper.device=cuda:0  current_device=0
```

**After** (passing `device="cuda:0"` and `device="cuda:3"`):
```
[worker0] target=cuda:0  wrapper.device=cuda:0  current_device=0
[worker1] target=cuda:3  wrapper.device=cuda:3  current_device=3
```

Unit + regression tests:
```
tests/models/test_rf3_device_selection.py ....... 7 passed
```

## Summary by CodeRabbit

* **New Features**
  * RF3Wrapper now accepts an optional device parameter to bind inference to a specific CUDA GPU for improved multi-GPU control.

* **Bug Fixes**
  * Device selection logic made consistent so the originally requested device is respected during model setup.

* **Tests**
  * Added GPU-focused tests validating CUDA device selection and wrapper device binding (skipped when GPUs unavailable).

### DIFF
--- a/src/sampleworks/models/rf3/wrapper.py
+++ b/src/sampleworks/models/rf3/wrapper.py
@@ -199,8 +199,8 @@ def _cuda_index(device: torch.device | str) -> int:
     Raises
     ------
     ValueError
-        If ``device`` is not a CUDA device. RF3's Fabric-backed accelerator
-        only supports GPU execution.
+        If ``device`` is not a CUDA device. Currently, sampleworks only supports
+        CUDA systems, so non-CUDA devices will fail here.
     """
     dev = torch.device(device)
     if dev.type != "cuda":
@@ -230,6 +230,10 @@ class RF3Wrapper:
             parallel jobs that must target distinct GPUs — passing an ``int``
             to Fabric (the default) always resolves to GPU 0, which serialises
             otherwise-parallel workers onto a single device.
+            
+            References: https://lightning.ai/docs/fabric/stable/fundamentals/launch.html
+              devices argument to fabric run
+            https://github.com/RosettaCommons/foundry/blob/b071919caa19ff334bc04b1b41145cac61eba819/src/foundry/trainers/fabric.py#L92
         """
         logger.info("Loading RF3 Inference Engine")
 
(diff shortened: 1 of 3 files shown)


src/sampleworks/models/rf3/wrapper.py
```diff
@@ -198,17 +224,27 @@ def __init__(
             Filesystem path to the checkpoint containing trained weights.
         msa_manager: MSAManager | None
             MSA manager for retrieving MSAs for input structures.
+        device: torch.device | str | None
+            CUDA device to bind the underlying Lightning Fabric to (e.g. ``"cuda:3"``).
```
k-chrispens, line 228: To make sure I understood this, I went searching in the foundry repo and found this line: https://github.com/RosettaCommons/foundry/blob/b071919caa19ff334bc04b1b41145cac61eba819/src/foundry/trainers/fabric.py#L92

Would probably be worth referencing this here for posterity

src/sampleworks/models/rf3/wrapper.py
```diff
@@ -183,13 +183,39 @@ def add_msa_to_chain_info(chain_info: dict, msa_path: str | Path | dict | None)
...
+    Returns
+    -------
+    int
+        Device index (``0`` when unspecified, matching Torch defaults).
+
+    Raises
+    ------
+    ValueError
+        If ``device`` is not a CUDA device. RF3's Fabric-backed accelerator
+        only supports GPU execution.
```
k-chrispens: it does appear that the latest foundry does support MPS, but my fork (which this repo depends on) does not and in general sampleworks doesn't support non-CUDA systems right now. I think it'd be worth clarifying that here

k-chrispens (approved):
Looks good, would be worth adding a few extra comments though I think in case these things break longer term

PR #223, opened 2026-04-17
eval: script for classifying conformational changes

k-chrispens:
Evaluation script for classifying conformational changes, which we use for downstream analysis

## Summary by CodeRabbit

* **Chores**
  * Broadened repository ignore patterns to cover additional archive/data file types and expanded dataset directory matching; added logic to better resolve structure file paths when locating CIF files.

* **New Features**
  * Added a structural-variant region classification tool that categorizes alternate-location spans, computes backbone similarity scores between variants, and outputs per-span scores and detailed CSV reports.

### DIFF
--- a/scripts/eval/classify_altloc_regions.py
+++ b/scripts/eval/classify_altloc_regions.py
@@ -1,0 +1,434 @@
+"""Classify altloc regions.
+
+This script consumes the output of ``scripts/eval/find_altloc_selections.py``
+and classifies each contiguous altloc span into one of four bins:
+
+1. ``side_chain_only`` : altloc atoms exist in the residue, but none of its
+   backbone atoms have altlocs.
+2. ``small_loop`` : a contiguous backbone altloc span whose mean per-residue
+   backbone lDDT score (defined below) between altlocs is above
+   ``--loop-lddt-threshold`` (default 0.75).
+3. ``large_loop`` : a contiguous backbone-altloc span whose mean per-residue
+   backbone lDDT score between altlocs is below ``--loop-lddt-threshold``.
+4. ``domain_shift`` : a single contiguous backbone-altloc span longer than
+   ``--domain-shift-min-span`` residues (default 50). Classified before the
+   loop lDDT test.
+
+Score definition (important, slightly different from canonical lDDT):
+
+    For a given pair of altlocs, the score is the **equal-weighted arithmetic
+    mean** of per residue backbone lDDT scores across the span:
+
+        score = (1 / N_span_residues) * sum_k score_k
+
+    Each ``score_k`` is the standard per-residue local lDDT from
+    :class:`sampleworks.metrics.lddt.AllAtomLDDT`, which is the fraction of residue
+    k's neighbor distances (within 15 Å) that are preserved between altlocs across
+    the four lDDT thresholds (0.5, 1, 2, 4 Å).
+
+    The canonical atom pair weighted lDDT would instead aggregate as
+    ``sum_k(score_k * n_pairs_k) / sum_k(n_pairs_k)``. This script's
+    equal residue mean is equivalent with that only when every span residue
+    has the same neighbor count. The 0.75 default is calibrated for this specific
+    calculation.
+
+Altloc pairing: when > 2 altlocs are present, the score above is
+computed for every combination of altloc pairs and the span is
+classified by the *minimum* score over pair combinations.
+
+Use ``find_altloc_selections.py --min-span 1`` to ensure single-residue side only
+selections.
+"""
+
+import argparse
+import json
+import sys
+from pathlib import Path
+
+import numpy as np
+import pandas as pd
+from loguru import logger
+from sampleworks.metrics.lddt import AllAtomLDDT
+from sampleworks.utils.atom_array_utils import (
+    BACKBONE_ATOM_TYPES,
+    BLANK_ALTLOC_IDS,
+    detect_altlocs,
+    filter_to_common_atoms,
+    load_structure_with_altlocs,
+    select_altloc,
+)
+
+
+sys.path.insert(0, str(Path(__file__).resolve().parent))
+from lddt_evaluation_script import translate_selection
+
+
+# np.isin requires a sequence; BLANK_ALTLOC_IDS is a set. Cache the list form.
+_BLANK_ALTLOC_ID_LIST = list(BLANK_ALTLOC_IDS)
+
+
+OUTPUT_COLUMNS = [
+    "protein",
+    "selection",
+    "chain",
+    "start_res",
+    "end_res",
+    "span_length",
+    "classification",
+    "worst_pair_mean_backbone_lddt",
+    "n_backbone_altloc_residues",
+    "n_altlocs",
+    "pair_lddts",
+]
+
+
+def _resolve_cif_path(row: pd.Series, cif_root: Path | None) -> Path:
+    """Resolve a CIF path from a row, preferring ``structure`` then ``structure_pattern``.
+
+    When resolving ``structure_pattern`` against ``cif_root``, this tries both
+    ``{cif_root}/{pattern}`` (flat layout) and ``{cif_root}/{protein}/{pattern}``
+    (per-protein subdirectory layout, as used by the initial_dataset processed dir).
+    """
+    if "structure" in row and isinstance(row["structure"], str) and row["structure"]:
+        p = Path(row["structure"])
+        if p.is_absolute() or p.exists():
+            return p
+        if cif_root is not None:
+            return cif_root / p
+        return p
+
+    if "structure_pattern" not in row or not row["structure_pattern"]:
+        raise ValueError(f"Row has neither 'structure' nor 'structure_pattern': {row.to_dict()}")
+
+    pattern = Path(row["structure_pattern"])
+    if pattern.is_absolute():
+        return pattern
+    if cif_root is None:
+        return pattern
+
+    flat = cif_root / pattern
+    if flat.exists():
+        return flat
+
+    protein = row.get("protein", "")
+    if isinstance(protein, str) and protein:
+        for candidate in (cif_root / protein / pattern, cif_root / protein.upper() / pattern):
+            if candidate.exists():
+                return candidate
+
+    return flat  # fall back to flat so caller's existence check emits the right error
+
+
+def _max_contiguous_run(sorted_res_ids: list[int]) -> int:
+    """Return the length of the longest contiguous run of integers in a sorted list."""
+    if not sorted_res_ids:
+        return 0
+    best = cur = 1
+    for prev, r in zip(sorted_res_ids, sorted_res_ids[1:]):
+        cur = cur + 1 if r == prev + 1 else 1
+        if cur > best:
+            best = cur
+    return best
+
+
+def _build_pairwise_altloc_arrays(
+    atom_array, altloc_ids: list[str]
+) -> dict[tuple[str, str], tuple[object, object]]:
+    """Return ``{(id_i, id_j): (array_i, array_j)}`` pre-filtered to common atoms.
+
+    For each unordered altloc pair we build the two per-altloc AtomArrays
+    (via ``select_altloc(return_full_array=True)``, which includes blank-altloc
+    atoms as shared context) and then run ``filter_to_common_atoms`` so the two
+    inputs have identical atom order and count.
+
+    We build per-pair so residues whose altloc set is a subset of those in the whole structure
+    (e.g. 2YL0 res 60–64 carry only altlocs A and B, not C) still get scored for the pairs where
+    they exist.
+    """
+    pairs: dict[tuple[str, str], tuple[object, object]] = {}
+    for i in range(len(altloc_ids)):
+        for j in range(i + 1, len(altloc_ids)):
+            a_i = select_altloc(atom_array, altloc_ids[i], return_full_array=True)
+            a_j = select_altloc(atom_array, altloc_ids[j], return_full_array=True)
+            try:
+                f_i, f_j = filter_to_common_atoms(a_i, a_j)
+            except RuntimeError as e:
+                logger.warning(
+                    f"could not match atoms between altlocs "
+                    f"{altloc_ids[i]} and {altloc_ids[j]}: {e}"
+                )
+                continue
+            pairs[(altloc_ids[i], altloc_ids[j])] = (f_i, f_j)
+    return pairs
+
+
+def _mean_residue_lddt_for_pair(
+    gt_array,
+    pred_array,
+    chain: str,
+    residues: list[int],
+) -> float:
+    """
+    Equal weighted arithmetic mean of per residue backbone lDDT across the span.
+    """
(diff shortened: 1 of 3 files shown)


scripts/eval/classify_altloc_regions.py
```diff
@@ -0,0 +1,403 @@
...
+    BLANK_ALTLOC_IDS,
+    detect_altlocs,
+    filter_to_common_atoms,
+    load_structure_with_altlocs,
+    select_altloc,
+)
+
+
+# np.isin requires a sequence; BLANK_ALTLOC_IDS is a set. Cache the list form.
+_BLANK_ALTLOC_ID_LIST = list(BLANK_ALTLOC_IDS)
```
marcuscollins: should BLANK_ALTLOC_IDS just be a set? I think we only ever use it with `in`, but I'd have to confirm

scripts/eval/classify_altloc_regions.py
```diff
@@ -0,0 +1,403 @@
...
+    "span_length",
+    "classification",
+    "worst_pair_mean_backbone_lddt",
+    "n_backbone_altloc_residues",
+    "n_altlocs",
+    "pair_lddts",
+]
+
+
+def _max_contiguous_run(sorted_res_ids: list[int]) -> int:
```
marcuscollins: I believe this can be vectorized. Does it get used much?

scripts/eval/classify_altloc_regions.py
```diff
@@ -0,0 +1,403 @@
...
+        return 0
+    best = cur = 1
+    for prev, r in zip(sorted_res_ids, sorted_res_ids[1:]):
+        cur = cur + 1 if r == prev + 1 else 1
+        if cur > best:
+            best = cur
+    return best
+
+
+def _build_pairwise_altloc_arrays(
```
marcuscollins, line 114: You can use grid_search_utils.map_altlocs_to_stack instead, if you apply selections first. Are you trying to avoid using selections? It seems like this is used to filter selections, so you could apply the selection then run map_altlocs_to_stack. It would reduce the number of special functions needed.

scripts/eval/classify_altloc_regions.py
```diff
@@ -0,0 +1,403 @@
...
+    atom_array, altloc_ids: list[str]
```
marcuscollins, line 115: Why not automatically do this over all discovered altlocs?

scripts/eval/classify_altloc_regions.py
```diff
@@ -0,0 +1,403 @@
...
+    except Exception as e:
+        logger.warning(f"lDDT compute failed for chain {chain} residues {residues}: {e}")
+        return float("nan")
+
+    residue_scores = result.get("residue_lddt_scores", {})
+    if not residue_scores:
+        return float("nan")
+
+    flat = [
+        v[0] if isinstance(v, (list, tuple, np.ndarray)) else v for v in residue_scores.values()
```
marcuscollins: This seems weird, for a couple reasons.
1. I don't like the use of .values() here--I guess now it should be ordered, but I prefer to know that order. 
2. Is it really the case that the value is sometimes a plain number, and other times some list-like thing? If so, I don't recall any good reason for that, and we should fix it upstream.

marcuscollins: OTOH, if there is a good reason for 2), you're silently suppressing that behavior here, which also isn't great.

scripts/eval/classify_altloc_regions.py
```diff
@@ -0,0 +1,403 @@
...
+    for prev, r in zip(sorted_res_ids, sorted_res_ids[1:]):
+        cur = cur + 1 if r == prev + 1 else 1
+        if cur > best:
+            best = cur
+    return best
+
+
+def _build_pairwise_altloc_arrays(
+    atom_array, altloc_ids: list[str]
+) -> dict[tuple[str, str], tuple[object, object]]:
```
marcuscollins: the values are tuples of AtomArrays, no?

scripts/eval/classify_altloc_regions.py
```diff
@@ -0,0 +1,403 @@
...
+
+    flat = [
+        v[0] if isinstance(v, (list, tuple, np.ndarray)) else v for v in residue_scores.values()
+    ]
+    return float(np.mean(flat))
+
+
+def _classify_selection(
+    atom_array,
+    pair_arrays: dict[tuple[str, str], tuple[object, object]],
```
marcuscollins: update the type annotation if the values are in fact tuple[AtomArray, AtomArray]

scripts/eval/classify_altloc_regions.py
```diff
@@ -0,0 +1,403 @@
...
+        return float("nan")
+
+    flat = [
+        v[0] if isinstance(v, (list, tuple, np.ndarray)) else v for v in residue_scores.values()
+    ]
+    return float(np.mean(flat))
+
+
+def _classify_selection(
+    atom_array,
```
marcuscollins: missing type annotation

scripts/eval/classify_altloc_regions.py
```diff
@@ -0,0 +1,403 @@
...
+                    f"could not match atoms between altlocs "
+                    f"{altloc_ids[i]} and {altloc_ids[j]}: {e}"
+                )
+                continue
+            pairs[(altloc_ids[i], altloc_ids[j])] = (f_i, f_j)
+    return pairs
+
+
+def _mean_residue_lddt_for_pair(
+    gt_array,
```
marcuscollins: missing type annotations.

scripts/eval/classify_altloc_regions.py
```diff
@@ -0,0 +1,403 @@
...
+    selection_str: str,
+    protein: str,
+    structure_altloc_mask: np.ndarray,
+    structure_backbone_mask: np.ndarray,
+    domain_shift_min_span: int,
+    loop_lddt_threshold: float,
+) -> tuple[dict, set[tuple[str, int]]] | None:
+    """Classify one contiguous altloc selection.
+
+    Returns ``(row_dict, covered_altloc_residues)`` on success or ``None`` if
```
marcuscollins: what is the "row_dict"?

This seems to be doing a fair amount of heavy lifting, the doc string should better describe what it does.

scripts/eval/classify_altloc_regions.py
```diff
@@ -0,0 +1,403 @@
...
+    the selection could not be applied. ``covered_altloc_residues`` is the set
+    of (chain_id, res_id) pairs inside the selection that carry any altloc,
+    used for the caller's residue-coverage invariant check.
+    """
+    try:
+        if not any(op in selection_str for op in ATOMWORKS_COMPARISON_OPS):
+            sel_mask = get_mask_from_old_selection_string(atom_array, selection_str)
+        else:
+            sel_mask = atom_array.mask(selection_str)
+    except Exception as e:
```
marcuscollins: What would the error actually be?

scripts/eval/classify_altloc_regions.py
```diff
@@ -0,0 +1,403 @@
...
+        logger.error(f"[{protein}] failed to apply selection '{selection_str}': {e}")
+        return None
+
+    if not sel_mask.any():
+        logger.warning(f"[{protein}] selection matched no atoms: {selection_str}")
+        return None
+
+    sel_res_ids = np.unique(atom_array.res_id[sel_mask])
+    sel_chain_ids = np.unique(atom_array.chain_id[sel_mask])
+    if len(sel_chain_ids) != 1:
```
marcuscollins: This is a weird error. Our masking _should_ handle this case. find_altloc_selections.py I don't think would ever output such a selection (IIRC it looks at each chain separately). But even so I don't know why the mask couldn't span multiple chains. The real issue is below, where you track backbone residue ids without chain labels. Maybe fix that instead?

scripts/eval/classify_altloc_regions.py
```diff
@@ -0,0 +1,403 @@
...
+    row = {
+        "protein": protein,
+        "selection": selection_str,
+        "chain": chain,
+        "start_res": int(sel_res_ids.min()),
+        "end_res": int(sel_res_ids.max()),
+        "span_length": int(len(sel_res_ids)),
+        "n_backbone_altloc_residues": n_backbone,
+        "n_altlocs": len(altloc_ids),
+        "pair_lddts": json.dumps({}),
```
marcuscollins, line 282: why dump this to a string?

scripts/eval/classify_altloc_regions.py
```diff
@@ -0,0 +1,403 @@
...
+        "worst_pair_mean_backbone_lddt": float("nan"),
+        "classification": "",
+    }
+
+    # Side chain only: no backbone altlocs anywhere in the span.
+    if n_backbone == 0:
+        row["classification"] = "side_chain_only"
+        return row, covered_altloc_residues
+
+    # Domain shift: contiguous backbone-altloc run exceeds threshold (default 50).
```
marcuscollins, line 292: 50 residues seems quite big to me--what about a single helix realignment? These are the kinds of hyperparameters I get worried about.

scripts/eval/classify_altloc_regions.py
```diff
@@ -0,0 +1,403 @@
...
+    if _max_contiguous_run(backbone_altloc_res_ids) > domain_shift_min_span:
+        row["classification"] = "domain_shift"
+        return row, covered_altloc_residues
+
+    # Single residue backbone altlocs cannot yield a meaningful lDDT, since you need at least two
```
marcuscollins: Huh? This comment seems.... wrong. Or I don't understand what it's trying to say.

scripts/eval/classify_altloc_regions.py
```diff
@@ -0,0 +1,403 @@
...
+    # residues for inter-residue distances. A lone backbone-altloc residue is
+    # a minor local perturbation by definition, which we label as small_loop.
+    if n_backbone < 2:
```
marcuscollins: I had the impression from above that "small_loop" referred to a small backbone displacement between altlocs which might include some more substantial side chain altlocs as well.

scripts/eval/classify_altloc_regions.py
```diff
@@ -0,0 +1,403 @@
...
+    # 3. Loop classification via pairwise lDDT across all altloc pairs
+    pair_lddts: dict[str, float] = {}
+    for i in range(len(altloc_ids)):
+        for j in range(i + 1, len(altloc_ids)):
+            pair = pair_arrays.get((altloc_ids[i], altloc_ids[j]))
+            gt, pred = pair if pair is not None else (None, None)
+            pair_lddts[f"{altloc_ids[i]}-{altloc_ids[j]}"] = _mean_residue_lddt_for_pair(
+                gt, pred, chain, backbone_altloc_res_ids
+            )
+    row["pair_lddts"] = json.dumps(pair_lddts)
```
marcuscollins, line 306: again, why dump to a string?

scripts/eval/classify_altloc_regions.py
k-chrispens: I use it with np.isin here, but can switch back

scripts/eval/classify_altloc_regions.py
k-chrispens: Ah, yes, this was the ty VSCode plugin autofill, and it didn't detect AtomArray

scripts/eval/classify_altloc_regions.py
k-chrispens: There is one "domain shift" in the dataset that I care about right now, and it is more like 130 residues, so this successfully detects it. in the future we definitely should make this more robust

scripts/eval/classify_altloc_regions.py
k-chrispens: no but easy to do with claude

scripts/eval/classify_altloc_regions.py
k-chrispens: that will drop selections where the whole structure has 3 altlocs, but some residues only have 2. Added comment on this.

scripts/eval/classify_altloc_regions.py
k-chrispens: I tried to make this clearer? There is no reason the value is a list-like thing, that was overly defensive on my part.

scripts/eval/classify_altloc_regions.py
k-chrispens: added docstring detail about this

scripts/eval/classify_altloc_regions.py
k-chrispens: After looking into this more, this was a misunderstanding on my part of how the LDDT we were calculating worked, since I expected that we were looking within a segment only, not considering the neighborhood. I realize this is not correct and that the regular LDDT calculation is done with the 15A neighborhood around each resiude

scripts/eval/classify_altloc_regions.py
k-chrispens: Yes, it should include this as well. I think this should be what it is now? If the LDDT calculation is what I think?

scripts/eval/classify_altloc_regions.py
k-chrispens: We can then read the json string back in when using the CSV downstream. I suppose we could also store new columns in the CSV?

scripts/eval/classify_altloc_regions.py
k-chrispens: Do you mean by using detect_altlocs in here?

scripts/eval/classify_altloc_regions.py
marcuscollins: Right, so if you do the selection first, then construct the stack on just that selection, it would work, wouldn't it?

scripts/eval/classify_altloc_regions.py
marcuscollins: Yes. I anticipate this is because you are using the entire atom array here, and not a subset. I have always thought about these issues by localizing to the subset first, and only after that worrying about which altlocs are present. I find it easier to think about this way, especially because I don't think there's any clear way to know whether distant altloc stretches are related to each other.

scripts/eval/classify_altloc_regions.py
marcuscollins: If this becomes a column in a data frame that you write out to CSV, it will automatically dump a dictionary to a string: 
```
In [1]: import pandas as pd

In [2]: data = [{"thing": 1, "other": {'blah': 2, 'huh': "yesa"}}]

In [3]: df = pd.DataFrame(data)

In [4]: df
Out[4]: 
   thing                       other
0      1  {'blah': 2, 'huh': 'yesa'}

In [5]: print(df.to_csv())
Out[5]:
,thing,other
0,1,"{'blah': 2, 'huh': 'yesa'}"

```
I'm not sure whether it can be coaxed to read it back in as a dictionary.

marcuscollins (approved):
I'm okay with it as is, but I would favor a selection-first approach that re-uses existing methods. I'm 99% sure it would work and be less code to maintain.

PR #224, opened 2026-04-20
feat: RSCC eval updates

k-chrispens:
Script for finding the max rmsd subsegment in altlocs greater than a certain length, and speedups for the RSCC grid search script.

This depends on #223 and should only be merged afterwards

## Release Notes

* **New Features**
  * Added evaluation scripts to classify alternate-location regions and to find maximum-RMSD residue subsegments.
  * Enhanced RSCC grid search with grouped, parallel processing.
  * Introduced reusable transformer-based density computation helpers for repeated runs.
* **Bug Fixes**
  * RSCC evaluation is more resilient: failures now degrade only the affected selections instead of aborting entire trials.
* **Tests**
  * Added/expanded integration and GPU/slow tests covering success, determinism, and failure paths.
* **Chores**
  * Added `joblib` to support parallel execution.


k-chrispens:
@coderabbitai review

scripts/eval/classify_altloc_regions.py
```diff
@@ -356,6 +317,8 @@ def _process_structure(
         # find_altloc_selections.py appends a combined all altloc selection
         # (atomworks-style with " or " clauses) at the end of each row. That one is
         # a union over every span we already processed individually, so skip it.
+        # NOTE: This will need to be addressed when we
```
marcuscollins, line 323: What will need to be addressed?

scripts/eval/classify_altloc_regions.py
```diff
@@ -356,6 +317,8 @@ def _process_structure(
...
+        # migrate to atomworks-style selections for everything
         if " or " in selection_str:
```
marcuscollins, line 325: We probably want to allow selections that have "or" in them, presumably there will be some that have discontinuous but physically contacting residues that we care about?

scripts/eval/find_max_rmsd_subsegment.py
```diff
@@ -0,0 +1,269 @@
+"""Find the maximum RMSD subsegment within each altloc selection.
+
+This script consumes the output of ``scripts/eval/find_altloc_selections.py``
+and, for each contiguous altloc span longer than ``--window-size`` residues,
+identifies the contiguous subsegment of that size with the highest
+RMSD between any pair of alternate conformations.
+
+Only residues with identical residue names across altlocs are considered.
```
marcuscollins: We should fix this. Maybe make an issue.

scripts/eval/find_max_rmsd_subsegment.py
```diff
@@ -0,0 +1,269 @@
...
+    parse_selection_string,
+)
+from sampleworks.utils.atom_array_utils import (
+    build_pairwise_altloc_arrays,
+    detect_altlocs,
+    load_structure_with_altlocs,
+)
+
+
+def _has_compositional_heterogeneity(arr_i, arr_j, mask: np.ndarray) -> bool:
```
marcuscollins: A couple things:
1. Didn't one of us create a method for this somewhere else? 

2. This will return True if there are simply different sequences that are the same length, which I wouldn't think of as compositional heterogeneity. Do you want to check for modified residues like CYS->CSO?

scripts/eval/find_max_rmsd_subsegment.py
```diff
@@ -0,0 +1,269 @@
...
+
+        for (alt_i, alt_j), (stack_i, stack_j) in pair_arrays.items():
+            arr_i = stack_i[0]
+            arr_j = stack_j[0]
+
+            mask = (arr_i.chain_id == chain) & np.isin(arr_i.res_id, window_res)
+            if mask.sum() == 0:
+                continue
+
+            if _has_compositional_heterogeneity(arr_i, arr_j, mask):
```
marcuscollins: Is this to prevent an error for biotite_rmsd? If so, again I'd think about whether it might make sense to check for modified but otherwise the same residues (e.g. sulfonates). You could take the rmsd on only the common atoms.

scripts/eval/find_max_rmsd_subsegment.py
```diff
@@ -0,0 +1,269 @@
...
+    _, names_i = get_residues(arr_i[mask])
+    _, names_j = get_residues(arr_j[mask])
+    return not np.array_equal(names_i, names_j)
+
+
+def _find_max_rmsd_window(
+    pair_arrays: dict[tuple[str, str], tuple[AtomArrayStack, AtomArrayStack]],
+    chain: str,
+    residues: list[int],
+    window_size: int = 3,
```
marcuscollins, line 41: Is this the value you usually use? 3 seems very short to me. Can we go for 5 or more?

scripts/eval/find_max_rmsd_subsegment.py
```diff
@@ -0,0 +1,269 @@
...
+        return []
+
+    pair_arrays = build_pairwise_altloc_arrays(atom_array, altloc_info.altloc_ids)
+    if not pair_arrays:
+        logger.warning(f"[{protein}] no valid altloc pairs could be built; skipping")
+        return []
+
+    output_rows: list[dict] = []
+    for sel_str in [s.strip() for s in selection_field.split(";") if s.strip()]:
+        # Skip atomworks-style catchall selections # NOTE: This will need to be addressed when we
```
marcuscollins: Let's just fix this now and make it work for atomworks style selections.

scripts/eval/find_max_rmsd_subsegment.py
```diff
@@ -0,0 +1,269 @@
...
+
+def main(args: argparse.Namespace) -> None:
+    input_df = pd.read_csv(args.input_csv)
+    required = {"protein", "selection"}
+    missing = required - set(input_df.columns)
+    if missing:
+        raise ValueError(f"Input CSV missing required columns: {missing}")
+
+    all_rows: list[dict] = []
+    for _, row in input_df.iterrows():
```
marcuscollins: There's a separate module for parallelizing things over rows of a dataframe, I forget what it is called, but it can speed things up a lot if _process_structure is a slow step.

scripts/eval/find_max_rmsd_subsegment.py
```diff
@@ -0,0 +1,269 @@
...
+                    "structure_pattern",
+                    "map_pattern",
+                    "base_map_dir",
+                    "resolution",
+                ]
+            )
+        )
+    else:
+        final_rows = []
+        for protein, group in detail_df.groupby("protein", sort=False):
```
marcuscollins: You should use a DataFrame.groupby.agg pattern here. No reason to use a for loop.

scripts/eval/rscc_grid_search_script.py
```diff
@@ -9,25 +9,30 @@
...
 ## Workflow:
 1. Scan the `grid_search_results` directory for completed trials
-2. For each trial with a `refined.cif`, compute the electron density map
+2. For each trial with a `refined.cif`, compute the electron density map (trials are grouped
+   by ``(protein, occupancy_key)`` and processed in parallel, configure with ``--n-jobs``)
 3. Compare against the corresponding base map and calculate RSCC
 4. Aggregate and visualize results by ensemble size, guidance weight, and scaler type
+
+Depending on the GPU, --n-jobs=8-16 work well. A CUDA RuntimeError in a worker is caught per-trial
+(the row gets ``rscc=nan``) but may affect other trials in the same worker.
```
marcuscollins: This is cryptic. Why is there always a row with RSCC=nan? It seems like we should just fix that rather than tolerate it.

Or do you mean that if there is a CUDA error, the row gets RSCC=nan?

scripts/eval/rscc_grid_search_script.py
```diff
@@ -38,20 +43,221 @@
...
+
+    Raises
+    ------
+    ValueError
+        If the base map cannot be loaded.
+    """
+    valid_selections = [s for s in protein_config.selection if s in group_ref_coords]
+    rows: list[dict] = []
+
+    device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
```
marcuscollins: We should make some effort to spread these out over the available GPUs.

scripts/eval/rscc_grid_search_script.py
```diff
@@ -38,20 +43,221 @@
...
+    # Load base map + transformer + reference once for the whole group.
+    # TODO: this needs to be better unified with what's in generate_synthetic_density
+    try:
+        # Load base map for canonical unit cell,
+        # don't overwrite the base map with selection map--we'll use the full map later too.
+        base_xmap = protein_config.load_map(base_map_path)
+        if base_xmap is None:
+            raise ValueError(f"Failed to load base map from {base_map_path}")
+
+        transformer, _ = build_density_transformer(base_xmap, em_mode=False, device=device)
```
marcuscollins, line 122: Probably this needs a comment to explain what the method does and what the transformer is used for later. This is the object that actually builds the map from coordinates, right?

tests/resources/1vme/1VME_0.25occA_0.75occB_1.00A.ccp4
marcuscollins, line 1: If there's a way to avoid a binary file format, I would do it. If this map changes for any reason, the entire file will be replaced, and you end up with a big file forever in your history either way.

tests/resources/1vme/1VME_0.5occA_0.5occB_1.00A.ccp4
marcuscollins, line 1: Ditto this binary file

tests/eval/test_rscc_grid_search_script.py
```diff
@@ -0,0 +1,258 @@
...
+    installed on ``sys.path``"""
+    spec = importlib.util.spec_from_file_location("rscc_grid_search_script", _SCRIPT_PATH)
+    assert spec is not None and spec.loader is not None
+    mod = importlib.util.module_from_spec(spec)
+    spec.loader.exec_module(mod)
+    return mod
+
+
+@pytest.fixture
+def rscc_script():
```
marcuscollins: why have this wrapper function? Do you use _load_script somewhere else where it needs to not be a fixture?

tests/eval/test_rscc_grid_search_script.py
```diff
@@ -0,0 +1,258 @@
...
+
+    def flaky_rscc(a, b, _sel=bad_selection):
+        if flaky_rscc.target_is_next:  # type: ignore[attr-defined]
+            flaky_rscc.target_is_next = False  # type: ignore[attr-defined]
+            raise RuntimeError("simulated rscc failure")
+        return real_rscc(a, b)
+
+    # The selection loop calls rscc once per selection, in the order configured
+    # in the CSV. Fail on the first call (i.e. the first selection).
+    flaky_rscc.target_is_next = True  # type: ignore[attr-defined]
```
marcuscollins: This is a little bit abusive of Python's dynamic nature. It's clever, but you could also create `flaky_rscc` as a class with the desired attribute `target_is_next` and a `__call__` method.

tests/eval/test_rscc_grid_search_script.py
```diff
@@ -0,0 +1,258 @@
...
+    assert np.isnan(df.loc[bad_selection, "rscc"])
+    assert pd.notna(df.loc[bad_selection, "error"])
+
+    good_selection = fx.selections[1]
+    assert not np.isnan(df.loc[good_selection, "rscc"])
+    assert pd.isna(df.loc[good_selection, "error"])
+
+
+@pytest.mark.slow
+def test_selections_missing_from_ref_coords_produce_no_row(rscc_fixture_factory, rscc_script):
```
marcuscollins: Do we actually do this? I think we shouldn't silently skip any selection, we should somehow warn the user.

tests/eval/test_rscc_grid_search_script.py
```diff
@@ -0,0 +1,258 @@
...
+
+    monkeypatch.setattr(rscc_script, "build_density_transformer", counting_build)
+
+    rscc_script.main(_make_args(fx))
+
+    assert len(calls) == 2, f"expected 2 builds (one per group), got {len(calls)}"
+
+
+@pytest.mark.slow
+def test_trials_grouped_by_sort_order(rscc_fixture_factory, rscc_script):
```
marcuscollins: Is this important? Some of these tests seem to be testing things that aren't really important. As the package grows and more people contribute this can lead to fixing behaviors we don't really care about. I suspect these are AI-generated? One thing I've noticed is that AI systems seem to test whatever they can, not what we actually care about. This can lead to extra tests than make our development rigid, and testing runs slower. I would remove anything that isn't functionally important.

marcuscollins (changes requested):
Some new comments, plus old ones... please address and I'll take another look through.

tests/eval/test_rscc_grid_search_script.py
k-chrispens: I also did this in the other eval script test, it was suggested by Claude and I thought it made things a bit clearer. Can remove from both if you would rather it be removed

scripts/eval/find_max_rmsd_subsegment.py
k-chrispens: 3 is what we use in the paper. aside from coming suggested by Jaime, when I test 3 vs 5, I got much more spread RSCC with 3

marcuscollins (approved):
Thanks for the changes.

PR #227, opened 2026-04-24
feat: add RMSD metric

k-chrispens:
Adds RMSD metric, which we can use to hook up to downstream clustering analysis if we want.

## Summary by CodeRabbit

* **New Features**
  * Added an All-atom RMSD metric: computes global RMSD with optional superposition, per-residue RMSD breakdown, and optional per-batch RMSD logging. Selections can limit reported residues without changing the global RMSD.

* **Tests**
  * Added tests and shared fixtures validating identity cases, selection behavior, and end-to-end RMSD values.

[Review Change Stack](https://app.coderabbit.ai/change-stack/diff-use/sampleworks/pull/227?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/sampleworks/metrics/rmsd.py
+++ b/src/sampleworks/metrics/rmsd.py
@@ -1,0 +1,172 @@
+"""RMSD metric, developed starting from :class:`sampleworks.metrics.lddt.AllAtomLDDT`.
+
+The metric returns a global per model RMSD and a per residue RMSD dictionary,
+so it can be plugged into the same clustering code as the LDDT
+metric.
+"""
+
+from typing import Any, cast
+
+import numpy as np
+from atomworks.io.transforms.atom_array import ensure_atom_array_stack
+from atomworks.ml.transforms.atom_array import add_global_token_id_annotation
+from biotite.structure import AtomArray, AtomArrayStack, rmsd, superimpose
+
+from sampleworks.metrics.metric import Metric
+from sampleworks.utils.atom_array_utils import filter_to_common_atoms
+
+
+class AllAtomRMSD(Metric):
+    """Computes all atom RMSD from AtomArrays.
+
+    Parameters
+    ----------
+    superimpose
+        If True, superimpose predicted models onto the reference via Kabsch
+        before computing RMSD. Default False is correct for comparisons in a shared crystallographic
+        frame, but set True when the predicted and reference structures live in different frames.
+    log_rmsd_for_every_batch
+        If True, include per model RMSDs as ``all_atom_rmsd_<i>`` in the
+        output dict.
+    """
+
+    def __init__(
+        self,
+        superimpose: bool = False,
+        log_rmsd_for_every_batch: bool = False,
+        **kwargs: Any,
+    ):
+        super().__init__(**kwargs)
+        self.superimpose = superimpose
+        self.log_rmsd_for_every_batch = log_rmsd_for_every_batch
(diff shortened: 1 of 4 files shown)


src/sampleworks/metrics/rmsd.py
```diff
@@ -0,0 +1,156 @@
...
+        gt_ref = gt_aa_stack[0]
+        if self.superimpose:
+            pred_aa_stack, _ = superimpose(gt_ref, pred_aa_stack)
+
+        tok_idx = cast(np.ndarray, gt_ref.token_id).astype(np.int64)
+
+        # Resolve the subset of tokens to report, if a residue selection was given.
+        selected_token_ids: set[int] | None = None
+        if selection is not None:
+            mask_fn = pred_aa_stack.mask
```
marcuscollins: is it necessary to create this variable `mask_fn`? Why not just use `pred_aa_stack.mask`?

marcuscollins: Also, won't this raise an attribute error if the mask function isn't there?

src/sampleworks/metrics/rmsd.py
```diff
@@ -0,0 +1,156 @@
...
+            if mask_fn is None:
+                raise RuntimeError(
+                    "pred_aa_stack does not support mask(). Load atom arrays with "
+                    "`atomworks.io.utils.io_utils.load_any()` to access this method."
+                )
+            mask = mask_fn(selection)
+            selected_arr = cast(AtomArray, pred_aa_stack[0, mask])
+            if selected_arr.token_id is not None:
+                selected_token_ids = {int(t) for t in np.unique(selected_arr.token_id)}
```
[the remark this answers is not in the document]
marcuscollins: @k-chrispens definitely look into this and harden it.

src/sampleworks/metrics/rmsd.py
```diff
@@ -0,0 +1,156 @@
...
+                continue
+            atom_mask = tok_idx == tk
+            residue_rmsd = np.asarray(rmsd(gt_ref[atom_mask], pred_aa_stack[:, atom_mask]))
+            residue_rmsd_scores[token_to_residue_id_map[tk_int]] = residue_rmsd.tolist()
+
+        result: dict[str, Any] = {
+            "best_of_1_rmsd": float(global_rmsd[0]),
+            f"best_of_{len(global_rmsd)}_rmsd": float(global_rmsd.min()),
+            "residue_rmsd_scores": residue_rmsd_scores,
+        }
```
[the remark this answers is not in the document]
marcuscollins: I'm not worried by this. Ignore.

tests/metrics/conftest.py
```diff
@@ -0,0 +1,23 @@
...
+import pytest
+from atomworks.io.transforms.atom_array import ensure_atom_array_stack
+from biotite.structure import AtomArrayStack
+from sampleworks.utils.atom_array_utils import (
+    select_altloc,
+    select_backbone,
+)
+
+
+@pytest.fixture(scope="module")
```
marcuscollins: Should these have something else for their scope now? I.e., is it a problem for them to be kept around for multiple test modules?

k-chrispens: I don't think so, so will change to session scope

marcuscollins (approved):

PR #230, opened 2026-05-05
feat: append JobResult timing and status to job_metadata.json

### LINKED ISSUE #212: Output metadata should be updated to include start and end time from JobResult
Right now, we write out metadata to a file `job_metadata.json` in the output directory, e.g., 
`/data/sampleworks-exp/occ_sweep/grid_search_results_rf3_chiral_off/1VME_0.75occA_0.25occB/rf3_X-RAY_DIFFRACTION/pure_guidance/ens8_gw0.1/job_metadata.json`. The data in this file is derived from the corresponding GuidanceConfig object, via its `as_dict` method. We should append this with other information in the `JobResult` object that is returned by the `run_guidance` method in guidance_script_utils.py, most importantly the start and end times of the trial.

Abdelsalam-Abbas:
## Summary
- Closes #212. `job_metadata.json` now includes `started_at`, `finished_at`, `runtime_seconds`, `status`, and `exit_code` from `JobResult` in addition to the existing `GuidanceConfig` snapshot.
- Adds `JobResult.as_dict()` mirroring `GuidanceConfig.as_dict()` so `output_dir` and `log_path` get container-to-host remapping. Without this, the merged file would regress to container paths in Docker runs.
- New helper `_write_job_metadata` is called both inside `save_everything` (config-only snapshot, in case the run crashes later) and at the end of `run_guidance` (enriched with `JobResult`, on both success and failure paths).

## Test plan
- [x] `pixi run -e boltz-dev cpu-tests tests/utils/test_guidance_script_utils.py` — 5 tests pass, including the new `test_write_job_metadata_remaps_job_result_paths_to_host` regression test that sets `SAMPLEWORKS_HOST_RESULTS_DIR` and asserts JobResult paths come out host-remapped.
- [ ] Manual: run a guidance job and confirm `job_metadata.json` now contains both the GuidanceConfig fields and the JobResult timing/status fields.

## Summary by CodeRabbit

* **New Features**
  * Job results now serialize to a dictionary and remap container paths to host paths for saved outputs.

* **Refactor**
  * Job metadata writing is centralized; runs always emit a merged metadata file including runtime, status, and exit information.

* **Tests**
  * Added tests and a fixture covering metadata writing, output-directory creation, and path remapping behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/diff-use/sampleworks/pull/230?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


src/sampleworks/utils/guidance_script_utils.py
```diff
@@ -333,12 +333,11 @@ def save_everything(
...
-    # Even though this is technically duplicated, keep it around as a backup in case metadata
-    # is lost in some CIF transform.
-    with open(Path(output_dir) / "job_metadata.json", "w") as fp:
-        # use the GuidanceConfig's as_dict() method to avoid serializing PosixPath objects
-        json.dump(metadata, fp)
+    # Write out the job parameters to a JSON file alongside refined.cif. Even though this is
+    # technically duplicated with the CIF category, keep it around as a backup in case metadata
+    # is lost in some CIF transform. JobResult fields (timing, status) are appended later in
+    # run_guidance once the run finishes.
+    _write_job_metadata(output_dir, args)
```
marcuscollins: `save_everything` is called only at the end of `_run_guidance`. Then in `run_guidance`, you call `_write_job_metadata` again, that time with the `job_result` argument. I think you probably don't need to write out the metadata here, since it will be immediately overwritten with the more complete version.

tests/utils/test_guidance_script_utils.py
```diff
@@ -38,3 +40,143 @@ def test_save_everything_uses_model_atom_array_for_mismatch(tmp_path: Path):
     )
 
     assert (tmp_path / "refined.cif").exists()
+
+
+def _make_job_result(output_dir: str) -> JobResult:
```
marcuscollins: This should probably be a test fixture in conftest.py

tests/utils/test_guidance_script_utils.py
```diff
@@ -38,3 +40,143 @@ def test_save_everything_uses_model_atom_array_for_mismatch(tmp_path: Path):
...
+        exit_code=0,
+        runtime_seconds=12.34,
+        started_at="2026-05-05T10:00:00",
+        finished_at="2026-05-05T10:00:12.340000",
+        log_path=str(Path(output_dir) / "run.log"),
+        output_dir=output_dir,
+    )
+
+
+def test_write_job_metadata_without_job_result_writes_guidance_config(tmp_path: Path):
```
marcuscollins: If you only call the method after you have the job result object, then maybe this test is unnecessary?

marcuscollins (commented):
The main question is whether we need to call the `_write_job_metadata` method twice. The only thing that happens in between the two calls, AFAICT, is fetching the JobResult object. Is it worth writing the file, and then immediately writing it again? If we can make a good case for that (fetching the JobResult is error prone, maybe?) then we can keep as is, and I'll approve. Otherwise I think we should simplify this, and adjust unit tests accordingly.

marcuscollins (approved):
LGTM, thanks for the changes.

PR #232, opened 2026-05-12
feat(diffuse): add ACTL-native Sampleworks wrappers

xraymemory:
## Summary
- Add ACTL/Kubernetes-native Sampleworks shell wrappers that run inside the already-started container instead of launching nested Docker.
- Add a generic `run_grid_search_actl.sh` wrapper plus ACTL copies for the all-model, RF3, Protenix, and MDC wrapper flows.
- Update the image entrypoint/Dockerfile to bundle and directly execute `*_actl.sh` wrappers.

## Validation
- `bash -n docker-entrypoint.sh`
- `for script in *_actl.sh; do bash -n "$script"; done`
- Searched ACTL wrappers for nested Docker launch patterns; none found.

## Notes
- This PR targets `add-diffuse-yaml` so the diff stays focused on ACTL wrapper changes and can stack on top of #225.
- Original Docker-launching wrappers are preserved.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -106,8 +106,9 @@ COPY pyproject.toml pixi.lock ./
 COPY src/ ./src/
 COPY scripts/ ./scripts/
 COPY run_grid_search.py ./
+COPY *_actl.sh /usr/local/bin/
 COPY docker-entrypoint.sh /usr/local/bin/entrypoint.sh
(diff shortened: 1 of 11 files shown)


xraymemory:
Superseded by #233, which targets main directly.

PR #233, opened 2026-05-12
feat(diffuse): add ACTL-native Sampleworks wrappers

xraymemory:
## Summary
- Add ACTL/Kubernetes-native Sampleworks shell wrappers that run inside the already-started container instead of launching nested Docker.
- Add a generic `run_grid_search_actl.sh` wrapper plus ACTL copies for all-model, RF3, Protenix, and MDC wrapper flows.
- Update the image entrypoint/Dockerfile to bundle and directly execute `*_actl.sh` wrappers.

## Validation
- `bash -n docker-entrypoint.sh`
- `for script in *_actl.sh; do bash -n "$script"; done`
- Searched ACTL wrappers for nested Docker launch patterns; none found.

## Notes
- Targets `main` directly.
- Original Docker-launching wrappers are preserved.

## Summary by CodeRabbit

* **New Features**
  * Added container-native wrapper scripts to run model grid searches inside Kubernetes/ACTL containers (no Docker-in-Docker).
  * Orchestration for running multiple model grid searches in parallel with configurable GPU pinning and per-job logs.
  * New specialized runners for RF3 partial-diffusion sweeps and Protenix variant runs, with env-var configuration and checkpoint support.
  * Improved entrypoint help and handling to recognize and directly execute bundled ACTL wrapper commands.

[Review Change Stack](https://app.coderabbit.ai/change-stack/diff-use/sampleworks/pull/233)

### DIFF
--- a/rf3_partial_occ_sweep_chiral_off_actl.sh
+++ b/rf3_partial_occ_sweep_chiral_off_actl.sh
@@ -1,0 +1,46 @@
+#!/bin/bash
+# Run RF3 occ-sweep partial grid search with chiral features disabled.
+
+set -euo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+RUN_GRID_SEARCH_ACTL="${RUN_GRID_SEARCH_ACTL:-$SCRIPT_DIR/run_grid_search_actl.sh}"
+
+if [[ "${1:-}" == "-h" || "${1:-}" == "--help" ]]; then
+    cat << 'EOF'
+Usage: rf3_partial_occ_sweep_chiral_off_actl.sh
+
+Environment overrides: DATA_DIR, RESULTS_DIR, MSA_CACHE_DIR, PROTEINS_CSV, GPU_DEVICES, RF3_CHECKPOINT
+EOF
+    exit 0
+fi
+
+DATA_DIR="${DATA_DIR:-/mnt/diffuse-private/raw/sampleworks/initial_dataset_40_occ_sweeps}"
+RESULTS_DIR="${RESULTS_DIR:-/data/sampleworks-exp/occ_sweep/grid_search_results_rf3_chiral_off}"
+MSA_CACHE_DIR="${MSA_CACHE_DIR:-${HOME}/sampleworks-exp/msa_cache}"
+PROTEINS_CSV="${PROTEINS_CSV:-$DATA_DIR/proteins.csv}"
+GPU_DEVICES="${GPU_DEVICES:-5}"
+
+if [[ ! -f "$PROTEINS_CSV" ]]; then
+    echo "Error: proteins CSV not found: $PROTEINS_CSV" >&2
+    exit 1
+fi
+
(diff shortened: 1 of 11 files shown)


xraymemory:
Addressed review feedback in bbc296d.

- Wrapped background `cmd | tee` launches in subshells with `set -o pipefail` so failures propagate through PID waits.
- Split concurrent grid-search output roots into per-model/per-run subdirectories to avoid `results.json`/worker artifact races while keeping top-level logs readable.
- Updated `run_grid_search_actl.sh` help to make `--proteins` required.
- Documented `RF3_CHECKPOINT` where used.
- Aligned entrypoint examples with main's current `run_grid_search.py` flags (`--model`, `--method`, `--model-checkpoint`).
- Intentionally did not change `$HOME` defaults solely for persistence/mount concerns because the container is treated as the runtime universe here.

Validation:
- `bash -n docker-entrypoint.sh`
- `for script in *_actl.sh; do bash -n "$script"; done`
- `for script in *_actl.sh; do "./$script" --help >/dev/null; done && ./docker-entrypoint.sh --help >/dev/null`
- Checked no nested Docker launch patterns remain in ACTL wrappers.

rf3_partial_occ_sweep_chiral_off_actl.sh
```diff
@@ -0,0 +1,46 @@
...
+Usage: rf3_partial_occ_sweep_chiral_off_actl.sh
+
+Environment overrides: DATA_DIR, RESULTS_DIR, MSA_CACHE_DIR, PROTEINS_CSV, GPU_DEVICES, RF3_CHECKPOINT
+EOF
+    exit 0
+fi
+
+DATA_DIR="${DATA_DIR:-/mnt/diffuse-private/raw/sampleworks/initial_dataset_40_occ_sweeps}"
+RESULTS_DIR="${RESULTS_DIR:-/data/sampleworks-exp/occ_sweep/grid_search_results_rf3_chiral_off}"
+MSA_CACHE_DIR="${MSA_CACHE_DIR:-${HOME}/sampleworks-exp/msa_cache}"
```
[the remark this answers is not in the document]
marcuscollins: Probably worth changing. `$HOME/sampleworks-exp/` is sort of specific to the particular bare-metal node we were using.

run_all_models_mdc_actl.sh
```diff
@@ -0,0 +1,145 @@
+#!/bin/bash
```
marcuscollins, line 1: Probably don't need this file.

run_grid_search_actl.sh
```diff
@@ -0,0 +1,114 @@
+#!/bin/bash
```
marcuscollins, line 1: I don't think we need this one anymore, do we? I think we should pare down these files to get to a minimal set.

marcuscollins (changes requested):
My main comment is to reduce the number of scripts we have. I think we can keep just:

run_all_models_actl.sh
run_protenix_mini_tiny_actl.sh

the others are all old derivatives of those, I'm pretty sure, and were created for some ad hoc purpose. We should purge them.

PR #234, opened 2026-05-14
feat(synthetic-sf): add script to generate synthetic structure factors

DorisMai:
## Summary

  - Add `generate_synthetic_sf.py` to produce MTZ files of structure factors using SFcalculator (SFC), with optional bulk solvent + default scaling allowed as well as R-free flag generation. Assumes no anomalous scattering for now.
  - Mirror `generate_synthetic_density.py` in terms of supporting single_structure vs batch_csv generation, altloc occupancy overwrite, and selection and stripping of hydrogens/waters/ligands.
  - Refactor out `assign_occupancies` and selection/stripping from `generate_density_sf.py` into `eval/synthetic_utils` for reuse in the SF scripts.
  - SFC is instantiated via AtomArray though `atomarray_to_gemmi`. As this function is likely reused in prototyping the new reward function, test is added. Test is specific to SFC environment.
  
  ## Test plan
  - SFC is kept as a separate dependency group, which results in a few new environments -- `analysis-dev-sfc`, `boltz-dev-sfc`, and `rf3-dev-sfc`, that eventually will be merged (or added to workflows).
:exclamation: There is dependency conflict with Protenix (Issue #235 ), hence no corresponding env.
  - [x] All 3 new environments passed `tests/eval/test_generate_synthetic_sf.py`
  - [x] Validate occupancy change effects visually in coot for 6B8X and 2YL0
  - [x] Validate batch_csv mode works
  - [x] Validate Ftotal and Fprotein values on 5SOP_chainA_altlocB with manual selection/stripping/overwrite in pymol  --> pymol saved to cif --> SFC from cif

## Next steps
- Prototype new reward function, focus on single chain for now.
- Test multi-chain synthetic generation, otherwise Ftotal does not make sense.
- Estimate memory usage/constraints in SFC. 

## Summary by CodeRabbit

* **New Features**
  * CLI to generate synthetic structure-factor amplitudes with single-file or CSV batch modes, MTZ output, per-row overrides, and optional saving of processed structures.

* **Refactor**
  * Centralized structure loading, selection/stripping, altloc detection, and occupancy assignment into shared utilities.

* **Tests**
  * Added tests for conversion, occupancy validation/behavior, and scattering-factor computations.

* **Bug Fixes / Robustness**
  * Improved batch parsing and A3M handling; safer CSV column handling; clearer validation and error reporting for metrics and model checkpoint inputs.

* **Chores**
  * Project config updated with SFC-related dependency and a pandas version override.

[Review Change Stack](https://app.coderabbit.ai/change-stack/diff-use/sampleworks/pull/234?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


tests/eval/test_generate_synthetic_sf.py
```diff
@@ -0,0 +1,122 @@
...
+            device,
+        )
+
+        arr_custom = assign_occupancies(stripped_atom_array, altloc_info, "custom", [0.2, 0.8, 0.0])
+        f_custom = _compute_fprotein(
+            atomarray_to_gemmi(arr_custom, stripped_gemmi.cell, stripped_gemmi.spacegroup_hm),
+            device,
+        )
+
+        assert not np.allclose(np.abs(f_default), np.abs(f_custom), atol=1e-3)
```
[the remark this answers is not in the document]
DorisMai: this test is not slow and should be able to run with cpu too

src/sampleworks/eval/synthetic_utils.py
```diff
@@ -0,0 +1,181 @@
...
+        atom_array = apply_selection(atom_array, selection)
+    except ValueError as e:
+        logger.error(f"Selection error for {structure_path}: {e}")
+        return None
+
+    atom_array = remove_hydrogens(atom_array) if strip_hydrogens else atom_array
+    atom_array = remove_waters(atom_array) if strip_waters else atom_array
+    atom_array = keep_polymer(keep_amino_acids(atom_array)) if strip_ligands else atom_array
+    # This is currently a sort of hacky way to remove ligands by keeping only polymer atoms
+    # TODO: there's probably a more robust way to do this
```
marcuscollins: This comment should be moved up, I think it applies to line 155.

src/sampleworks/eval/synthetic_utils.py
```diff
@@ -0,0 +1,181 @@
...
+    altloc_info = detect_altlocs(atom_array)  # ty: ignore[invalid-argument-type]
+    if occ_values:
+        if occ_mode != "custom":
+            logger.warning(
+                f"Custom occupancy values provided for {structure_path}, "
+                f"but occ_mode is '{occ_mode}'. Using 'custom' mode."
+            )
+            occ_mode = "custom"
+        try:
+            atom_array = assign_occupancies(atom_array, altloc_info, "custom", occ_values)
```
marcuscollins: Having changed `occ_mode` to "custom", you may as well use that here.

src/sampleworks/eval/generate_synthetic_sf.py
```diff
@@ -0,0 +1,634 @@
+"""Generate synthetic structure factor amplitudes via SFcalculator-torch.
+
+Produces an MTZ file of |F_protein| (no bulk solvent, no scaling) for each input
+PDB/mmCIF structure. v1 scope: protein-only structure factors, with optional
+per-row override of unit cell + space group and optional R-free flag column.
```
marcuscollins: Just for clarity, you should specify what each row is (a structure, I guess?)

src/sampleworks/eval/generate_synthetic_sf.py
```diff
@@ -0,0 +1,634 @@
...
+from loguru import logger
+from sampleworks.eval.synthetic_utils import load_structure_for_synthetic_reward
+from sampleworks.utils.atom_array_utils import BLANK_ALTLOC_IDS
+from sampleworks.utils.torch_utils import try_gpu
+from SFC_Torch import SFcalculator
+from SFC_Torch.io import array2hier, PDBParser
+
+
+@dataclass
+class BatchRow:
```
marcuscollins: I think we have another BatchRow somewhere, don't we? Should we give them different, more distinguishing names (e.g., here, BatchRowForMTZ or somesuch.)?

src/sampleworks/eval/generate_synthetic_sf.py
```diff
@@ -0,0 +1,634 @@
...
+    """A row from the batch processing CSV file.
+
+    Attributes
+    ----------
+    filename
+        Path to the structure file (relative to base_dir)
```
marcuscollins: where does base_dir come from?

src/sampleworks/eval/generate_synthetic_sf.py
```diff
@@ -0,0 +1,634 @@
...
+    mtzfile
+        Optional custom output filename for the MTZ
+    unit_cell
+        Optional unit cell to override the one in the structure file
+    space_group
+        Optional space group (in Hermann-Mauguin string format) to override the
```
marcuscollins: maybe add a link that defines Hermann-Mauguin format?

src/sampleworks/eval/generate_synthetic_sf.py
```diff
@@ -0,0 +1,634 @@
...
+        one in the structure file.
+    selection
+        Optional atom selection string in pyMOL-like syntax (e.g., 'chain A and resi 10-50')
+    occ_values
+        Custom list of occupancy values for altlocs, must be in range [0.0, 1.0]
+    """
+
+    VALID_EXTENSIONS: ClassVar[frozenset[str]] = frozenset({".pdb", ".cif", ".mmcif", ".ent"})
```
marcuscollins: What is .ent?

marcuscollins: FWIW I would prefer to push everyone to use mmCIF only.

src/sampleworks/eval/generate_synthetic_sf.py
```diff
@@ -0,0 +1,634 @@
...
+
+    def __post_init__(self) -> None:
+        ext = Path(self.filename).suffix.lower()
+        if ext not in self.VALID_EXTENSIONS:
+            raise ValueError(
+                f"Invalid file extension '{ext}' for '{self.filename}'. "
+                f"Expected one of: {', '.join(sorted(self.VALID_EXTENSIONS))}"
+            )
+        for i, v in enumerate(self.occ_values):
+            if not 0.0 <= v <= 1.0:
```
marcuscollins: You should probably check that they sum to 1.0?

src/sampleworks/eval/generate_synthetic_sf.py
```diff
@@ -0,0 +1,634 @@
...
+                raise ValueError(f"Occupancy value {v} at index {i} is out of range [0.0, 1.0]")
+
+    @classmethod
+    def from_dict(cls, row: dict[str, Any]) -> "BatchRow":
+        """Create a BatchRow from a CSV row dictionary.
+
+        CSV columns: filename (required), mtzfile, unit_cell (six floats
```
marcuscollins: Please put one column per line, it is easier to read that way.

src/sampleworks/eval/generate_synthetic_sf.py
```diff
@@ -0,0 +1,634 @@
...
+                    f"unit_cell must be 6 colon-separated values (a:b:c:alpha:beta:gamma), "
+                    f"got {len(parts)}: {row['unit_cell']!r}"
+                )
+            unit_cell = gemmi.UnitCell(*parts)
+
+        space_group: str | None = None
+        if row.get("space_group"):
+            space_group = row["space_group"]
+            if space_group.isdigit():
+                space_group = gemmi.SpaceGroup(int(space_group)).hm
```
marcuscollins, line 114: does `gemmi.SpaceGroup().hm` always return a string?

src/sampleworks/eval/generate_synthetic_sf.py
```diff
@@ -0,0 +1,634 @@
...
+    Anisotropic B-factors are set to zero since biotite does not store them.
+    Blank altloc labels are converted from biotite's '' to gemmi's '\\x00'.
+
+    Parameters
+    ----------
+    atom_array
+        Input structure with occupancy and b_factor annotations
+    unit_cell
+        Crystallographic unit cell for the structure. If None, gemmi defaults
+        to (1.0, 1.0, 1.0, 90.0, 90.0, 90.0).
```
marcuscollins: 1 Angstrom? Or does it do everything in normalized unit cell lengths (so, dimensionless?) either way please make a note of how gemmi interprets these numbers.

src/sampleworks/eval/generate_synthetic_sf.py
```diff
@@ -0,0 +1,634 @@
...
+    -------
+    gemmi.Structure
+        Structure ready to be wrapped by SFC_Torch.io.PDBParser
+    """
+    n = len(atom_array)
+    cra_names = [
+        f"{atom_array.chain_id[i]}-0-{atom_array.res_name[i]}-{atom_array.atom_name[i]}"
+        for i in range(n)
+    ]
+    # gemmi uses '\x00' for blank altloc; biotite uses ''
```
marcuscollins: the code is correct, but the comment indicates biotite uses '', IIRC it sometimes also uses ' ' (a space)

src/sampleworks/eval/generate_synthetic_sf.py
```diff
@@ -0,0 +1,634 @@
...
+    atom_altloc = ["\x00" if a in BLANK_ALTLOC_IDS else a for a in atom_array.altloc_id]
+    structure = array2hier(
```
marcuscollins: please annotate the type of `structure`

src/sampleworks/eval/generate_synthetic_sf.py
```diff
@@ -0,0 +1,634 @@
...
+        res_id=atom_array.res_id,
+    )
+    if unit_cell is not None:
+        structure.cell = unit_cell
+    if space_group is not None:
+        structure.spacegroup_hm = space_group
+    return structure
+
+
+def write_amplitudes_to_mtz(
```
marcuscollins: SFC doesn't have its own method for this?

src/sampleworks/eval/generate_synthetic_sf.py
```diff
@@ -0,0 +1,634 @@
...
+    sfc: SFcalculator,
+    output_path: Path,
+    test_fraction: float = 0.05,
+    seed: int | None = None,
+    hkl_attr: str = "Hasu_array",
+    f_attr: str = "Ftotal_asu",
+    ccp4_convention: bool = False,
+    sigf_scale: float = 0.2,
```
marcuscollins: better to use longer names for clarity, e.g., sigma_f_scale, structure_factor_column, miller_index_column

src/sampleworks/eval/generate_synthetic_sf.py
```diff
@@ -0,0 +1,634 @@
...
+        is False, which uses Phenix convention (1 = test, 0 = working).
+    sigf_scale: float
+        Scale factor to make a fake SIGFP column from FP values so that
+        SFcalculator can load the output MTZ file as synthetic reward
+        without errors. The actual SIGFP values only matter when
+        computing R-factor.
+    """
+    output_path.parent.mkdir(parents=True, exist_ok=True)
+    dataset = sfc.prepare_dataset(hkl_attr, f_attr)
+    dataset.rename(columns={"FMODEL": "FP"}, inplace=True)
```
marcuscollins: Does SFC expose any of these key strings as constants? If not, please file an issue in their repo and maybe  make a PR there?

src/sampleworks/eval/generate_synthetic_sf.py
```diff
@@ -0,0 +1,634 @@
...
+        If True, use CCP4 convention for R-free flag assignment. Default
+        is False, which uses Phenix convention (1 = test, 0 = working).
+    sigf_scale: float
+        Scale factor to make a fake SIGFP column from FP values so that
+        SFcalculator can load the output MTZ file as synthetic reward
+        without errors. The actual SIGFP values only matter when
+        computing R-factor.
+    """
+    output_path.parent.mkdir(parents=True, exist_ok=True)
+    dataset = sfc.prepare_dataset(hkl_attr, f_attr)
```
marcuscollins: It would help to have a type annotation here.

src/sampleworks/eval/generate_synthetic_sf.py
```diff
@@ -0,0 +1,634 @@
...
+    dataset["SIGFP"] = dataset["FP"] * sigf_scale
+    dataset["SIGFP"] = dataset["SIGFP"].astype(rs.StandardDeviationDtype())
+    if test_fraction > 0:
+        dataset = rs.utils.add_rfree(
+            dataset,
+            ccp4_convention=ccp4_convention,
+            fraction=test_fraction,
+            seed=seed,
+        )
+    dataset.write_mtz(str(output_path))
```
marcuscollins: would you ever do all the things above (prepare the dataset) and _not_ write it out, just use it in-memory? If so, some further factoring-out of methods makes sense.

src/sampleworks/eval/generate_synthetic_sf.py
```diff
@@ -0,0 +1,634 @@
...
+
+    Parameters
+    ----------
+    row
+        BatchRow describing the input structure and optional per-row overrides
+    base_dir
+        Base directory for resolving relative structure file paths
+    output_dir
+        Directory where output MTZ files will be written
+    dmin
```
marcuscollins: give this a more descriptive name.

src/sampleworks/eval/generate_synthetic_sf.py
```diff
@@ -0,0 +1,634 @@
...
+        sfc = SFcalculator(
+            pdbmodel=PDBParser(gemmi_structure),
+            mtzdata=None,
+            dmin=dmin,
+            mode=mode,
+            anomalous=False,
+            set_experiment=False,
+            device=device,
+        )
+        logger.info(
```
marcuscollins: I might move this to `logger.debug`, I don't think you want to write this out all the time.

src/sampleworks/eval/generate_synthetic_sf.py
```diff
@@ -0,0 +1,634 @@
...
+        sfc.calc_fprotein()
+        if simulate_solvent_and_scale:
+            sfc.inspect_data()
+            sfc.calc_fsolvent()
+            sfc.init_scales(requires_grad=False)
+            sfc.calc_ftotal()
+            F_attribute = "Ftotal_asu"
+        else:
+            F_attribute = "Fprotein_asu"
+        logger.info(f"Computed {F_attribute} for {row.filename} on {device}")
```
marcuscollins: move to debug

src/sampleworks/eval/generate_synthetic_sf.py
```diff
@@ -0,0 +1,634 @@
...
+
+    Parameters
+    ----------
+    csv_path
+        Path to CSV file listing structures to process
+    base_dir
+        Base directory for resolving relative structure file paths
+    output_dir
+        Directory where output MTZ files will be written
+    dmin
```
marcuscollins: as above, rename

src/sampleworks/eval/generate_synthetic_sf.py
```diff
@@ -0,0 +1,634 @@
...
+        High-resolution limit in Angstroms
+    mode
```
marcuscollins: should probably rename this too, e.g. "scattering_factor_mode"

src/sampleworks/eval/generate_synthetic_sf.py
```diff
@@ -0,0 +1,634 @@
...
+        Scattering factor type: 'xray' or 'cryoem'
+    occ_mode
```
marcuscollins: write out "occupancy"

src/sampleworks/eval/generate_synthetic_sf.py
```diff
@@ -0,0 +1,634 @@
...
+    strip_waters
+        If True, remove water molecules before computing structure factors.
+    strip_ligands
+        If True, keep only polymer amino-acid atoms (removes ligands and waters).
+    simulate_solvent_and_scale
+        If True, compute bulk solvent and scale factors in addition to F_protein.
+    save_structure
+        If True, save each processed structure as mmCIF to output_dir.
+    """
+    from joblib import delayed, Parallel
```
marcuscollins, line 472: why not import at the top level?

src/sampleworks/eval/generate_synthetic_sf.py
```diff
@@ -0,0 +1,634 @@
...
+            seed=seed,
+            device=device,
+            strip_hydrogens=strip_hydrogens,
+            strip_waters=strip_waters,
+            strip_ligands=strip_ligands,
+            simulate_solvent_and_scale=simulate_solvent_and_scale,
+            save_structure=save_structure,
+        )
+        for row in rows
+    )
```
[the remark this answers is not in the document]
marcuscollins: This is actually a fairly serious issue, but it depends on memory. It will probably be the case that even though you're running in parallel the jobs will all go to one or just a couple GPUs. I would consider specifying the device to use more specifically. It's fine to create an issue for this for now, and ask for help from the engineers.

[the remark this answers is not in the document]
marcuscollins: But please also add a TODO with the issue # here.

tests/eval/test_generate_synthetic_sf.py
```diff
@@ -0,0 +1,122 @@
...
+        PDBParser(gemmi_structure),
+        mtzdata=None,
+        dmin=DMIN,
+        mode="xray",
+        anomalous=False,
+        set_experiment=False,
+        device=device,
+    )
+    sfc.calc_fprotein()
+    return assert_numpy(sfc.Fprotein_asu)
```
marcuscollins, line 73: Are there any numerical tests you can assert? This kind of test really only shows that the code runs without errors, which is a low bar.

tests/eval/test_generate_synthetic_sf.py
```diff
@@ -0,0 +1,122 @@
...
+        assert result.b == pytest.approx(expected.b)
+        assert result.c == pytest.approx(expected.c)
+        assert result.alpha == pytest.approx(expected.alpha)
+        assert result.beta == pytest.approx(expected.beta)
+        assert result.gamma == pytest.approx(expected.gamma)
+
+    def test_space_group_matches_pdb(self, gemmi_structure_from_atomarray, stripped_gemmi):
+        assert gemmi_structure_from_atomarray.spacegroup_hm == stripped_gemmi.spacegroup_hm
+
+    def test_atom_count_matches_pdb(self, gemmi_structure_from_atomarray, stripped_gemmi):
```
marcuscollins: can you test whether the atoms are actually the same? Does order matter anywhere?

tests/eval/test_generate_synthetic_sf.py
```diff
@@ -0,0 +1,122 @@
...
+        assert (
+            gemmi_structure_from_atomarray[0].count_atom_sites()
+            == stripped_gemmi[0].count_atom_sites()
+        )
+
+    def test_occupancy_change_is_applied(self, stripped_atom_array, stripped_gemmi):
```
marcuscollins, line 104: you might also want to test some of the logger warnings related to occupancy are correctly generated.

tests/eval/test_generate_synthetic_sf.py
marcuscollins: I see some below, it makes me wonder whether this test is redundant.

tests/eval/test_generate_synthetic_sf.py
```diff
@@ -0,0 +1,122 @@
...
+    ):
+        f_atomarray = _compute_fprotein(gemmi_structure_from_atomarray, device)
+        f_direct = _compute_fprotein(stripped_gemmi, device)
+        np.testing.assert_allclose(np.abs(f_atomarray), np.abs(f_direct), atol=1e-3)
+
+    def test_fprotein_changes_with_occupancy(self, stripped_atom_array, stripped_gemmi, device):
+        altloc_info = detect_altlocs(stripped_atom_array)
+
+        arr_default = assign_occupancies(stripped_atom_array, altloc_info, "default")
+        f_default = _compute_fprotein(
```
marcuscollins: I would explicitly set the occupancy to something different here, just in case for some reason we change the underlying PDB files at some future point.

pyproject.toml
```diff
@@ -63,13 +64,16 @@ pyarrow = "==17.0.0"
 [tool.pixi.environments]
 analysis = {features = ["analysis"]}
 analysis-dev = {features = ["analysis", "dev"]}
+analysis-dev-sfc = {features = ["analysis", "dev", "sfc"]}
```
marcuscollins: Is there a reason not to just make sfc part of the core requirements (I recall maybe there were dependency conflicts?) At some point we will want to use sfc more generally, right?

marcuscollins (changes requested):
Please have a look at the comments. I don't see anything major but there are some improvements to make. Also, there are merge conflicts that need to be resolved.

src/sampleworks/eval/generate_synthetic_sf.py
DorisMai: yes, [c++ doc](https://gemmi.readthedocs.io/en/latest/api.html#_CPPv4N5gemmi10SpaceGroup2hmE)

src/sampleworks/eval/generate_synthetic_sf.py
DorisMai: SFC has a method `prepare_dataset` which returns an `rs.DataSet` object that is like a pandas dataframe, and I uses this method here. Normally it's just one extra line to write that dataset out as .mtz file. Here this function looks a bit longer to 1) add a fake SigF column, 2) rename column name, 3) add r-free-flag column, because we want the synthetic data to appear like experimental data, kind of a contrived/unusual use case.

src/sampleworks/eval/generate_synthetic_sf.py
DorisMai: I think I ll just add a flag to allow write or just return the modified rs.DataSet object? each modification (add r-free-column, rename column, add sigf) seems few lines of code already

src/sampleworks/eval/generate_synthetic_sf.py
DorisMai: `joblib` is not an explicit dependency in any of the environment. It does exist in `analysis` environment (and maybe `boltz`, but not `rf3`) through other dependency. Since this script might have functions to be imported (e.g. `atomarray2gemmi`) elsewhere, I thought I should make the import lazy. @marcuscollins Would you recommend an alternative way to handle this?

src/sampleworks/eval/generate_synthetic_sf.py
DorisMai: parent directory of structures, such that we don't need to specify the whole path for all structures in each row of the csv file.

src/sampleworks/eval/generate_synthetic_sf.py
DorisMai: SFC doesn't expose the column names "FMODEL" and "PHIFMODEL" as constants. Technically, I don't have to do the renaming, it was more of personal preference / laziness so I don't have to pass the argument expcolumns = ["FMODEL", "SIGFMODEL"]. I am removing the renaming now and "FMODEL" can be auto-detected based on column dtype.

tests/eval/test_generate_synthetic_sf.py
DorisMai: this is not a test, I just imported SFC's `assert_numpy()` function which really just converts torch Tensor (or a list/tuple) to numpy array

pyproject.toml
DorisMai: yes sfc conflicts with protenix, but Minhuan found a workaround by overriding the pandas dependency (#235), I am adding the fix now.  Nevertheless, I am wondering if we should add reward-related dependencies all together in the core requirements. In the future we might have more of dependencies like SFC for other experimental modalities, but users might not care all experimental modalities, and their dependencies might conflict each other. I thought a more composable environment would be safer, though less convenient. @marcuscollins lmk your thoughts

marcuscollins: @DorisMai this is always a thorny issue. One approach is to push others not to pin dependencies down to the sub minor version level. It is almost never actually necessary but there seems to be an increasing trend to do this. I think we should discuss with the engineers what are the possible approaches to additional experimental modalities. It would be nice if we could organize this around added features, like pip does, rather than ending up with a Cartesian product of pixi environments. 

For now, let's try to keep reciprocal space data in the core dependencies.

src/sampleworks/eval/generate_synthetic_sf.py
marcuscollins: At some level that's fine as long as the code never gets called except in the analysis environment, right? Is the issue that it raises an import error in type checking? Or are there tests that fail in certain environments? If so, there are ways to mark the tests to only run under certain conditions. 

In any case, we should include joblib as an explicit dependency, at least in the analysis environments. I thought I had added it, please add it back in if it isn't there.

But I don't feel strongly about it. The case you describe is one of the (few) reasons for inline imports like this.

marcuscollins (approved):
Please go ahead and fix as much of the ty warnings/errors as you can easily, make an issue for the rest (link to the log file on GitHub) and assign to me for further triage. 

I did add a comment on `joblib`--it should be an explicit dependency in the `analysis` environments. I thought it was, but maybe it has been accidentally removed. 

One last thing: please make sure you have docstrings everywhere, including tests. 

I'll approve and leave it up to you to judge when these last few things are done.

PR #240, opened 2026-05-21
ACTL sampleworks image + run_experiments script

manzuoni-astera:
## Summary

This PR makes the ACTL Sampleworks experiment workflow usable from the prebuilt image `pixi-with-checkpoints`. The user-facing entry point is `run_experiments`: it reads TOML presets and launches the right grid-search jobs, pixi environments, GPU assignments, logs, shared data paths, result directories, and MSA cache paths. The dir `/home/dev/workspace` is synced to the local dir that actl is launched from, which should be this sampleworks repo.

## What changed

- Add standalone model presets for `boltz`, `boltz1`, `boltz2`, `boltz2_xrd`, `boltz2_md`, `rf3`, and `protenix`, plus comparison presets.
- Add `gpu_count = N` preset support so jobs auto-assign visible GPUs in order; single-job presets default to all 8 GPUs.
- Move experiment presets into `experiments/` and simplify target syntax (`run_experiments rf3`, `run_experiments boltz`, `run_experiments full_8gpu --jobs rf3,protenix`).
- Prefer baked pixi env Python in ACTL images, with clear failures instead of slow runtime reinstalls.
- Harden ACTL defaults for shared storage, checkpoints, source path selection, cache dirs, dry-run behavior, and runner process cleanup.
- Clarify worker GPU logs by showing both local CUDA index and selected `CUDA_VISIBLE_DEVICES` entry.
- Add focused loader, CLI, and runner tests.

## Using `run_experiments`

In an ACTL Sampleworks pod, from `/home/dev/workspace`:

```bash
run_experiments --list        # show available presets
run_experiments --show boltz  # inspect resolved commands
run_experiments --dry-run rf3 # print commands without running
run_experiments boltz         # run Boltz2 X-ray + MD
run_experiments rf3           # run RF3 on all 8 GPUs by default
run_experiments protenix      # run Protenix on all 8 GPUs by default
```

Presets live in `experiments/*.toml`. The default full run can be subset by job name:

```bash
run_experiments full_8gpu --jobs rf3,protenix
```

Useful launch-time overrides:

```bash
run_experiments rf3 --set jobs.rf3.gpu_count=4
run_experiments rf3 --set jobs.rf3.args.gradient-weights="0.0 0.01 0.02"
```

By default, data comes from `/mnt/diffuse-shared/raw/sampleworks`, results go under `/mnt/diffuse-shared/results/sampleworks/<pod>/<target>/`, and MSA cache goes under `/mnt/diffuse-shared/cache/sampleworks/msa`. Override with `DATA_DIR`, `RESULTS_DIR`, or `MSA_CACHE_DIR`.


manzuoni-astera:
CI is currently blocked waiting for `gpu-testing` environment approval before the GPU test jobs can start. GitHub lists @marcuscollins / @k-chrispens as eligible reviewers for the pending deployment. The baked-image smoke verification already passed on ACTL with `run_all_models.sh --dry-run` against `harbor.astera.sh/library/sampleworks:cuda12.4-2026-05-21-pr236-scientist3`.

src/sampleworks/models/boltz/wrapper.py
```diff
@@ -360,6 +360,9 @@ def process_structure_for_boltz(
     if recycling_steps is None:
         recycling_steps = 3
 
+    # Keep Boltz dataloading in-process by default. Kubernetes pods usually get
```
marcuscollins, line 363: Have you tested what performance impact this might have? I guess it isn't big since we're only doing inference, but might be worth checking.

marcuscollins: Also, it appears we're only doing this for Boltz? Was this the only one with a non-zero default?

README.md
```diff
@@ -152,6 +152,49 @@ Output layout: `grid_search_results/<protein>/<model>[_<method>]/<scaler>/ens<N>
 Instructions for running evaluation and metrics scripts are coming soon.
 
 
+## ACTL preset experiments (`run_experiments`)
+
+Use ACTL to get one 8-GPU pod with baked pixi environments, checkpoints, the
```
marcuscollins: This is going to be unique to Astera, not accessible externally, but we want others to be able to run these grid search evaluations as well. Will that still be possible? Can you make this clear so external users won't get confused?

README.md
```diff
@@ -152,6 +152,49 @@ Output layout: `grid_search_results/<protein>/<model>[_<method>]/<scaler>/ens<N>
...
+MD, RF3, and Protenix.
+
+Presets live in the synced repo at `src/sampleworks/runs/presets/*.toml`. To change an experiment, either edit/copy a preset locally and let ACTL sync it, or override values at launch:
+
+```bash
+run_experiments all_models --only rf3,protenix
+run_experiments rf3_partial --set jobs.rf3.gpus=0
+```
+
+On smaller pods, make sure preset GPU IDs only reference visible pod GPUs
```
marcuscollins: When playing around with this with Mag, it seems that this script was trying to access GPUs that clearly didn't exist. (4-7 on a 4-GPU pod), and it didn't fail fast. Might want to double check that this works as expected.

run_experiments
```diff
@@ -0,0 +1,225 @@
+#!/usr/bin/env bash
+# ACTL-native entry point for Sampleworks preset runs.
+#
+# The TOML preset is the source of truth. This wrapper uses the synced ACTL
+# checkout at /home/dev/workspace for Sampleworks code, while reusing the
+# prebuilt pixi environments/checkpoints from the image at /app.
+
+set -euo pipefail
+
+script_path="${BASH_SOURCE[0]}"
```
marcuscollins, line 13: Please add a comment explaining what this block does. I have the vague sense that you're trying to find the scripts directory, but it isn't obvious to me.

run_experiments
```diff
@@ -0,0 +1,225 @@
...
+
+resolve_repo_root() {
+    local source_override="${SAMPLEWORKS_SOURCE_DIR:-}"
+    if [[ -n "$source_override" ]]; then
+        printf '%s\n' "$source_override"
+        return 0
+    fi
+
+    # ACTL syncs the user's local checkout here. Prefer it over any stale code
+    # that may have been used only to build the image's pixi environments.
```
marcuscollins, line 88: What happens if the user has new package requirements in their local checkout?

run_experiments
```diff
@@ -0,0 +1,225 @@
...
+if [[ "$preset" == *.toml || "$preset" == */* ]]; then
+    if [[ "$preset" != /* ]]; then
+        preset="$repo_root/$preset"
+    fi
+fi
+preset_label="${preset##*/}"
+preset_label="${preset_label%.toml}"
+
+run_name="${SAMPLEWORKS_ACTL_RUN_NAME:-$(hostname -s 2>/dev/null || printf 'sampleworks')}"
+default_data_dir="/mnt/diffuse-shared/raw/sampleworks/initial_dataset_40_occ_sweeps"
```
marcuscollins: I think I prefer that we set this explicitly each time we run, so that it is easily captured and recorded.

run_experiments
```diff
@@ -0,0 +1,225 @@
...
+default_results_dir="/mnt/diffuse-shared/results/sampleworks/${run_name}/${preset_label}"
+default_msa_cache_dir="/mnt/diffuse-shared/cache/sampleworks/msa"
+
+export DATA_DIR="${DATA_DIR:-${SAMPLEWORKS_DATA_DIR:-$default_data_dir}}"
```
marcuscollins: correspondingly, this should have no default and raise an error if unset.

run_experiments
```diff
@@ -0,0 +1,225 @@
...
+export RESULTS_DIR="${RESULTS_DIR:-${SAMPLEWORKS_RESULTS_DIR:-$default_results_dir}}"
+export MSA_CACHE_DIR="${MSA_CACHE_DIR:-${SAMPLEWORKS_MSA_CACHE_DIR:-$default_msa_cache_dir}}"
+export SAMPLEWORKS_GRID_SEARCH_SCRIPT="${SAMPLEWORKS_GRID_SEARCH_SCRIPT:-$repo_root/run_grid_search.py}"
+export PYTHONPATH="$repo_root/src${PYTHONPATH:+:$PYTHONPATH}"
```
marcuscollins, line 247: Can you explain this one? Why no ':' after $repo_root/src? And what does `${PYTHONPATH:+:$PYTHONPATH}` mean? I'm just curious having never seen this construction before.

run_experiments
```diff
@@ -0,0 +1,225 @@
...
+export PIXI_CACHE_DIR="${PIXI_CACHE_DIR:-/tmp/pixi-cache}"
+export UV_CACHE_DIR="${UV_CACHE_DIR:-/tmp/uv-cache}"
+# The ACTL image is expected to provide ready-to-use pixi envs under /app/.pixi.
+# Do not let sampleworks.runs.runner call `pixi run` just to "prepare" envs;
+# that can reinstall the CUDA stack inside the pod, especially when the synced
```
marcuscollins: Is the risk of reinstalling CUDA that it will break (I've seen that frequently) or just time? 

I'm concerned that this doesn't leave us with a way to easily test code that has new dependencies.

marcuscollins: Or do we unset `SAMPLEWORKS_REQUIRE_PREBUILT_PIXI` to use the sync'd pixi environment?

run_experiments
```diff
@@ -0,0 +1,225 @@
...
+    "RF3_CHECKPOINT rf3_foundry_01_24_latest.ckpt" \
+    "PROTENIX_CHECKPOINT protenix_base_default_v0.5.0.pt"; do
+    read -r checkpoint_var checkpoint_file <<<"$checkpoint_var_and_file"
+    checkpoint_path="$shared_checkpoint_dir/$checkpoint_file"
+    if [[ -z "${!checkpoint_var:-}" && -f "$checkpoint_path" ]]; then
+        export "$checkpoint_var=$checkpoint_path"
+    fi
+done
+
+source_proteins_csv="${PROTEINS_CSV:-$DATA_DIR/proteins.csv}"
```
marcuscollins: This is another thing we should probably be explicit about, rather than having defaults.

run_experiments
```diff
@@ -0,0 +1,194 @@
...
+runner_python="${SAMPLEWORKS_RUNNER_PYTHON:-$pixi_project_dir/.pixi/envs/$runner_env/bin/python}"
+
+needs_runtime_paths=1
+for arg in "$@"; do
+    case "$arg" in
+        --dry-run|--show|--list|-h|--help)
+            needs_runtime_paths=0
+            ;;
+    esac
+done
```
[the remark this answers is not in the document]
marcuscollins: Definitely should fix this.

run_grid_search.py
```diff
@@ -69,7 +69,28 @@ def get_job_status(job: JobConfig) -> str:
 def detect_gpus() -> list[str]:
     cuda_visible = os.environ.get("CUDA_VISIBLE_DEVICES", "")
     if cuda_visible:
-        return [g.strip() for g in cuda_visible.split(",") if g.strip()]
+        gpus = [g.strip() for g in cuda_visible.split(",") if g.strip()]
+        try:
+            result = subprocess.run(
+                ["nvidia-smi", "--query-gpu=index", "--format=csv,noheader"],
```
marcuscollins: If possible, can we do this with actual python modules? I thought there was a way to get this information without having to shell out?

run_grid_search.py
```diff
@@ -69,7 +69,28 @@ def get_job_status(job: JobConfig) -> str:
...
+                    missing = sorted(set(gpus).difference(visible), key=int)
+                    if missing:
+                        raise ValueError(
+                            "CUDA_VISIBLE_DEVICES references GPUs that are not visible "
+                            f"in this container: {missing}. Visible GPUs: {visible}. "
+                            "Check the preset jobs.*.gpus values for this pod size."
+                        )
+        except FileNotFoundError:
+            pass
+        return gpus
```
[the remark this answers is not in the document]
marcuscollins: Definitely should fix this, just in case.

run_grid_search.py
```diff
@@ -69,7 +69,28 @@ def get_job_status(job: JobConfig) -> str:
 def detect_gpus() -> list[str]:
     cuda_visible = os.environ.get("CUDA_VISIBLE_DEVICES", "")
     if cuda_visible:
-        return [g.strip() for g in cuda_visible.split(",") if g.strip()]
+        gpus = [g.strip() for g in cuda_visible.split(",") if g.strip()]
+        try:
```
marcuscollins: Some of this try/except seems to be repeated below. Do we need both?

run_grid_search.py
```diff
@@ -223,16 +244,86 @@ def run_guidance_queue_script(args: tuple[str, int, str, int]):
...
     pixi_env = get_pixi_env(model)
     script_path = Path(__file__).parent / "scripts" / "run_guidance_pipeline.py"
-    cmd = f"pixi run -e {pixi_env} python {script_path} --job-queue-path {job_queue_path}"
-    cmd = cmd.split()
+    env_python = get_pixi_env_python(pixi_env)
+    if env_python:
+        cmd = [env_python, str(script_path), "--job-queue-path", job_queue_path]
+        env = get_pixi_env_process_env(env_python)
+    else:
+        cmd = [
```
marcuscollins, line 336: can you do this the way I had it before (the original 226/227)? I find it much easier to read.

run_grid_search.py
```diff
@@ -223,16 +244,86 @@ def run_guidance_queue_script(args: tuple[str, int, str, int]):
...
+            "--job-queue-path",
+            job_queue_path,
+        ]
+        env = os.environ.copy()
     log.info(f"Running worker {worker_num}: {cmd} on GPU {worker_num % max_workers}")
-    # env = os.environ.copy()
 
     with open(job_queue_path.replace(".pkl", ".log"), "w") as log_file:
-        result = subprocess.run(cmd, stdout=log_file, stderr=subprocess.STDOUT)
+        result = subprocess.run(cmd, stdout=log_file, stderr=subprocess.STDOUT, env=env)
```
marcuscollins: this is a pixi environment, right? How does the `env` keyword work here?

marcuscollins (changes requested):

run_experiments
xraymemory: It prepends the synced checkout’s src to PYTHONPATH, and only adds :$PYTHONPATH if PYTHONPATH was already set. That avoids leaving a trailing :, which Python treats like “also search the current directory.”

src/sampleworks/models/boltz/wrapper.py
xraymemory: I haven’t benchmarked it formally, but this only affects Boltz preprocessing/featurization, not the model forward pass, so the expected impact is small compared with inference. It’s Boltz-only because this is the only wrapper exposing that DataLoader-style worker knob here; RF3/Protenix don’t have an equivalent multiprocessing default in this path.

README.md
xraymemory: Theoretically yes they could still run it using the pixi-with-checkpoints dockerfile. I have expanded upon this in the comments

README.md
xraymemory: This has been fixed

run_experiments
xraymemory: This resolves the real path of the script even if it was invoked through a symlink like /usr/local/bin/run_experiments. We use that location later as a fallback for finding the Sampleworks checkout.

run_experiments
xraymemory: So this is a question I had for you. How often do we expect the pixi env to be changed / updated? Currently we detect that case by comparing the synced checkout’s pyproject.toml and pixi.lock against the baked /app copies. By default we fail with a clear “rebuild/use a matching image” message; for intentional dev/debugging, SAMPLEWORKS_ALLOW_RUNTIME_PIXI=1 switches to the synced checkout and lets pixi update on the pod. 

However we can change that so ALLOW_RUNTIME_PIXI is 1 by default and will always update

run_experiments
xraymemory: Yes - exactly, we can set the RUNTIME_PIXI env var. I wasn't sure how often you'd be updating the pixi environment so we can make that the default if you expect it to change often

run_grid_search.py
xraymemory: Yes so when we bypass pixi run and call /app/.pixi/envs/<env>/bin/python directly, env= passes the subprocess an environment that looks like that pixi env is activated (PATH, CONDA_PREFIX, CUDA_HOME, etc.). If we fall back to pixi run, env= is just the normal process environment and pixi handles activation itself.

marcuscollins (approved):
Thanks for the changes. Approved.

PR #244, opened 2026-05-27
Split public and Astera images

manzuoni-astera:
## Summary
- split the public `pixi-with-checkpoints` Docker build into explicit multi-stage targets
- add a private `Dockerfile.astera` overlay that starts from the public `pixi-with-checkpoints` image and adds EXT/workspace conveniences for Harbor
- update Docker CI to publish the public image first, then build the Harbor `library/pixi-with-checkpoints` image from the exact public SHA tag
- refresh docs and image names to keep `pixi-with-checkpoints` as the runtime image

## Validation
- `git diff --check`
- `bash -n docker-entrypoint.sh && bash -n run_experiments && bash -n run_experiments.sh && bash -n run_all_models.sh`
- `python3 -m py_compile src/sampleworks/runs/runner.py`
- `ruby -e "require \"yaml\"; YAML.load_file(".github/workflows/docker.yml")"`
- `docker buildx build --target source-check --build-arg BASE_IMAGE=ubuntu:22.04 -t sampleworks-source-check --load .`
- `docker run --rm sampleworks-source-check --help`

Full image build was not run locally because it pulls the checkpoint layer and installs all pixi environments.

## Summary by CodeRabbit

* **New Features**
  * Builds now accept configurable base and checkpoint images for flexible deployments
  * Added a private overlay image that layers developer tools and runtime extensions on the public base
  * CI now mirrors checkpoint images to a public registry and publishes a digest-pinned public image reference
  * Added a fast "smoke" build target for quick container sanity checks

* **Documentation**
  * Updated docs and examples for the two-layer build, concrete build commands, and container startup/help usage

[Review Change Stack](https://app.coderabbit.ai/change-stack/diff-use/sampleworks/pull/244?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -17,7 +17,6 @@ env:
   # public registry path.
   PUBLIC_REGISTRY: ${{ vars.SAMPLEWORKS_PUBLIC_REGISTRY || 'docker.io' }}
   PUBLIC_IMAGE_NAME: ${{ vars.SAMPLEWORKS_PUBLIC_IMAGE || 'diffuseproject/pixi-with-checkpoints' }}
-  CHECKPOINTS_IMAGE: ${{ vars.SAMPLEWORKS_CHECKPOINTS_IMAGE }}
   CHECKPOINTS_SOURCE_IMAGE: ${{ vars.SAMPLEWORKS_CHECKPOINTS_SOURCE_IMAGE || 'harbor.astera.sh/library/sampleworks-checkpoints@sha256:ff1dd2b42c3afd593e034cf81df1bf1862c27ae491f46cc10e105629f1102781' }}
   CHECKPOINTS_DOCKERHUB_IMAGE: ${{ vars.SAMPLEWORKS_CHECKPOINTS_DOCKERHUB_IMAGE || 'docker.io/diffuseproject/sampleworks-checkpoints:latest' }}
   CUDA_BASE_IMAGE: ${{ vars.SAMPLEWORKS_CUDA_BASE_IMAGE || 'nvidia/cuda:12.4.1-devel-ubuntu22.04@sha256:da6791294b0b04d7e65d87b7451d6f2390b4d36225ab0701ee7dfec5769829f5' }}
@@ -31,11 +30,13 @@ env:
   DOCKER_METADATA_SHORT_SHA_LENGTH: 7
 
 jobs:
-  sync-checkpoints:
+  sync_checkpoints:
     name: Sync checkpoints to Docker Hub
     runs-on: ubuntu-latest
     permissions:
       contents: read
+    outputs:
+      checkpoints_image: ${{ steps.verify_checkpoints.outputs.checkpoints_image }}
 
     steps:
       - name: Install skopeo
@@ -54,6 +55,10 @@ jobs:
             echo "Checkpoint source and Docker Hub destination image refs must be set."
             exit 1
           fi
+          if [ "${CHECKPOINTS_SOURCE_IMAGE}" = "${CHECKPOINTS_SOURCE_IMAGE%@*}" ]; then
+            echo "CHECKPOINTS_SOURCE_IMAGE must be digest-pinned."
+            exit 1
+          fi
           if [ -z "${HARBOR_USERNAME}" ] || [ -z "${HARBOR_PASSWORD}" ]; then
             echo "HARBOR_USERNAME and HARBOR_PASSWORD secrets are required to pull checkpoint image."
             exit 1
@@ -70,6 +75,7 @@ jobs:
             "docker://${CHECKPOINTS_DOCKERHUB_IMAGE}"
 
       - name: Verify Docker Hub checkpoint digest
+        id: verify_checkpoints
         env:
           DOCKERHUB_USERNAME: ${{ secrets.DOCKERHUB_USERNAME }}
           DOCKERHUB_TOKEN: ${{ secrets.DOCKERHUB_TOKEN }}
@@ -85,10 +91,14 @@ jobs:
             exit 1
           fi
 
+          dockerhub_ref="${CHECKPOINTS_DOCKERHUB_IMAGE%@*}"
+          dockerhub_ref="${dockerhub_ref%:*}"
+          echo "checkpoints_image=${dockerhub_ref}@${actual_digest}" >> "$GITHUB_OUTPUT"
+
   public:
     name: Public pixi-with-checkpoints image
     runs-on: ubuntu-latest
-    needs: sync-checkpoints
+    needs: sync_checkpoints
     permissions:
       contents: read
     outputs:
@@ -131,10 +141,12 @@ jobs:
             type=semver,pattern={{version}}
             type=semver,pattern=v{{version}}
 
-      - name: Validate checkpoint image input
+      - name: Validate mirrored checkpoint image input
+        env:
+          CHECKPOINTS_IMAGE: ${{ needs.sync_checkpoints.outputs.checkpoints_image }}
         run: |
           if [ -z "${CHECKPOINTS_IMAGE}" ]; then
-            echo "SAMPLEWORKS_CHECKPOINTS_IMAGE must be set to a checkpoint image ref, preferably digest-pinned."
+            echo "sync-checkpoints did not produce a digest-pinned checkpoint image ref."
             exit 1
           fi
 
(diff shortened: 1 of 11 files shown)


Dockerfile.astera
```diff
@@ -0,0 +1,47 @@
...
+    && mkdir -p /home/dev/.config /home/dev/.cache /home/dev/.local/share /home/dev/workspace \
+    && install -m 0755 /app/run_experiments /usr/local/bin/run_experiments \
+    && install -m 0755 /app/run_experiments.sh /usr/local/bin/run_experiments.sh \
+    && install -m 0755 /app/run_all_models.sh /usr/local/bin/run_all_models.sh
+
+COPY --from=ext-cli /ext /usr/local/bin/ext
+
+RUN chmod 0755 /usr/local/bin/ext
+
+# Inherit Sampleworks' entrypoint and default command from the public base.
```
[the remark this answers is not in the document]
manzuoni-astera: Leaving the runtime user as root intentionally for this ACTL image. The inherited /app/.pixi environments are root-owned and current DevSpace/shared PVC workflows assume the existing root-oriented runtime behavior; switching USER here would be a behavior change that can break scientist runs. Added an explicit Dockerfile comment documenting that constraint.

Dockerfile
```diff
@@ -1,149 +1,118 @@
...
+# pushed to the public registry.
+
+# Public default. Astera CI may override this with its Docker Hub cache mirror,
+# e.g. harbor.astera.sh/dockerhub-cache/nvidia/cuda:12.4.1-devel-ubuntu22.04.
+ARG BASE_IMAGE=nvidia/cuda:12.4.1-devel-ubuntu22.04
+
+# Public default for the large checkpoint layer. Private builders can override
+# this with an internal mirror, but the public Dockerfile itself does not depend
+# on Harbor.
+ARG CHECKPOINTS_IMAGE=diffuseproject/sampleworks-checkpoints:latest
```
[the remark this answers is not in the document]
manzuoni-astera: Addressed without hard-coding an unavailable public checkpoint digest: BASE_IMAGE now defaults to a digest-pinned CUDA manifest, CHECKPOINTS_IMAGE no longer has a mutable tag default (source-check uses scratch), and CI validates that SAMPLEWORKS_CHECKPOINTS_IMAGE is explicitly configured so it can be set to the digest-pinned public checkpoint image when that artifact exists.

.github/workflows/docker.yml
```diff
@@ -13,14 +13,27 @@ on:
...
+  # Override these with repository variables if Diffuse Hub uses a different
+  # public registry path.
+  PUBLIC_REGISTRY: ${{ vars.SAMPLEWORKS_PUBLIC_REGISTRY || 'docker.io' }}
+  PUBLIC_IMAGE_NAME: ${{ vars.SAMPLEWORKS_PUBLIC_IMAGE || 'diffuseproject/pixi-with-checkpoints' }}
+  CHECKPOINTS_IMAGE: ${{ vars.SAMPLEWORKS_CHECKPOINTS_IMAGE }}
+  CUDA_BASE_IMAGE: ${{ vars.SAMPLEWORKS_CUDA_BASE_IMAGE || 'nvidia/cuda:12.4.1-devel-ubuntu22.04@sha256:da6791294b0b04d7e65d87b7451d6f2390b4d36225ab0701ee7dfec5769829f5' }}
+
+  ASTERA_REGISTRY: harbor.astera.sh
+  ASTERA_IMAGE_NAME: library/pixi-with-checkpoints
+  EXT_CLI_IMAGE: ${{ vars.EXT_CLI_IMAGE || 'harbor.astera.sh/library/ext-cli:sha-724abde27fdfa15b23adb112d2e1dc83dd533a57' }}
```
marcuscollins, line 28: does this sha need to be updated every time we update the image?

marcuscollins: Oh, is this the checkpoints image? I guess that won't change often.

Dockerfile
```diff
@@ -1,149 +1,124 @@
...
     pixi install -e rf3 --frozen
 
-# ============================================================================
-# Pre-compile CUDA extensions to avoid JIT compilation at runtime
-# This triggers the dilate_points_cuda extension build
-# ============================================================================
-RUN pixi run -e boltz python -c "\
+# Pre-compile CUDA extensions only when the builder exposes NVIDIA devices.
+# If a GPU is present, failures should stop the build instead of being masked.
+RUN if [ ! -e /dev/nvidiactl ] && [ ! -e /proc/driver/nvidia/version ]; then \
```
marcuscollins, line 109: I guess this means we can only build a GPU image on an instance with GPUs... but maybe that was the case anyway?

Dockerfile.astera
```diff
@@ -0,0 +1,50 @@
...
+    XDG_DATA_HOME=/home/dev/.local/share \
+    SHELL=/bin/bash \
+    SAMPLEWORKS_PIXI_PROJECT_DIR=/app \
+    SAMPLEWORKS_REQUIRE_PREBUILT_PIXI=1 \
+    SAMPLEWORKS_SKIP_ENV_PREPARE=1
+
+RUN apt-get update && apt-get install -y --no-install-recommends \
+        nano \
+        rsync \
+        vim \
```
marcuscollins, line 36: could we add `emacs`? :)

marcuscollins (approved):
Some minor comments/questions, but LGTM.

Dockerfile.astera
xraymemory: Added terminal Emacs via `emacs-nox`, which provides the `emacs` command without pulling the GTK dependency chain into the ACTL image.

.github/workflows/docker.yml
xraymemory: This SHA is for the EXT CLI overlay image, not the checkpoints/public image. It only needs changing when we intentionally want a newer EXT CLI; otherwise the repo variable `EXT_CLI_IMAGE` can override it without editing the workflow. Added a workflow comment to make that explicit.

Dockerfile
xraymemory: No, a GPU is not required to build the image. The Dockerfile skips CUDA extension pre-compilation when no NVIDIA device is visible; if a GPU is visible, compile failures stop the build instead of being masked. Added a comment to make this explicit.

xraymemory (approved):

.github/workflows/docker.yml
```diff
@@ -13,15 +13,87 @@ on:
   workflow_dispatch:
 
 env:
-  REGISTRY: harbor.astera.sh
-  IMAGE_NAME: library/pixi-with-checkpoints
+  # Override these with repository variables if Diffuse Hub uses a different
+  # public registry path.
+  PUBLIC_REGISTRY: ${{ vars.SAMPLEWORKS_PUBLIC_REGISTRY || 'docker.io' }}
+  PUBLIC_IMAGE_NAME: ${{ vars.SAMPLEWORKS_PUBLIC_IMAGE || 'diffuseproject/pixi-with-checkpoints' }}
+  CHECKPOINTS_IMAGE: ${{ vars.SAMPLEWORKS_CHECKPOINTS_IMAGE }}
```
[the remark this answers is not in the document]
manzuoni-astera: Addressed in a3b2cce. The sync_checkpoints job now verifies the Docker Hub mirror digest, emits a digest-pinned checkpoints_image output, and the public Docker build consumes needs.sync_checkpoints.outputs.checkpoints_image as CHECKPOINTS_IMAGE instead of reading a separate repo variable.

.github/workflows/docker.yml
```diff
@@ -13,15 +13,87 @@ on:
...
+  CHECKPOINTS_SOURCE_IMAGE: ${{ vars.SAMPLEWORKS_CHECKPOINTS_SOURCE_IMAGE || 'harbor.astera.sh/library/sampleworks-checkpoints@sha256:ff1dd2b42c3afd593e034cf81df1bf1862c27ae491f46cc10e105629f1102781' }}
+  CHECKPOINTS_DOCKERHUB_IMAGE: ${{ vars.SAMPLEWORKS_CHECKPOINTS_DOCKERHUB_IMAGE || 'docker.io/diffuseproject/sampleworks-checkpoints:latest' }}
```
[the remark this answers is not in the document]
manzuoni-astera: Addressed in a3b2cce. sync_checkpoints now emits a digest-pinned Docker Hub checkpoint ref and public uses that output for the CHECKPOINTS_IMAGE build arg; the stale SAMPLEWORKS_CHECKPOINTS_IMAGE env path was removed.

PR #245, opened 2026-05-28
feat(runs): add analysis preset harness

xraymemory:
## Summary
- add analysis preset TOML configs and run_analysis entrypoints
- extend sampleworks-runs loading/execution support for analysis presets
- document evaluation harness usage and add runner coverage

## Tests
- Not run locally (PR creation only).

## Summary by CodeRabbit

* **New Features**
  * Added analysis workflow system via `run_analysis` command for preset-driven evaluation jobs
  * Added support for grid-search analysis, altloc detection/classification, and external evaluation tools
  * New CLI options: `--depth` for controlling CIF file search depth; `--skip-pre-jobs` to re-run analyses without repeating patching steps

* **Documentation**
  * Updated evaluation guide with analysis workflow examples and configuration details

* **Tests**
  * Expanded test coverage for analysis orchestration and CLI validation

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -112,7 +112,8 @@ RUN if [ ! -e /dev/nvidiactl ] && [ ! -e /proc/driver/nvidia/version ]; then \
     else \
         pixi run -e boltz python -c "\
 from sampleworks.core.forward_models.xray.real_space_density_deps.ops import dilate_atom_centric; \
-print('CUDA extensions compiled successfully')" || echo "CUDA extension pre-compilation skipped (no GPU during build)"
+print('CUDA extensions compiled successfully')"; \
+    fi
 
 # This image carries pixi environments and checkpoints. Runtime source should
 # come from ACTL's synced checkout at /home/dev/workspace, not from stale code
--- a/Dockerfile
+++ b/Dockerfile
@@ -120,7 +120,8 @@ COPY --from=harbor.astera.sh/library/sampleworks-checkpoints:latest /checkpoints
 # smaller CI runners (ubuntu-latest can be 72 GB or 145 GB).
 RUN pixi install -e boltz --frozen && \
     pixi install -e protenix --frozen && \
-    pixi install -e rf3 --frozen
+    pixi install -e rf3 --frozen && \
+    pixi install -e analysis --frozen
 
 # ============================================================================
 # Pre-compile CUDA extensions to avoid JIT compilation at runtime
@@ -133,10 +134,11 @@ print('CUDA extensions compiled successfully')" || echo "CUDA extension pre-comp
 # This image carries pixi environments and checkpoints. Runtime source should
 # come from ACTL's synced checkout at /home/dev/workspace, not from stale code
 # baked into /app during image construction.
-RUN rm -rf /app/src /app/scripts /app/experiments /app/run_grid_search.py \
+RUN rm -rf /app/src /app/scripts /app/experiments /app/analyses \
+    /app/run_grid_search.py /app/run_analysis \
     && mkdir -p /home/dev/workspace
 
-COPY --chmod=755 run_experiments run_experiments.sh run_all_models.sh /usr/local/bin/
+COPY --chmod=755 run_experiments run_experiments.sh run_all_models.sh run_analysis run_analysis.sh /usr/local/bin/
 RUN printf '\n# ACTL scientist workflow: land in the synced Sampleworks checkout.\nif [[ $- == *i* ]] && [ -z "${SAMPLEWORKS_NO_AUTO_CD:-}" ] && [ -d /home/dev/workspace ]; then\n    cd /home/dev/workspace\nfi\n' >> /root/.bashrc
 
 ENV SAMPLEWORKS_PIXI_PROJECT_DIR=/app \
--- a/analyses/altloc_classify.toml
+++ b/analyses/altloc_classify.toml
@@ -1,11 +1,11 @@
 description = "Classify altloc selections into side-chain, loop, and domain-shift categories."
 
 [defaults]
-GRID_SEARCH_RESULTS_DIR = "/data/results/grid_search_results"
-GRID_SEARCH_INPUTS_DIR = "/data/inputs"
-ALTLOC_SELECTIONS_CSV = "${GRID_SEARCH_RESULTS_DIR}/analysis/altloc_selections.csv"
-ALTLOC_CLASSIFICATIONS_CSV = "${GRID_SEARCH_RESULTS_DIR}/analysis/altloc_region_classifications.csv"
-CIF_ROOT = "${GRID_SEARCH_INPUTS_DIR}"
+ALTLOC_ANALYSIS_DIR = "/data/results/altloc_analysis"
+ALTLOC_INPUTS_DIR = "/data/inputs"
+ALTLOC_SELECTIONS_CSV = "${ALTLOC_ANALYSIS_DIR}/altloc_selections.csv"
+ALTLOC_CLASSIFICATIONS_CSV = "${ALTLOC_ANALYSIS_DIR}/altloc_region_classifications.csv"
+CIF_ROOT = "${ALTLOC_INPUTS_DIR}"
 DOMAIN_SHIFT_MIN_SPAN = "50"
 LOOP_LDDT_THRESHOLD = "0.75"
 
--- a/analyses/altloc_classify.toml
+++ b/analyses/altloc_classify.toml
@@ -1,0 +1,25 @@
+description = "Classify altloc selections into side-chain, loop, and domain-shift categories."
+
+[defaults]
+GRID_SEARCH_RESULTS_DIR = "/data/results/grid_search_results"
+GRID_SEARCH_INPUTS_DIR = "/data/inputs"
+ALTLOC_SELECTIONS_CSV = "${GRID_SEARCH_RESULTS_DIR}/analysis/altloc_selections.csv"
+ALTLOC_CLASSIFICATIONS_CSV = "${GRID_SEARCH_RESULTS_DIR}/analysis/altloc_region_classifications.csv"
+CIF_ROOT = "${GRID_SEARCH_INPUTS_DIR}"
+DOMAIN_SHIFT_MIN_SPAN = "50"
+LOOP_LDDT_THRESHOLD = "0.75"
+
+[shared_args]
+input-csv = "${ALTLOC_SELECTIONS_CSV}"
+cif-root = "${CIF_ROOT}"
+output-file = "${ALTLOC_CLASSIFICATIONS_CSV}"
+domain-shift-min-span = "${DOMAIN_SHIFT_MIN_SPAN}"
+loop-lddt-threshold = "${LOOP_LDDT_THRESHOLD}"
+
+[[jobs]]
+name = "classify_altloc_regions"
+env = "analysis"
+gpus = "none"
+script = "scripts/eval/classify_altloc_regions.py"
+output_arg = ""
+output_subdir = "analysis/altloc_classify"
--- a/analyses/altloc_find.toml
+++ b/analyses/altloc_find.toml
@@ -1,10 +1,10 @@
 description = "Build an analysis protein-config CSV by finding altloc selections in input CIFs."
 
 [defaults]
-GRID_SEARCH_RESULTS_DIR = "/data/results/grid_search_results"
-GRID_SEARCH_INPUTS_DIR = "/data/inputs"
-PROTEINS_CSV = "${GRID_SEARCH_INPUTS_DIR}/proteins.csv"
-ALTLOC_SELECTIONS_CSV = "${GRID_SEARCH_RESULTS_DIR}/analysis/altloc_selections.csv"
+ALTLOC_ANALYSIS_DIR = "/data/results/altloc_analysis"
+ALTLOC_INPUTS_DIR = "/data/inputs"
+PROTEINS_CSV = "${ALTLOC_INPUTS_DIR}/proteins.csv"
+ALTLOC_SELECTIONS_CSV = "${ALTLOC_ANALYSIS_DIR}/altloc_selections.csv"
 ALTLOC_MIN_SPAN = "5"
 ALTLOC_LABEL = "label_alt_id"
 
--- a/analyses/altloc_find.toml
+++ b/analyses/altloc_find.toml
@@ -1,0 +1,23 @@
+description = "Build an analysis protein-config CSV by finding altloc selections in input CIFs."
+
+[defaults]
+GRID_SEARCH_RESULTS_DIR = "/data/results/grid_search_results"
+GRID_SEARCH_INPUTS_DIR = "/data/inputs"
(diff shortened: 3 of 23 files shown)


analyses/altloc_classify.toml
```diff
@@ -0,0 +1,25 @@
+description = "Classify altloc selections into side-chain, loop, and domain-shift categories."
+
+[defaults]
+GRID_SEARCH_RESULTS_DIR = "/data/results/grid_search_results"
```
marcuscollins: this script shouldn't need the results of the grid search. We use it to come up with the parts of protein structures to analyze in more detail.

marcuscollins: Does the script run without it?

analyses/altloc_classify.toml
```diff
@@ -0,0 +1,25 @@
...
+GRID_SEARCH_INPUTS_DIR = "/data/inputs"
+ALTLOC_SELECTIONS_CSV = "${GRID_SEARCH_RESULTS_DIR}/analysis/altloc_selections.csv"
+ALTLOC_CLASSIFICATIONS_CSV = "${GRID_SEARCH_RESULTS_DIR}/analysis/altloc_region_classifications.csv"
+CIF_ROOT = "${GRID_SEARCH_INPUTS_DIR}"
+DOMAIN_SHIFT_MIN_SPAN = "50"
+LOOP_LDDT_THRESHOLD = "0.75"
+
+[shared_args]
```
marcuscollins, line 12: I find this name, "shared_args" a little confusing. These are just the arguments to the script, right?

analyses/altloc_find.toml
```diff
@@ -0,0 +1,23 @@
+description = "Build an analysis protein-config CSV by finding altloc selections in input CIFs."
+
+[defaults]
+GRID_SEARCH_RESULTS_DIR = "/data/results/grid_search_results"
```
marcuscollins: this one also shouldn't need this variable.

analyses/grid_search.toml
```diff
@@ -0,0 +1,53 @@
+description = "Core grid-search evaluations: RSCC, LDDT clustering, and bond geometry."
```
marcuscollins: Rename this preset--it makes me think of running the grid search generation code. How about "analyze_grid_search.toml"?

src/sampleworks/runs/cli.py
```diff
@@ -5,16 +5,87 @@
 import argparse
 import os
 import sys
+from dataclasses import dataclass
 from pathlib import Path
 
 from . import loader, runner
-from .schema import Preset
+from .schema import Job, Preset
```
marcuscollins: Generally I favor absolute imports.

src/sampleworks/runs/schema.py
```diff
@@ -68,6 +93,10 @@ def __post_init__(self) -> None:
             raise ValueError(f"Job {self.name!r}: gpu_count must be positive")
         if not self.output_subdir:
             raise ValueError(f"Job {self.name!r}: output_subdir must be non-empty")
+        if self.output_arg.startswith("--"):
```
marcuscollins: Same comment code rabbit came up with, maybe just `startswith("-")`

src/sampleworks/runs/schema.py
```diff
@@ -87,6 +116,9 @@ class Preset:
     shared_args : dict of str to Any, optional
         Args merged into every job's ``args`` before argv is built. Per-job
         ``args`` win on collision.
+    pre_jobs : list of Job
```
marcuscollins, line 119: Will the pre-jobs always run? Can we cache results? I can imagine a scenario (because it has happened) where we might want to modify an analysis, but we don't actually need to re-patch CIF files.

analyses/external_tools.toml
```diff
@@ -0,0 +1,45 @@
...
+
+[shared_args]
+grid-search-results-path = "${GRID_SEARCH_RESULTS_DIR}"
+grid-search-inputs-path = "${GRID_SEARCH_INPUTS_DIR}"
+protein-configs-csv = "${PROTEIN_CONFIGS_CSV}"
+target-filename = "${TARGET_FILENAME}"
+occupancies = [0.0, 0.25, 0.5, 0.75, 1.0]
+n-jobs = "${N_JOBS}"
+
+[[pre_jobs]]
```
marcuscollins, line 24: I mention this below in runner.py, but it may be better to run this script separately, or use cached results, since we might want to modify analysis scripts and rerun them, without needing to re-run patching, which is unfortunately a little slow (it can be trivially parallelized, but still takes some time)

analyses/analyze_grid_search.toml
```diff
@@ -0,0 +1,53 @@
...
+
+[shared_args]
+grid-search-results-path = "${GRID_SEARCH_RESULTS_DIR}"
+grid-search-inputs-path = "${GRID_SEARCH_INPUTS_DIR}"
+protein-configs-csv = "${PROTEIN_CONFIGS_CSV}"
+target-filename = "${TARGET_FILENAME}"
+occupancies = [0.0, 0.25, 0.5, 0.75, 1.0]
+n-jobs = "${N_JOBS}"
+
+[[pre_jobs]]
```
marcuscollins, line 24: Ditto here, best if we can cache the results or optionally skip this step.

analyses/grid_search.toml
```diff
@@ -0,0 +1,53 @@
...
+
+[defaults]
+GRID_SEARCH_RESULTS_DIR = "/data/results/grid_search_results"
+GRID_SEARCH_INPUTS_DIR = "/data/inputs"
+PROTEIN_CONFIGS_CSV = "${GRID_SEARCH_INPUTS_DIR}/protein_analysis_config.csv"
+TARGET_FILENAME = "refined-patched.cif"
+N_JOBS = "16"
+PATCH_CIF_PATTERN = "refined.cif"
+PATCH_DEPTH = "4"
```
marcuscollins: all of the scripts should need this depth argument... I think. It's a maximum directory depth to recurse into looking for CIF files. If they don't need this argument, that's on me and I'll fix it separately.

src/sampleworks/runs/loader.py
marcuscollins, line 230: Are the candidates not all unique?

src/sampleworks/runs/runner.py
```diff
@@ -466,6 +556,67 @@ def _grid_search_script() -> str:
...
+    -------
+    list of pathlib.Path
+        Candidate directories in precedence order, deduplicated after
+        expansion. The synced ACTL checkout and current working directory are
+        preferred over stale files under the baked image.
+    """
+    candidates: list[Path] = []
+    for env_var in ("SAMPLEWORKS_SOURCE_DIR", "SAMPLEWORKS_SCRIPT_ROOT"):
+        override = os.environ.get(env_var)
+        if override:
```
marcuscollins: There's a neat new-ish piece of Python that shortens these kinds of statements: https://peps.python.org/pep-0572/. You can now use:

```
if override := os.environ.get(env_var):
    candidates.append(Path(override))
```

marcuscollins: In fact, you can even use:
```
candidates = [Path(override) for env_var in (...) if (override := os.environ.get(env_var) is not None)]
```

src/sampleworks/runs/runner.py
```diff
@@ -466,6 +556,67 @@ def _grid_search_script() -> str:
...
+        return str(expanded)
+
+    for root in _source_root_candidates():
+        candidate = root / expanded
+        if candidate.exists():
+            return str(candidate)
+    return str(expanded)
+
+
+def _source_root_candidates() -> list[Path]:
```
marcuscollins, line 589: It feels odd that we need this, like it is a sign of a deeper structural problem that we don't know where the scripts are and need to look for them. Thoughts? Is there a better way to do this?

tests/runs/test_runner.py
```diff
@@ -267,6 +268,136 @@ def test_dry_run_does_not_create_directories(
...
+    assert runner.run(preset, results_dir=tmp_path / "results") == 0
+    assert (tmp_path / "main").read_text() == "main"
+
+
+def test_analysis_cli_lists_analysis_presets(capsys: pytest.CaptureFixture[str]) -> None:
+    """The analysis entrypoint lists analyses/*.toml instead of experiments/*.toml."""
+    assert analysis_cli.main(["--list"]) == 0
+    listed = set(capsys.readouterr().out.splitlines())
+    assert {
+        "all",
```
marcuscollins, line 384: I guess this will have to be updated if we add additional .toml files/presets?

Dockerfile
```diff
@@ -111,8 +112,17 @@ RUN if [ ! -e /dev/nvidiactl ] && [ ! -e /proc/driver/nvidia/version ]; then \
...
         pixi run -e boltz python -c "\
 from sampleworks.core.forward_models.xray.real_space_density_deps.ops import dilate_atom_centric; \
-print('CUDA extensions compiled successfully')"; \
-    fi
+print('CUDA extensions compiled successfully')" || echo "CUDA extension pre-compilation skipped (no GPU during build)"
+
+# This image carries pixi environments and checkpoints. Runtime source should
+# come from ACTL's synced checkout at /home/dev/workspace, not from stale code
+# baked into /app during image construction.
+RUN rm -rf /app/src /app/scripts /app/experiments /app/analyses \
```
marcuscollins, line 121: Do we need to rebuilt the current images for us to be able to update code on the fly in actl?

marcuscollins (approved):
I will leave it to you to decide what to do with my comments. I don't see anything blocking.

analyses/altloc_classify.toml
manzuoni-astera: Agreed. I removed the grid-search-specific defaults from the altloc presets in `fc9e822`: they now use `ALTLOC_ANALYSIS_DIR` and `ALTLOC_INPUTS_DIR`. `classify_altloc_regions.py` only needs the selections CSV, a CIF root, output path, and thresholds, so yes, it runs without grid-search results as long as those inputs are supplied.

analyses/altloc_classify.toml
manzuoni-astera: Yes — in this schema `shared_args` means common CLI arguments merged into every job's script invocation, with per-job `args` overriding them. I left the name in place for this PR because it is already shared by experiment and analysis presets, but I agree `script_args`/`common_args` would be clearer if we want to add a compatibility alias later.

analyses/altloc_find.toml
manzuoni-astera: Agreed, fixed in `fc9e822`. `altloc_find` now uses `ALTLOC_ANALYSIS_DIR` and `ALTLOC_INPUTS_DIR` instead of depending on `GRID_SEARCH_RESULTS_DIR`/`GRID_SEARCH_INPUTS_DIR`.

analyses/grid_search.toml
manzuoni-astera: Done in `fc9e822`: renamed the preset to `analyses/analyze_grid_search.toml`. I kept `grid_search` as an accepted CLI alias for compatibility, but the listed/canonical preset name is now `analyze_grid_search`.

src/sampleworks/runs/cli.py
manzuoni-astera: Done for the new runner CLI imports in `fc9e822` — switched them to absolute `sampleworks.runs...` imports.

src/sampleworks/runs/schema.py
manzuoni-astera: Done in `fc9e822`: changed validation to `startswith("-")` and added a regression test for `-output-dir`.

src/sampleworks/runs/schema.py
manzuoni-astera: They used to always run for non-dry-run executions. I added `--skip-pre-jobs` in `fc9e822` so reruns can reuse already-patched CIFs. I chose an explicit skip over implicit caching for now because cheap/stable invalidation across the full results tree is a little subtle; we can add timestamp/content-based cache logic later if this becomes frequent.

analyses/external_tools.toml
manzuoni-astera: Agreed. `fc9e822` adds `--skip-pre-jobs`, so `run_analysis external_tools --skip-pre-jobs ...` reruns the tool analyses without redoing patching. If this workflow keeps coming up, splitting patching into its own preset is a reasonable next step too.

analyses/analyze_grid_search.toml
manzuoni-astera: Agreed, same fix here: `--skip-pre-jobs` lets `analyze_grid_search` rerun the selected analyses while reusing existing patched CIFs.

analyses/grid_search.toml
manzuoni-astera: Added in `fc9e822`: `parse_eval_args()` now exposes `--depth`, `setup_evaluation_parameters()` passes it to `scan_grid_search_results()`, and the analysis presets pass `GRID_SEARCH_DEPTH` to all eval jobs. `PATCH_DEPTH` defaults to the same value so the patch step and eval scan stay aligned.

src/sampleworks/runs/loader.py
manzuoni-astera: Not necessarily. The raw candidate list can hit the same directory via env overrides, `/home/dev/workspace`, cwd parents, or the installed module path, and symlinks can converge after resolution. The de-dupe after `resolve(strict=False)` preserves precedence while avoiding duplicate preset discovery.

src/sampleworks/runs/runner.py
manzuoni-astera: Good call — applied the simple walrus form in `fc9e822`. I kept the explicit loop rather than a compact list comprehension because it reads cleaner here and avoids the precedence trap around `os.environ.get(env_var) is not None`.

src/sampleworks/runs/runner.py
manzuoni-astera: I agree it's a bit of a smell. The reason is ACTL: source code is synced to `/home/dev/workspace`, while the baked image/package can live under `/app`, and we want script edits to take effect without rebuilding the image. The current search prefers explicit env vars and the synced checkout over baked paths. Longer term, I'd like the wrapper to always set `SAMPLEWORKS_SOURCE_DIR` and have the runner require that for relative scripts, with package-resource fallback only for packaged/local use.

tests/runs/test_runner.py
manzuoni-astera: For adding presets, no: this is an `issubset` check, so extra `.toml` files won't break it. It only needs an update when we intentionally rename/remove one of the core bundled presets, which happened here with `grid_search` -> `analyze_grid_search` and is reflected in `fc9e822`.

Dockerfile
manzuoni-astera: For code-only changes, no rebuild should be needed: ACTL syncs the checkout to `/home/dev/workspace`, and the runner resolves scripts from that live checkout first. Rebuilds are still needed for dependency/pixi-lock changes or wrapper/image changes baked into `/usr/local/bin`; `RUNTIME_PIXI=1` is the escape hatch when intentionally testing a dependency change inside an existing pod.

PR #254, opened 2026-06-04
chore(docker): install ACTL shell and debug tools

manzuoni-astera:
## Summary
- add the shared ACTL shell/network/debug tooling to the Astera Sampleworks overlay image
- install curl, wget, htop, tmux, ncdu, ping, dig, tini, git, and base shell/cert packages alongside the existing editors/zsh
- expand the build-time command check to fail fast if any expected tool is missing

## Validation
- git diff --check

## Related
- Catalog/image rollup: Astera-org/asteractl#74

### DIFF
--- a/Dockerfile.astera
+++ b/Dockerfile.astera
@@ -34,12 +34,14 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
         nano \
         rsync \
         vim \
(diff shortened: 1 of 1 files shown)


marcuscollins (approved):
LGTM

PR #256, opened 2026-06-12
Mdc/issue 231 dont delete structure file

### LINKED ISSUE #231: bug: sampleworks-guidance deletes the --structure input file
I was running some supplementary things and files started getting removed from our occ_sweep directory. Luckily we have multiple backups of these!

`get_reward_function_and_structure` unconditionally `unlink()`s the user's `--structure` CIF after parsing. https://github.com/diff-use/sampleworks/blob/963f70e47f171c34890226fb1ff871eb799d70f2/src/sampleworks/utils/guidance_script_utils.py#L237-L238

`structure_path` comes from argparse as `str` but `safe_structure_path` is `Path`. `Path("/x") != "/x"` is always `True`, so on the no-rewrite branch of `resolve_mixed_hetatm_atom_altlocs` (where it returns the input path unchanged) the cleanup deletes the caller's real input instead of a sampleworks tempfile. Deletes any CIF that doesn't trigger the mixed HETATM/ATOM altloc fix. Introduced in 15de1e9.

I think we can fix by coercing both sides to `Path`.

```python
input_structure_path = Path(structure_path)
safe_structure_path = resolve_mixed_hetatm_atom_altlocs(input_structure_path)
...
if safe_structure_path != input_structure_path:
   safe_structure_path.unlink()
```

marcuscollins:
This makes a small change to make sure we do not delete original input structure files, and adds a test to ensure this behavior remains correct. This scenario arose because we started checking for unusual CIF files that include different sequences at altloc positions (when there is compositional heterogeneity in the crystal structure, in our case because some fraction of the protein in the crystal is chemically modified). If there were no changes needed, we returned the original file path. We wanted to delete any temporarily created files, but checked that the file was temporary by testing that the file was not the original, but only tested that the file name object was different than the original--these could be Path and str objects which were logically the same but not strictly the same type, and therefore we could accidentally delete the original file by unlinking it. 

Resolves #231 

## Summary by CodeRabbit

* **New Features**
  * Added macOS ARM (osx-arm64) support alongside existing Linux environments

* **Bug Fixes**
  * Prevented accidental deletion of original structure files when invoking reward-generation with path strings

* **Chores**
  * Reorganized dependency and feature configuration for clearer environment management
  * Moved toolchain packages into platform-specific declarations and expanded workspace platforms
  * Added environment wiring for the new macOS feature

### DIFF
--- a/pixi.lock
+++ b/pixi.lock
@@ -1,6 +1,7 @@
 version: 7
 platforms:
 - name: linux-64
+- name: osx-arm64
 environments:
   analysis:
     channels:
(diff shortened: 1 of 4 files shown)


k-chrispens (approved):
Looks good to me! Thanks for the fix.

PR #263, opened 2026-06-16
feat(eval): canonicalize mixed-altloc modified residues

k-chrispens:
Adds a processing utility that is necessary for getting certain proteins (6NI6/6NI5) to work properly in the evaluation scripts, since those contain compositional heterogeneity in the form of a modified amino acid.

## Summary by CodeRabbit

* **New Features**
  * Added canonicalization for mixed alternate-location (altloc) residue names by mapping modified amino acids (e.g., CSO→CYS, MSE→MET) to canonical forms to improve altloc consistency.
  * Altloc reference loading now applies this canonicalization automatically.

* **Bug Fixes**
  * Improves stacking reliability by ensuring mixed-altloc residues converge to a shared compatible atom set and by clearing `hetero` for canonicalized residues.
  * When convergence isn’t possible, the residue is left unchanged and a warning is emitted.

* **Tests**
  * Added regression and unit tests covering canonicalization, atom dropping, warnings, and non-mutation of inputs.

### DIFF
--- a/src/sampleworks/eval/structure_utils.py
+++ b/src/sampleworks/eval/structure_utils.py
@@ -16,7 +16,7 @@ from loguru import logger
 from sampleworks.core.rewards.protocol import RewardInputs
 from sampleworks.eval.eval_dataclasses import ProteinConfig
 from sampleworks.models.protocol import GenerativeModelInput
-from sampleworks.utils.atom_array_utils import load_structure_with_altlocs
+from sampleworks.utils.atom_array_utils import BLANK_ALTLOC_IDS, load_structure_with_altlocs
 from sampleworks.utils.atom_reconciler import AtomReconciler
 from sampleworks.utils.framework_utils import match_batch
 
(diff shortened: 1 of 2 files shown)


src/sampleworks/eval/structure_utils.py
```diff
@@ -427,6 +430,75 @@ def get_asym_unit_from_structure(
...
+        Canonical three letter amino acid name, or ``None`` if HETATM.
+    """
+    one_letter = get_1_from_3_letter_code(
+        res_name, ChainType.POLYPEPTIDE_L, use_closest_canonical=True
+    )
+    parent = get_3_from_1_letter_code(one_letter, ChainType.POLYPEPTIDE_L)
+    return None if parent == UNKNOWN_AA else parent
+
+
+def canonicalize_mixed_altloc_residues(
```
marcuscollins, line 477: Does this solve the same problem as `resolve_mixed_hetatm_atom_altlocs`? If so, IIRC AtomWorks inserts an extra residue, and it looks to me like this might not catch that. 

Also, if this does solve that same problem, I'd rather we have one solution, not two which could diverge. I'm not saying my solution was better (it is a bit of a hack, tbh, to have to write a temporary CIF file). But before we put this in, we should unify the two (again, if they are the same)

src/sampleworks/eval/structure_utils.py
```diff
@@ -427,6 +430,75 @@ def get_asym_unit_from_structure(
...
+    for position, res_name in zip(positions, res_names):
+        res_names_by_position[position].add(res_name)
+
+    parent_cache: dict[str, str | None] = {}
+    for i, (position, res_name) in enumerate(zip(positions, res_names)):
+        if len(res_names_by_position[position]) == 1:
+            continue  # single consistent res_name means it doesn't have comp het
+        parent = parent_cache.setdefault(res_name, _closest_canonical_amino_acid(res_name))
+        if parent is not None and parent != res_name:
+            out.res_name[i] = parent
```
marcuscollins, line 556: What happens if the two residues don't have the same atoms? I see a couple issues here. One is that AtomWorks' parse will insert extra atoms when there's compositional heterogeneity. So, for example in 6NI6, you would end up with _two_ CA, C, O, etc... as well as extra atoms that aren't compatible with the residue type anymore. I think probably we need to handle that and canonicalize the atoms, not just the residue names.

src/sampleworks/eval/structure_utils.py
```diff
@@ -427,6 +430,75 @@ def get_asym_unit_from_structure(
...
+    res_names_by_position: dict[tuple, set[str]] = defaultdict(set)
+    for position, res_name in zip(positions, res_names):
+        res_names_by_position[position].add(res_name)
+
+    parent_cache: dict[str, str | None] = {}
+    for i, (position, res_name) in enumerate(zip(positions, res_names)):
+        if len(res_names_by_position[position]) == 1:
+            continue  # single consistent res_name means it doesn't have comp het
+        parent = parent_cache.setdefault(res_name, _closest_canonical_amino_acid(res_name))
+        if parent is not None and parent != res_name:
```
marcuscollins: Are you trying to canonicalize all residues, or just make the sequences the same? What if a structure has two different non-canonicals at the same position, and no canonicals?

src/sampleworks/eval/structure_utils.py
```diff
@@ -443,7 +515,12 @@ def get_reference_atomarraystack(
     ref_path = protein_config.get_reference_structure_path(altloc_occupancies)
     if ref_path is None:
         return None, None
-    ref_struct = load_structure_with_altlocs(ref_path)
+
+    # A modified residue (e.g. CSO) recorded as an altloc at the same position as its canonical
+    # form (e.g. CYS) makes map_altlocs_to_stack fail: the conformers disagree on res_name/hetero
+    # so biotite.stack() rejects them. Canonicalize these residues to get map_altlocs_to_stack to
+    # work
+    ref_struct = canonicalize_mixed_altloc_residues(load_structure_with_altlocs(ref_path))
```
marcuscollins, line 611: I think looking at your code, you should try to replace my method with yours although I'm pretty sure you need to get rid of extra atoms still.

marcuscollins: If you don't have time, make an issue and tag with engineering.

marcuscollins (changes requested):
I like this and I think it should ultimately replace the more involved method I previously wrote to handle compositional heterogeneity. But I think there are a couple of problems, mainly that I think this will leave extra atoms (most of which are duplicates of each other and _probably_ have an altloc id that is different, but some will be atoms that are incompatible with the new residue type).

src/sampleworks/eval/structure_utils.py
```diff
@@ -427,6 +430,75 @@ def get_asym_unit_from_structure(
...
+        res_names_by_position[position].add(res_name)
+
+    parent_cache: dict[str, str | None] = {}
+    for i, (position, res_name) in enumerate(zip(positions, res_names)):
+        if len(res_names_by_position[position]) == 1:
+            continue  # single consistent res_name means it doesn't have comp het
+        parent = parent_cache.setdefault(res_name, _closest_canonical_amino_acid(res_name))
+        if parent is not None and parent != res_name:
+            out.res_name[i] = parent
+            out.hetero[i] = False
```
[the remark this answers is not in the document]
k-chrispens: Canonical residues will not have hetatm=True

manzuoni-astera (changes requested):
The previous blocker remains at current head `4b8cc2a`; the rebase left both changed-file blobs identical to the reviewed version.

1. Add an end-to-end 6NI6/6NI5 regression that runs `get_reference_atomarraystack()` through `map_altlocs_to_stack()` and verifies conformers stack cleanly, atom identities are unique, and CSO-only atoms such as `OD` are removed. The current CA-only unit fixture does not exercise the production failure.
2. Only canonicalize a mixed position when all recognized residue names converge to one canonical parent. The current partial rewrite can still leave mismatched residue names that fail during stacking.
3. Unify this behavior with `resolve_mixed_hetatm_atom_altlocs()`, or document why evaluation and guidance intentionally require different compositional-heterogeneity policies and link a concrete follow-up.

Please rebase onto current `main` and rerun required checks after addressing these points.

src/sampleworks/eval/structure_utils.py
```diff
@@ -448,6 +451,146 @@ def get_asym_unit_from_structure(
...
+    mixed_positions = {pos for pos, names in res_names_by_position.items() if len(names) > 1}
+
+    # 1. Only canonicalize positions whose altlocs all resolve to the same canonical parent. A
+    #    partial rename that still leaves mismatched res_names at the position wouldn't be
+    #    stackable anyway, so it's better to leave those untouched and warn.
+    parent_cache: dict[str, str | None] = {}
+    canonical_parent_by_position: dict[tuple, str] = {}
+    for position in mixed_positions:
+        parents = {
+            parent_cache.setdefault(name, _closest_canonical_amino_acid(name))
```
marcuscollins: I imagine that `_closest_canonical_amino_acid` is not that slow? Maybe parent_cache isn't needed?

marcuscollins (approved):
LGTM thanks. Maybe file that issue to try to unify our two canonicalization paths?

PR #265, opened 2026-06-18
Ext as default shell

denis-zaitsev:
## Summary by CodeRabbit

* **New Features**
  * Enabled eBPF-based process tracing in the Astera extension configuration.
  * Automatically initializes the extension shell in interactive sessions and installs shell hooks during image setup.
* **Documentation**
  * Updated Docker/CI documentation to remove the deprecated `EXT_CLI_IMAGE` override.
* **Chores**
  * Removed reliance on `EXT_CLI_IMAGE` and refreshed how the Astera image installs the extension.
  * Added stricter installation validation and behavior during setup.
  * Updated `.gitignore` to ignore `.idea/`.

### DIFF
--- a/docker/astera/install-ext.sh
+++ b/docker/astera/install-ext.sh
@@ -1,7 +1,10 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
-curl -fsSL https://extshell.org/install.sh | bash
+EXT_VERSION="${EXT_VERSION:-v0.1.1}"
+EXT_INSTALL_DIR="${EXT_INSTALL_DIR:-/usr/local/bin}"
+
+curl -fsSL https://extshell.org/install.sh | bash -s -- --version "${EXT_VERSION}" --dir "${EXT_INSTALL_DIR}"
 install -d -m 0755 /home/dev/.local/share/ext
 install -m 0644 /usr/local/share/sampleworks/astera/ext-config.toml /home/dev/.local/share/ext/config.toml
 command -v ext >/dev/null 2>&1
--- a/docker/astera/install-ext.sh
+++ b/docker/astera/install-ext.sh
@@ -1,0 +1,8 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+curl -fsSL https://extshell.org/install.sh | bash
+install -d -m 0755 /home/dev/.local/share/ext
+install -m 0644 /usr/local/share/sampleworks/astera/ext-config.toml /home/dev/.local/share/ext/config.toml
+command -v ext >/dev/null 2>&1
+ext --help >/dev/null
--- a/docker/astera/sampleworks-ext-shell.sh
+++ b/docker/astera/sampleworks-ext-shell.sh
@@ -1,6 +1,6 @@
-# Start ext only for real interactive terminal sessions. Non-interactive
-# commands such as `actl pod exec -- ...`, CI, and VS Code server probes must
-# keep their original process shape.
+# Enter ext only when ACTL sets EXT_SHELL=1.
+[ "${EXT_SHELL:-}" = "1" ] || return 0 2>/dev/null || exit 0
+
 case "$-" in
   *i*) ;;
   *) return 0 2>/dev/null || exit 0 ;;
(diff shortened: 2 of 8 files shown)


docker/astera/install-ext.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+curl -fsSL https://extshell.org/install.sh | bash
```
Abdelsalam-Abbas: thread an ARG EXT_VERSION in Dockerfile.astera and pass --version "$EXT_VERSION" (or set EXT_VERSION) in install-ext.sh, so builds are pinned and reproducible. The installer already supports exactly this, so it's a one-line change.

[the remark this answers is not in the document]
marcuscollins: What is the reason for getting this from an https endpoint? I assume we (Astera) own that URL, but can we have some verification?

docker/astera/install-ext.sh
```diff
@@ -0,0 +1,8 @@
...
+install -d -m 0755 /home/dev/.local/share/ext
+install -m 0644 /usr/local/share/sampleworks/astera/ext-config.toml /home/dev/.local/share/ext/config.toml
+command -v ext >/dev/null 2>&1
```
marcuscollins, line 10: What does this do?

docker/astera/sampleworks-ext-shell.sh
```diff
@@ -0,0 +1,40 @@
+# Start ext only for real interactive terminal sessions. Non-interactive
+# commands such as `actl pod exec -- ...`, CI, and VS Code server probes must
```
marcuscollins: Might create an issue to explore whether we can also track remote commands executed with something like `actl pod exec`, it would be useful for "production" runs.

docker/astera/sampleworks-ext-shell.sh
```diff
@@ -0,0 +1,40 @@
...
+# keep their original process shape.
+case "$-" in
+  *i*) ;;
+  *) return 0 2>/dev/null || exit 0 ;;
+esac
+
+[ -t 0 ] || return 0 2>/dev/null || exit 0
+[ -t 1 ] || return 0 2>/dev/null || exit 0
+[ -z "${SAMPLEWORKS_DISABLE_EXT_SHELL:-}" ] || return 0 2>/dev/null || exit 0
```
marcuscollins: These seem a little brittle to maintain. Can you at least link to where we might find these variables in case they ever change name?

docker/astera/sampleworks-ext-shell.sh
```diff
@@ -0,0 +1,40 @@
+# Start ext only for real interactive terminal sessions. Non-interactive
+# commands such as `actl pod exec -- ...`, CI, and VS Code server probes must
+# keep their original process shape.
+case "$-" in
+  *i*) ;;
+  *) return 0 2>/dev/null || exit 0 ;;
+esac
+
+[ -t 0 ] || return 0 2>/dev/null || exit 0
```
marcuscollins, line 9: Are these two tests redundant with the `case "$-"` statement above? I had to look these up but I guess the -t tests are just checking whether the session is interactive?

docker/astera/sampleworks-ext-shell.sh
```diff
@@ -0,0 +1,40 @@
...
+[ -z "${ZSH_EXECUTION_STRING:-}" ] || return 0 2>/dev/null || exit 0
+command -v ext >/dev/null 2>&1 || return 0 2>/dev/null || exit 0
+
+__sampleworks_ext_config_template="/usr/local/share/sampleworks/astera/ext-config.toml"
+__sampleworks_ext_data_home="${XDG_DATA_HOME:-${HOME:-/home/dev}/.local/share}"
+__sampleworks_ext_config_dir="${__sampleworks_ext_data_home}/ext"
+__sampleworks_ext_config="${__sampleworks_ext_config_dir}/config.toml"
+mkdir -p "${__sampleworks_ext_config_dir}" 2>/dev/null || true
+if [ ! -e "${__sampleworks_ext_config}" ] && [ -r "${__sampleworks_ext_config_template}" ]; then
+  cp "${__sampleworks_ext_config_template}" "${__sampleworks_ext_config}" 2>/dev/null || true
```
marcuscollins, line 23: what's the purpose of the ` || true`?

marcuscollins (changes requested):
I have one or two real issues, and some other questions, then can approve.

docker/astera/install-ext.sh
denis-zaitsev: Checks that ext is in path after install.
Failing the build if it is not

docker/astera/sampleworks-ext-shell.sh
denis-zaitsev: Good point. Probably we can just eliminate it until someone finally needs that type of behaviour.
Removed.

docker/astera/sampleworks-ext-shell.sh
denis-zaitsev: We just don't won't to fail ssh to container if for some reason we wasn't able to copy pre defined config.
In this case ext will work with it's internal default configuration.

docker/astera/sampleworks-ext-shell.sh
denis-zaitsev: 1 checking whether shell is interactive
2 checks that stdin/stdout is streamed into terminal - i.e. not streamed into file for example

Abdelsalam-Abbas (commented):
LGTM

marcuscollins (approved):
Thanks for answering my questions. LGTM.

PR #267, opened 2026-06-18
fix(protpardelle): complete denoising step wiring

### LINKED ISSUE #257: Debug and complete adding Protpardelle
@marcuscollins has started integrating Protpardelle-1c into sampleworks on branch `mdc/add-protpardelle`. I can get it started on an actl pod using the following command:

```
PROTPARDELLE_MODEL_PARAMS=/mnt/diffuse-shared/marcus/model_params \
RUNTIME_PIXI=1 pixi run -e protpardelle sampleworks-guidance \
--model protpardelle --guidance-type pure_guidance --protein 1VME \
--model-checkpoint /mnt/diffuse-shared/marcus/model_params/weights/cc89_epoch415.pth \
--structure /mnt/diffuse-shared/raw/sampleworks/initial_dataset_40/processed/1VME/1VME_single_001_density_input.cif \
--density /mnt/diffuse-shared/raw/sampleworks/initial_dataset_40/processed/1VME/1VME_uniform_1.00A.ccp4 \
--resolution 1.0 --output-dir /mnt/diffuse-shared/marcus/output/ppdl/ \
--guidance-start 100 --ensemble-size 4 --augmentation --align-to-input
```

Currently this will load the model and start things running, but runs into issues, right now tensors on incompatible devices (cuda:0 and cpu). I'm putting an issue here to get help from engineering (@xraymemory) as I need to work on other things. 

After this is done and the basic setup works, we need to pipe more options through, I'll add a separate issue for that.

manzuoni-astera:
## Summary

Completes the Protpardelle debugging work from ENG-75 / #257 on top of `mdc/add-protpardelle`.

This PR intentionally targets `mdc/add-protpardelle` so the review contains only the final Protpardelle fixes, not the whole in-progress integration branch.

Key changes:

- Restores the EDM sampler/model-wrapper call shape to the protocol-compatible `step(noisy_state, t_hat, features=features)` after removing the temporary extra `eps` argument.
- Implements Protpardelle step-time noise handling via `_expand_noise_level()`, broadcasting scalar or per-batch EDM timesteps to Protpardelle's expected `B x L` tensor.
- Moves Protpardelle step inputs/conditioning tensors onto the wrapper device before the model forward pass.
- Keeps `prot_lens_per_chain` on CPU when calling Protpardelle's sampling helper to avoid CPU/GPU device mismatch in helper-created residue indices.
- Maps Atomworks selenomethionine atom name `SE` into Protpardelle atom37's methionine sulfur slot `SD`.
- Fixes `ProtpardelleConditioning` immutability so dataclass construction can complete before selected conditioning fields become frozen.
- Allows packaged Protpardelle config/data under `src/sampleworks/data/**` through `.actlignore` so ACTL sync includes the runtime YAML config.
- Re-enables the slow `step()` behavior test and adds coverage for the MSE selenium atom mapping.

## Validation

Local static checks:

- `uvx ty check src/sampleworks/models/protpardelle/wrapper.py src/sampleworks/core/samplers/edm.py tests/models/protpardelle/test_protpardelle_wrapper.py`
- `uvx ruff check src/sampleworks/models/protpardelle/wrapper.py src/sampleworks/core/samplers/edm.py tests/models/protpardelle/test_protpardelle_wrapper.py`

Remote ACTL / Protpardelle environment checks:

- `pixi run -e protpardelle-dev pytest tests/models/protpardelle -m "not slow"`
  - `27 passed, 1 deselected`
- `pixi run -e protpardelle-dev pytest tests/models/protpardelle/test_protpardelle_wrapper.py::TestStep::test_step_returns_coords`
  - `1 passed`
- Exact reduced Protpardelle guidance smoke command from #257 using the shared checkpoint and 1VME inputs
  - completed successfully with `Guidance run successfully!`
  - wrote results to `output/protpardelle-smoke-guided/`

Non-blocking warnings observed during the smoke run were expected environment warnings for unavailable optional model/tool paths and missing mirror environment variables; they did not prevent the Protpardelle run from completing.

Refs #257.

### DIFF
--- a/src/sampleworks/models/protpardelle/wrapper.py
+++ b/src/sampleworks/models/protpardelle/wrapper.py
@@ -14,10 +14,12 @@ forward pass), this is a ``StructureModelWrapper`` rather than a
 ``FlowModelWrapper``.
 """
 
+from __future__ import annotations
+
 import copy
 from dataclasses import dataclass, field
-from pathlib import Path
 from importlib.resources import files
+from pathlib import Path
 from typing import Any
 
 import biotite.structure as struc
@@ -45,6 +47,10 @@ ATOM37_NUM_ATOMS = residue_constants.atom_type_num
 # Number of aatype tokens for the 21-token alphabet (20 canonical + X).
 # Matches ``data.n_aatype_tokens`` in the Protpardelle-1c configs (e.g. cc89.yaml).
 NUM_AATYPE_TOKENS = 21
+
+# Atomworks keeps selenomethionine atoms as ``SE`` while Protpardelle's atom37
+# representation uses the canonical methionine sulfur slot, ``SD``.
+ATOM37_ATOM_NAME_ALIASES = {"SE": "SD"}
 
 # get default model configurations from protpardelle-1c/cc89
 sampling_partial_diffusion_all_atom = files("protpardelle").joinpath(
@@ -103,23 +109,28 @@ class ProtpardelleConditioning:
     residue_index: Float[Tensor, "batch L"]
     chain_index: Float[Tensor, "batch L"]
     atom_mask: Float[Tensor, "batch L 37"]
-    atom37_residue_index: Int[Tensor, "atoms"]
-    atom37_atom_index: Int[Tensor, "atoms"]
+    atom37_residue_index: Tensor
+    atom37_atom_index: Tensor
     sampling_kwargs: dict[str, Any]
     sequences: tuple[str, ...]
     x_self_conditioning: Float[Tensor, "batch L 37"] | None = None
+    _initialized: bool = field(default=False, init=False, repr=False)
 
     _FROZEN = frozenset(
         {"aatype", "seq_mask", "residue_index", "chain_index",
          "atom_mask", "sampling_kwargs", "sequences"}
     )
 
+    def __post_init__(self) -> None:
+        """Mark construction complete so selected conditioning fields become immutable."""
+        object.__setattr__(self, "_initialized", True)
+
     def __setattr__(self, key, value):
-        if key in self._FROZEN:
+        if key in self._FROZEN and getattr(self, "_initialized", False):
             raise AttributeError(
                 f"Cannot set attribute {key!r} on {self.__class__.__name__}, it is frozen!"
             )
-        super().__setattr__(key, value)
+        object.__setattr__(self, key, value)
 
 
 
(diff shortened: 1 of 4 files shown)


src/sampleworks/models/protpardelle/wrapper.py
```diff
@@ -103,23 +109,28 @@ class ProtpardelleConditioning:
     residue_index: Float[Tensor, "batch L"]
     chain_index: Float[Tensor, "batch L"]
     atom_mask: Float[Tensor, "batch L 37"]
-    atom37_residue_index: Int[Tensor, "atoms"]
-    atom37_atom_index: Int[Tensor, "atoms"]
+    atom37_residue_index: Tensor
```
marcuscollins, line 112: why get rid of the dimension annotations?

src/sampleworks/models/protpardelle/wrapper.py
```diff
@@ -103,23 +109,28 @@ class ProtpardelleConditioning:
...
+    _initialized: bool = field(default=False, init=False, repr=False)
 
     _FROZEN = frozenset(
         {"aatype", "seq_mask", "residue_index", "chain_index",
          "atom_mask", "sampling_kwargs", "sequences"}
     )
 
+    def __post_init__(self) -> None:
+        """Mark construction complete so selected conditioning fields become immutable."""
+        object.__setattr__(self, "_initialized", True)
```
marcuscollins, line 126: This is new to me... why not use `setattr` directly?

src/sampleworks/models/protpardelle/wrapper.py
```diff
@@ -103,23 +109,28 @@ class ProtpardelleConditioning:
...
+
     def __setattr__(self, key, value):
-        if key in self._FROZEN:
+        if key in self._FROZEN and getattr(self, "_initialized", False):
             raise AttributeError(
                 f"Cannot set attribute {key!r} on {self.__class__.__name__}, it is frozen!"
             )
-        super().__setattr__(key, value)
+        object.__setattr__(self, key, value)
```
marcuscollins, line 133: ditto.

src/sampleworks/models/protpardelle/wrapper.py
```diff
@@ -608,12 +665,20 @@ def step(
...
+        x_t = x_t.to(device=self.device)
 
         # Our x_t is B x N x 3 (N = number of atoms); Protpardelle expects
         # B x L x 37 x 3. Scatter into the atom37 layout (gradient-preserving).
         x_t_atom37 = self._convert_to_atom37(cond, x_t)
 
-        noise_level = torch.full((cond.seq_mask.shape[0],), sigma_float, device=self.device)
+        noise_level = self._expand_noise_level(t, cond, x_t_atom37.dtype)
+
+        seq_mask = cond.seq_mask.to(device=self.device, dtype=x_t_atom37.dtype)
```
marcuscollins, line 676: We should perhaps make a method of the `type(cond)` class to move everything to the specified device.

tests/models/protpardelle/test_protpardelle_wrapper.py
```diff
@@ -230,6 +230,17 @@ def test_atom37_index_maps_present(self, protpardelle_wrapper):
         assert int(cond.atom37_atom_index.max()) < 37
         assert int(cond.atom37_residue_index.max()) == len(SEQ_A) - 1
 
+    def test_mse_selenium_maps_to_methionine_sd(self, protpardelle_wrapper):
```
marcuscollins, line 233: Just FYI I'm mulling over whether we should more broadly canonicalize the amino acids. Karson has a PR related to this when we have compositional heterogeneity where there are different amino acids in different altlocs, e.g. the protein crystal was grown with a mix of two different proteins that are nearly identical.

marcuscollins (commented):
I left a few comments, I will leave it up to you whether/how to address them, but otherwise looks great, so approving.

Q's:
1. Why use `object.__setattr__` rather than the builtin `setattr`? 
2. Is it possible to keep more of the dimension annotations, e.g., you change `Int[Tensor, "atoms"]` to just `Tensor`?

PR #272, opened 2026-06-27
feat(rewards): add reciprocal-space reward (Fprotein from SFC)

DorisMai:
## What changed
- Added `StructureFactorRewardFunction` (`core/rewards/structure_factor.py`) via SFC. Two-phase construction (`__init__` config + `prepare(atom_array)`) as SFC requires knowing topology.
  - ensembles combine as a complex sum.: fits default scaled `|Ftotal| = |Fprotein| + |Fsolvent|`. The solvent part can have different modes: off (Fprotein only), combined (one mask for averaged protein), or one mask per conformation. This requires SFC to be the latest version that fixes the solvent scaling issue. 
  - restricted by SFC to occupancy and B-factor shared across batch. For the solvent mask (per-conformation mode), occupancy is effectively required to be exactly 1/E, which is the production result now.
  - exposes `normalize_amplitude` for testing and `batch_partition` in case of OOM
- Synthetic data generation fixes:
  - Patched `eval/generate_synthetic_sf.py` to generate test data with both Fprotein and Ftotal in the same mtz for debugging/development . 
  - Fixed issues for the round trip of cif --> atomarray --> gemmi --> cif --> atomarray in `eval/synthetic_utils.py`. Tests added in `tests/eval/test_generate_synthetic_sf.py`.
- Refactored and added reward tests. 
  - Extracted common functions from `test_real_space_density_reward.py` to `test_reward_function_contract.py`. Real space specific tests that are unused (e.g. vmap related) remain untouched.
  - Added `test_structure_factor_reward.py` and synthetic test data (1vme cif and mtz) generated by:
     ```
     pixi run -e <env> python -m sampleworks.eval.generate_synthetic_sf \
         --structure tests/resources/1vme/1vme_final.cif \
         --selection "chain A" \
         --occupancy-mode custom --occupancy-values 0.5:0.5 \
         --remove-hydrogens --remove-waters \
         --resolution 1.8 \
         --simulate-solvent-and-scale \
         --save-structure \
         --output tests/resources/1vme/1vme_final_crystalframe_0.5occA_0.5occB_1.80A.mtz
     ```

## Next steps
- Wire the reward into a guidance step scaler for initial 40 experiment
- Add scaling optimization

## Summary of changes

* **New Features**
  * Added reciprocal-space structure-factor reward evaluation (F/E) with configurable bulk-solvent handling (`off`, `combined`, `per_conformer`).
  * Enhanced synthetic MTZ generation to write multiple labeled outputs (including protein-only vs protein+total) in a single file.

* **Bug Fixes**
  * Improved MTZ amplitude/sigma column detection with safer overrides.
  * Enforced consistency of occupancy and B-factors across the batch for structure-factor scoring.
  * Improved AtomArray → gemmi conversion for more reliable altloc/annotation preservation.

* **Tests**
  * Expanded synthetic MTZ, mmCIF/altloc round-trip, and structure-factor reward coverage (including vmap/contract checks).


tests/rewards/test_reward_function_contract.py
```diff
@@ -0,0 +1,300 @@
...
+
+from tests.rewards.reward_input_helpers import build_scattering_indices
+
+
+# Every test exercises CUDA-targeted reward code on the `device` fixture (try_gpu), so the
+# whole module is gpu-marked. Deliberately NOT `slow`: measured warm per-test time is <2.5s
+# (slowest is the SFC gradient-descent loop at ~2.4s; the rest are sub-second). The ~11s of
+# fixed cost is one-time import + session-scoped reward construction, which is paid once per
+# pytest invocation and cannot be skipped by `slow`-marking these tests.
+pytestmark = pytest.mark.gpu
```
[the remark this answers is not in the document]
DorisMai: There exist tests that are already marked as gpu but now slow (e.g. `test_lddt_metrics.py`, `test_atom_reconciler.py`). This suite of tests is gpu-and-fast. Using the following command shows that except for the one-time setup that takes a few seconds, all other calls take less than a second:
`pixi run -e boltz-dev pytest tests/rewards/test_reward_function_contract.py --durations=0 --durations-min=0 -vv`

src/sampleworks/core/rewards/structure_factor.py
```diff
@@ -0,0 +1,351 @@
...
+    amplitude_cols = ds.select_mtzdtype(rs.StructureFactorAmplitudeDtype()).columns
+    sigma_cols = ds.select_mtzdtype(rs.StandardDeviationDtype()).columns
+    if len(amplitude_cols) == 0 or len(sigma_cols) == 0:
+        raise ValueError(
+            f"MTZ '{mtzfile}' needs a structure-factor-amplitude column and a "
+            f"standard-deviation column; found amplitudes={list(amplitude_cols)}, "
+            f"sigmas={list(sigma_cols)}."
+        )
+    spacegroup = ds.spacegroup.hm if ds.spacegroup is not None else None
+    return ds.cell, spacegroup, [amplitude_cols[0], sigma_cols[0]]
```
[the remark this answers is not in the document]
DorisMai: Raise error is too strict as MTZ could contain multiple (and not necessarily paired columns). I rework instead to 1) raise error if user supplied column names not found, 2) when auto-detecting (no user supplied names), log warning when multiple columns exists, pick the first amplitude, and 3) attempt to find matching sigma column name, default back to first sigma column with warning if not found.

src/sampleworks/core/rewards/structure_factor.py
```diff
@@ -0,0 +1,513 @@
...
+        """
+        assert self.sfc is not None  # prepare() built it; __call__ guards before dispatching here
+        self.sfc.Fprotein_HKL = Fprotein_HKL  # drives calc_ftotal on the HKL set
+        if self.bulk_solvent == "per_conformer":
+            # calc_fsolvent_batch masks each conformer (from Fprotein_asu_batch, set by
+            # calc_fprotein_batch); the mean applies the 1/E weight -> <mask(rho)>.
+            Fmask_HKL_batch = self.sfc.calc_fsolvent_batch(
+                Return=True, PARTITION=self.batch_partition
+            )
+            self.sfc.Fmask_HKL = Fmask_HKL_batch.mean(dim=0)
```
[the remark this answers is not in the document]
DorisMai: In the guidance pipeline `RewardInputs.from_atom_array` overrides occupancy to a uniform 1/ensemble_size, so occ ≡ 1/E and `mean(dim=0)` is exactly the 1/E population average that matches the protein sum's weighting — the occ ≠ 1/batch_size state isn't reachable from production.

src/sampleworks/eval/synthetic_utils.py
```diff
@@ -194,3 +196,106 @@ def load_structure_for_synthetic_reward(
...
+        if key != prev_key:
+            if prev_key is not None:
+                assert current_res is not None  # prev_key set implies a residue exists
+                current_chain.add_residue(current_res)
+            current_res = gemmi.Residue()
+            current_res.name = str(atom_array.res_name[i])
+            current_res.seqid = gemmi.SeqId(str(res_id))  # author numbering
+            current_res.label_seq = res_id  # label numbering -> label_seq_id on write
+            current_res.subchain = chain_id  # label_asym_id == chain name
+            prev_key = key
```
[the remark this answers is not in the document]
DorisMai: Currently the entire repo does not use ins_code,  and key on (chain_id, res_id) or (chain_id, res_id, atom_name). I am not sure when or how often we encounter ins_code. I did notice that Vratin uses (chain_id, res_id, ins_code) in WaterFlow. @marcuscollins Do you think this is worth opening an issue to fix repo-wide or low priority / assume no ins code handling? either way I don't think it should block this PR.

src/sampleworks/eval/generate_synthetic_sf.py
```diff
@@ -15,20 +17,18 @@
 from typing import Any, ClassVar
 
 import gemmi
-import numpy as np
 import reciprocalspaceship as rs
 import reciprocalspaceship.utils
 import torch
-from biotite.structure import AtomArray
 from loguru import logger
 from sampleworks.eval.synthetic_utils import (
```
marcuscollins, line 24: Please make an issue to move `synthetic_utils` out of `sampleworks.eval` (and probably the present file too, since we're clearly using these in other places besides eval.

src/sampleworks/eval/synthetic_utils.py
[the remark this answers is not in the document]
marcuscollins: @coderabbit go ahead and create the issue. Please indicate in the issue that we need to research when ins_code is used and why. That should inform our decision on what to do.

src/sampleworks/eval/synthetic_utils.py
```diff
@@ -194,3 +196,106 @@ def load_structure_for_synthetic_reward(
...
+    -----
+    Built directly rather than via ``SFC_Torch.io.array2hier``: that helper packs atom
+    identity into a hyphen-delimited ``cra_name`` string, sets only the author seqid, and
+    keys residue boundaries on ``res_id`` alone (merging residues that share a res_id across
+    a chain boundary). Building the hierarchy here from the AtomArray annotations avoids the
+    string round-trip and lets us set the author seqid, ``label_seq`` (so ``label_seq_id``
+    survives a write -> atomworks reload; otherwise res_id collapses to -1), and ``subchain``
+    (a <=1-char ``label_asym_id`` that SFcalculator's PDB-header step accepts) at construction.
+    """
+    n = len(atom_array)
```
marcuscollins, line 239: It looks like this is only used once. In that case I'd favor just using `for i in range(len(atom_array)):`

src/sampleworks/eval/synthetic_utils.py
```diff
@@ -194,3 +196,106 @@ def load_structure_for_synthetic_reward(
...
+        atom_altloc = ["\x00" if a in BLANK_ALTLOC_IDS else a for a in atom_array.altloc_id]
+    else:
+        atom_altloc = ["\x00"] * n
+
+    model = gemmi.Model("1")  # numeric name -> valid mmCIF pdbx_PDB_model_num
+    current_chain: gemmi.Chain | None = None
+    current_res: gemmi.Residue | None = None
+    prev_key: tuple[str, int] | None = None  # (chain_id, res_id) of the previous atom
+
+    for i in range(n):
```
marcuscollins, line 252: Add a comment describing what this for loop does.

src/sampleworks/eval/synthetic_utils.py
```diff
@@ -194,3 +196,106 @@ def load_structure_for_synthetic_reward(
...
+        chain_id = str(atom_array.chain_id[i])
+        res_id = int(atom_array.res_id[i])
+        key = (chain_id, res_id)
+
+        if current_chain is None or chain_id != current_chain.name:
```
marcuscollins, line 257: Add a comment indicating this is the functionality needed on encountering a new or first chain.

src/sampleworks/eval/synthetic_utils.py
```diff
@@ -194,3 +196,106 @@ def load_structure_for_synthetic_reward(
...
+                assert current_res is not None  # prev_key set implies a residue exists
+                current_chain.add_residue(current_res)
+            current_res = gemmi.Residue()
+            current_res.name = str(atom_array.res_name[i])
+            current_res.seqid = gemmi.SeqId(str(res_id))  # author numbering
+            current_res.label_seq = res_id  # label numbering -> label_seq_id on write
+            current_res.subchain = chain_id  # label_asym_id == chain name
+            prev_key = key
+
+        assert current_res is not None  # created on the first atom / every chain switch
```
marcuscollins, line 278: I would use `raise` instead. `assert` is less idiomatic except in a rarely used case where assertion statements are ignored when running pre-compiled bytecode. I forget the details of it because I don't think I've _ever_ used this feature of Python. It theoretically allows for error checking that is optimized out in the final bytecode.

src/sampleworks/eval/generate_synthetic_sf.py
```diff
@@ -1,9 +1,11 @@
 """Generate synthetic structure factor amplitudes via SFcalculator-torch.
 
-Produces an MTZ file of |Fmodel| (or |Fprotein| if not simulate solvent and
-scale) for each input PDB/mmCIF structure. The MTZ file has dummy values for
-SIGFP and optionally R-free flag column. Each structure can be optionally
-overridden with unit cell, space group, atom selection, and occupancy.
+For each input PDB/mmCIF structure, produces an MTZ file of protein structure factors
+only, or when ``--simulate-solvent-and-scale`` is set, both the protein and total sets
+(Fprotein/SIGFprotein/PHIFprotein and Ftotal/SIGFtotal/PHIFtotal) in the same
```
marcuscollins, line 5: I've been confused in the past about what the standard names of these are. Are there other common conventions? I feel like there were a couple. Should we address that?

src/sampleworks/eval/generate_synthetic_sf.py
```diff
@@ -127,62 +127,59 @@ def from_dict(cls, row: dict[str, Any]) -> "BatchRowForMTZ":
         )
 
 
-def atomarray_to_gemmi(
-    atom_array: AtomArray,
-    unit_cell: gemmi.UnitCell | None = None,
-    space_group: str | None = None,
-) -> gemmi.Structure:
-    """Convert a biotite AtomArray to a gemmi.Structure for SFcalculator.
+def _amplitude_phase_columns(
```
marcuscollins, line 130: maybe rename `_extract_amplitude_and_phase_columns`?

src/sampleworks/eval/generate_synthetic_sf.py
```diff
@@ -127,62 +127,59 @@ def from_dict(cls, row: dict[str, Any]) -> "BatchRowForMTZ":
...
-        atom_pos=atom_array.coord,
-        atom_b_aniso=np.zeros((n, 3, 3), dtype=np.float64),
-        atom_b_iso=atom_array.b_factor,
-        atom_occ=atom_array.occupancy,
-        atom_name=atom_array.element,
-        cra_name=cra_names,
-        atom_altloc=atom_altloc,
-        res_id=atom_array.res_id,
+    dataset: rs.DataSet = sfc.prepare_dataset(miller_index_column, structure_factor_column)
+    amplitude_column = dataset.select_mtzdtype(rs.StructureFactorAmplitudeDtype()).columns[0]
```
marcuscollins, line 165: If there's more than one amplitude column, how do you know which one you need?

src/sampleworks/eval/generate_synthetic_sf.py
```diff
@@ -127,62 +127,59 @@ def from_dict(cls, row: dict[str, Any]) -> "BatchRowForMTZ":
...
+    phase_column = dataset.select_mtzdtype(rs.PhaseDtype()).columns[0]
```
marcuscollins, line 166: Ditto.

src/sampleworks/eval/generate_synthetic_sf.py
```diff
@@ -193,14 +190,20 @@ def process_amplitudes_to_dataset(
     ----------
     sfc: SFcalculator
         SFcalculator instance
+    structure_factor_columns: dict[str, str]
+        Mapping of ``label -> SFcalculator attribute``. One structure-factor set
```
marcuscollins, line 194: a little confusing since it is a mapping from string to string; I guess SFcalculator attribute is also a string? If so, what are/how do we know the available SFcalculator attributes?

src/sampleworks/eval/generate_synthetic_sf.py
```diff
@@ -214,18 +217,18 @@ def process_amplitudes_to_dataset(
...
-    dataset[sigma_f_column] = dataset[sigma_f_column].astype(rs.StandardDeviationDtype())
+    if not structure_factor_columns:
+        raise ValueError("structure_factor_columns must contain at least one entry.")
+    column_items = iter(structure_factor_columns.items())
+    label, attribute = next(column_items)
+    dataset = _amplitude_phase_columns(sfc, label, attribute, miller_index_column, sigma_f_scale)
+    for label, attribute in column_items:
+        ds = _amplitude_phase_columns(sfc, label, attribute, miller_index_column, sigma_f_scale)
+        for col in ds.columns:
+            dataset[col] = ds[col]
```
marcuscollins, line 231: This is a straaaange construction. Much clearer:
```
dataset = rs.DataSet(...)  # fill in whatever is needed, maybe nothing.
for label, attribute in structure_factor_columns.items():
    ds = _amplitude_phase_columns(sfc, label, attribute, miller_index_column, sigma_f_scale)
    dataset = dataset.merge(ds, ...)  # you'll have to specify pandas style merge columns, probably hkl?
```
rs.DataSet is just a pd.DataFrame with some extra attributes, so I think this will be much cleaner.

src/sampleworks/eval/generate_synthetic_sf.py
```diff
@@ -354,15 +358,17 @@ def _process_single_row(
...
-            F_attribute = "Ftotal_asu"
-        else:
-            F_attribute = "Fprotein_asu"
-        logger.debug(f"Computed {F_attribute} for {row.filename} on {device}")
+            structure_factor_columns.update({"total": "Ftotal_asu"})
+        logger.debug(
+            f"Computed {'Fprotein + Ftotal' if simulate_solvent_and_scale else 'Fprotein'} "
+            f"for {row.filename} on {device}"
+        )
     except Exception as e:
```
marcuscollins, line 372: Please file an issue to work through this file and make the `except` statements more specific.

src/sampleworks/eval/generate_synthetic_sf.py
```diff
@@ -552,7 +558,11 @@ def parse_args() -> argparse.Namespace:
     sf_group.add_argument(
         "--simulate-solvent-and-scale",
         action="store_true",
-        help="Compute bulk solvent and overall scale factors (outputs Ftotal instead of Fprotein)",
+        help=(
+            "Compute bulk solvent and overall scale factors and write both protein and "
+            "total structure factor in one MTZ. Without this flag, protein only. Each "
+            "set contains F{label}/SIGF{label}/PHIF{label}."
```
marcuscollins, line 564: say what `label` is.

tests/conftest.py
```diff
@@ -712,6 +717,59 @@ def test_coordinates_1vme(structure_1vme_density, device: torch.device):
...
+
+@pytest.fixture(scope="session")
+def test_coordinates_1vme_sf(structure_1vme_sf, device: torch.device):
+    # No occupancy>0 filter here (unlike the density fixture) as SFC topology is fixed.
+    atom_array = structure_1vme_sf
+    coords = torch.from_numpy(atom_array.coord).to(device=device, dtype=torch.float32)
+    return coords, atom_array
+
+
+@pytest.fixture(scope="session")
```
marcuscollins, line 755: probably could put this in the specific test file and scope to module.

src/sampleworks/core/rewards/structure_factor.py
```diff
@@ -0,0 +1,513 @@
...
+    )
+    return [amplitude, sigma]
+
+
+class StructureFactorRewardFunction:
+    """Reward for fitting structure-factor amplitudes via SFcalculator.
+
+    The reward compares the model amplitudes ``|Fcalc|`` against a target
+    ``|Fobs|`` loaded from an MTZ. For an ensemble (batch dimension), the
+    members are combined as a single multi-conformer model in reciprocal space:
```
marcuscollins, line 178: Do we sum over the different conformers separately? That is, do we consider separate crystals of each conformer and average them? I.e., what is `e`?

src/sampleworks/core/rewards/structure_factor.py
```diff
@@ -0,0 +1,513 @@
...
+        f"No expcolumns provided; auto-detected SFC columns: "
+        f"amplitude='{amplitude}', sigma='{sigma}'."
+    )
+    return [amplitude, sigma]
+
+
+class StructureFactorRewardFunction:
+    """Reward for fitting structure-factor amplitudes via SFcalculator.
+
+    The reward compares the model amplitudes ``|Fcalc|`` against a target
```
marcuscollins, line 176: This docstring probably belongs in __init__

src/sampleworks/core/rewards/structure_factor.py
```diff
@@ -0,0 +1,513 @@
...
+        ``(unit_cell, space_group_hm, amplitude_columns, sigma_columns)``, where the last
+        two are *all* structure-factor-amplitude and standard-deviation columns found.
+        Pairing/selecting among them is left to :func:`_resolve_expcolumns`.
+
+    Raises
+    ------
+    ValueError
+        If the MTZ has no amplitude column or no sigma column.
+    """
+    import reciprocalspaceship as rs
```
marcuscollins, line 98: is there a reason for this not to be a module-level import?

src/sampleworks/core/rewards/structure_factor.py
```diff
@@ -0,0 +1,513 @@
...
+
+    ds = rs.read_mtz(mtzfile)
```
marcuscollins, line 100: this is presumably slow, is there a way to just pass in the ds object itself, so the file doesn't have to be loaded and then most of the data thrown away?

src/sampleworks/core/rewards/structure_factor.py
```diff
@@ -0,0 +1,513 @@
...
+    """
+    if expcolumns is not None:
+        valid = set(amplitude_cols) | set(sigma_cols)
+        unknown = [c for c in expcolumns if c not in valid]
+        if unknown:
+            raise ValueError(
+                f"Provided expcolumns {list(expcolumns)} include {unknown}, not found as MTZ "
+                f"amplitude/sigma columns (amplitudes={amplitude_cols}, sigmas={sigma_cols})."
+            )
+        return list(expcolumns)
```
marcuscollins, line 150: isn't it already a list?

src/sampleworks/core/rewards/structure_factor.py
```diff
@@ -0,0 +1,513 @@
...
+
+    amplitude = amplitude_cols[0]
```
marcuscollins, line 152: There's some similar logic around L160 in generate_synthetic_sf.py. This raises a concern that maybe we're making this choice in multiple places and they need to stay aligned. Perhaps this method needs to go in synthetic_utils.py and then used here and in generate_synthetic_sf.py.

src/sampleworks/core/rewards/structure_factor.py
```diff
@@ -0,0 +1,513 @@
...
+    if len(amplitude_cols) > 1:
+        logger.warning(
```
marcuscollins, line 154: Should we instead require the user to specify the columns to use if there is more than one?

src/sampleworks/core/rewards/structure_factor.py
```diff
@@ -0,0 +1,513 @@
...
+        self.loss: AmplitudeLoss = loss if loss is not None else torch.nn.MSELoss()
+        self.normalize_amplitude = normalize_amplitude  # |F| vs resolution-bin normalized |E|
+        self.sfcalculator_kwargs = dict(sfcalculator_kwargs) if sfcalculator_kwargs else {}
+
+        # Resolve crystal metadata / column names against the MTZ.
+        self._resolve_mtz_metadata(unit_cell, space_group, expcolumns)
+
+        # All SFcalculator init kwargs are known except `pdbmodel` (needs the model
+        # atom array), which is injected in prepare().
+        self._sfc_kwargs: dict = dict(
```
marcuscollins, line 280: the extra `dict` annotation is unnecessary.

src/sampleworks/core/rewards/structure_factor.py
```diff
@@ -0,0 +1,513 @@
...
+                f"bulk_solvent must be one of {_BULK_SOLVENT_MODES}, got {bulk_solvent!r}."
+            )
+        self.bulk_solvent = bulk_solvent
+        self.exclude_free_reflections = exclude_free_reflections
+        if batch_partition <= 0:
+            raise ValueError(f"batch_partition must be a positive integer, got {batch_partition}.")
+        self.batch_partition = batch_partition
+        self.loss: AmplitudeLoss = loss if loss is not None else torch.nn.MSELoss()
+        self.normalize_amplitude = normalize_amplitude  # |F| vs resolution-bin normalized |E|
+        self.sfcalculator_kwargs = dict(sfcalculator_kwargs) if sfcalculator_kwargs else {}
```
marcuscollins, line 273: isn't `sfcalculator_kwargs` already a dict if it isn't None?

src/sampleworks/core/rewards/structure_factor.py
```diff
@@ -0,0 +1,513 @@
...
+        self._sfc_kwargs: dict = dict(
+            mtzdata=self.mtzfile,
+            dmin=self.resolution,
+            mode=self.scattering_factor_mode,
+            anomalous=False,
+            set_experiment=True,
+            expcolumns=self.expcolumns,
+            device=self.device,
+        )
+        self._sfc_kwargs.update(self.sfcalculator_kwargs)
```
marcuscollins, line 289: It feels odd to keep two versions of SFCalc kwargs. I think that could lead to confusion. Is there a reason to keep them separate?

src/sampleworks/core/rewards/structure_factor.py
```diff
@@ -0,0 +1,513 @@
...
+        real values.)
+        """
+        cell, spacegroup, amplitude_cols, sigma_cols = _detect_mtz_metadata(self.mtzfile)
+
+        if unit_cell is not None and not unit_cell.is_similar(
+            cell, _CELL_LENGTH_REL_TOL, _CELL_ANGLE_DEG_TOL
+        ):
+            logger.warning(
+                f"Provided unit_cell {unit_cell.parameters} differs from the MTZ's "
+                f"{cell.parameters}; using the provided value."
```
marcuscollins, line 317: when would we do this? It seems very dangerous to me.

src/sampleworks/core/rewards/structure_factor.py
```diff
@@ -0,0 +1,513 @@
...
+            )
+        self.unit_cell = unit_cell if unit_cell is not None else cell
+
+        if space_group is not None and space_group != spacegroup:
+            logger.warning(
```
marcuscollins, line 322: Ditto, seems possibly dangerous.

src/sampleworks/core/rewards/structure_factor.py
```diff
@@ -0,0 +1,513 @@
...
+        """Set ``self.unit_cell`` / ``space_group`` / ``expcolumns``, resolved against the MTZ.
+
+        The MTZ is the source of truth (it must be consistent with its own
+        reflections), so it is always read. For any argument the caller supplied
+        (non-``None``), that value is used but a warning is logged on disagreement, so a
+        stale or mismatched override is visible. (SFcalculator reads the cell/space group
+        from the gemmi structure built in :meth:`prepare`, not the MTZ, so we must supply
+        real values.)
+        """
+        cell, spacegroup, amplitude_cols, sigma_cols = _detect_mtz_metadata(self.mtzfile)
```
marcuscollins, line 310: Append something to these names so that it is clear where they come from, e.g., `spacegroup_mtz`. Otherwise you end up with expressions like `space_group == spacegroup` which is both hard to read and error prone.

src/sampleworks/core/rewards/structure_factor.py
```diff
@@ -0,0 +1,513 @@
...
+        if space_group is not None and space_group != spacegroup:
+            logger.warning(
+                f"Provided space_group {space_group!r} differs from the MTZ's "
+                f"{spacegroup!r}; using the provided value."
+            )
+        self.space_group = space_group if space_group is not None else spacegroup
+
+        self.expcolumns = _resolve_expcolumns(expcolumns, amplitude_cols, sigma_cols)
+
+    def prepare(self, atom_array: AtomArray) -> None:
```
marcuscollins, line 330: will this work for AtomArrayStack?

src/sampleworks/core/rewards/structure_factor.py
```diff
@@ -0,0 +1,513 @@
...
+        )
+        self.sfc = SFcalculator(pdbmodel=PDBParser(gemmi_structure), **self._sfc_kwargs)
+        # inspect_data estimates solvent percentage and grid size from atom positions
+        # and vdW radii, independent of occupancy / B-factor.
+        self.sfc.inspect_data()
+
+        # Ftotal uses the default (unrefined) scales kiso=1, kmask=0.35, small uaniso,
+        # matching generate_synthetic_sf's Ftotal. They don't depend on coordinates, so
+        # set them once here. _set_scales only needs atom_pos_frac (set at construction)
+        # and n_bins (set by the experiment init) for dtype/device and bin count.
```
marcuscollins, line 363: what is the "experiment init"?

src/sampleworks/core/rewards/structure_factor.py
```diff
@@ -0,0 +1,513 @@
...
+        if self.bulk_solvent != "off":
+            self.sfc._set_scales(requires_grad=False)
+
+        # |Eo| are computed in SFC's experiment init (inside a try/except).
```
marcuscollins, line 367: remind me, what is Eo?

src/sampleworks/core/rewards/structure_factor.py
```diff
@@ -0,0 +1,513 @@
...
+            raise RuntimeError(
+                "normalize_amplitude=True requires sfc.Eo, but "
+                "SFcalculator did not populate them from this MTZ."
+            )
+
+        # Reflection mask: drop outliers, and optionally the free (test) set, when
+        # computing the loss. Outlier / free_flag are numpy bool arrays from SFC.
+        mask_np = ~self.sfc.Outlier
+        if self.exclude_free_reflections:
+            mask_np = mask_np & ~self.sfc.free_flag
```
marcuscollins, line 378: Nit: I actually prefer `x &= y` to `x = x & y`.

src/sampleworks/core/rewards/structure_factor.py
```diff
@@ -0,0 +1,513 @@
...
+            annotations (its ``b_factor``/``occupancy`` are baked as defaults but
+            overridden each ``__call__``). A missing ``altloc_id`` is defaulted to
+            blank inside ``atomarray_to_gemmi``.
+        """
+        gemmi_structure = atomarray_to_gemmi(
+            atom_array,
+            unit_cell=self.unit_cell,
+            space_group=self.space_group,
+        )
+        self.sfc = SFcalculator(pdbmodel=PDBParser(gemmi_structure), **self._sfc_kwargs)
```
marcuscollins, line 355: so does `self.sfc` contain the actual amplitudes and sigmas? I don't see anywhere here that explicitly stores the data in the mtz file. I would call out somewhere where that data lives to make the code easier to understand.

src/sampleworks/core/rewards/structure_factor.py
```diff
@@ -0,0 +1,513 @@
...
+                    "batch dim (SFcalculator has no per-conformer occupancy/B axis); got "
+                    "per-conformer values."
+                )
+        self.sfc.atom_b_iso = b_factors[0]
+        self.sfc.atom_occ = occupancies[0]
+
+        # Multi-conformer combination: complex sum over the ensemble [batch, n_hkl].
+        # occ = 1/E (set per call) makes the summed |F| the multi-conformer total.
+        Fprotein_batch = self.sfc.calc_fprotein_batch(
+            coordinates, Return=True, PARTITION=self.batch_partition
```
marcuscollins, line 448: omg we need to talk to Minhuan about his Python style. capital letters kwarg names!?

marcuscollins (commented):
Lots of comments to look at. The biggest concern I have is about picking the first type-matching column for amplitudes, etc...

I'm also confused about how some of the changes in generate_synthetic_sf.py relate to the changes in structure_factor.py. They appear disconnected, but maybe I'm missing something. Can you explain?

src/sampleworks/core/rewards/structure_factor.py
DorisMai: `e` is `ensemble`, so Fprotein = average(Fprotein_e). Things get a bit trickier when Fsolvent is considered (the "combined" vs "per-conformer" mode)

src/sampleworks/core/rewards/structure_factor.py
DorisMai: I am not sure metadata like cell and spacegroup are always enforced to be written to mtz (can be default to dummy values like P1 and "1 1 1 90 90 90")

src/sampleworks/core/rewards/structure_factor.py
DorisMai: SFC's internal method (SFcalculator.init_mtz()) called in its own `__init__`

src/sampleworks/core/rewards/structure_factor.py
DorisMai: observed amplitudes normalized in each resolution bin

src/sampleworks/eval/generate_synthetic_sf.py
DorisMai: Each refinement program has there own names used, so I don't think there is a standard (which is why auto-detect by column type). Usually F* is for amplitude, SIGF* is for amplitude uncertainty, PHIF* is for phase, and that's the convention I am following here.

src/sampleworks/eval/generate_synthetic_sf.py
DorisMai: for this case (a dataset produced by SFcalculator), it's not that complicated -- there should be only 1 amplitude and 1 phase columns. I auto-select first column here to avoid using the undocumented hardcoded str names in SFC ("Fmodel") which was a concern you raised in previous pr.

src/sampleworks/eval/generate_synthetic_sf.py
DorisMai: SFC sets the attributes as they become populated/available rather than declare all of them at init, which is why this dict looks confusing. For practicality, we really just need to look at Fprotein_asu and/or Ftotal_asu that we care about for now. Let me know if you think this needs better handling and how.

src/sampleworks/eval/generate_synthetic_sf.py
DorisMai: this is just a generic safe guard to mirror the generate_synthetic_density file (line 205-209)

DorisMai:
> Lots of comments to look at. The biggest concern I have is about picking the first type-matching column for amplitudes, etc...
> 
> I'm also confused about how some of the changes in generate_synthetic_sf.py relate to the changes in structure_factor.py. They appear disconnected, but maybe I'm missing something. Can you explain?

My previous PR implemented atomarray_to_gemmi() that correctly feed into SFC, and optionally saved a cif file for my visual inspection. However, that cif file doesn't correctly round trip back to atomarray. I noticed this issue/needed to correct it in order to generate the tests for the structure factor reward function, which happens in the middle of this PR, hence the diffs look a bit messy.

src/sampleworks/core/rewards/structure_factor.py
DorisMai: fixed

src/sampleworks/core/rewards/structure_factor.py
DorisMai: fixed

src/sampleworks/core/rewards/structure_factor.py
DorisMai: fixed

src/sampleworks/core/rewards/structure_factor.py
DorisMai: fixed

src/sampleworks/core/rewards/structure_factor.py
DorisMai: fixed

src/sampleworks/core/rewards/structure_factor.py
DorisMai: fixed

src/sampleworks/core/rewards/structure_factor.py
DorisMai: fixed

src/sampleworks/core/rewards/structure_factor.py
DorisMai: fixed

src/sampleworks/eval/generate_synthetic_sf.py
DorisMai: added

src/sampleworks/eval/generate_synthetic_sf.py
DorisMai: renamed as `_build_rs_dataset_for_one_label`

src/sampleworks/eval/generate_synthetic_sf.py
DorisMai: The index is identical, so when there are multiple labels, we just need column concat rather than a merge/join. I can get rid of the iter/next and do directly column assignment/concatenation if this reads better:
```
dataset: rs.DataSet | None = None
for label, attribute in structure_factor_columns.items():
    ds = _build_rs_dataset_for_one_label(sfc, label, attribute, miller_index_column, sigma_f_scale)
    if dataset is None:
        dataset = ds                 # first set carries cell / spacegroup / dtypes
    else:
        dataset[ds.columns] = ds     # graft the remaining columns on (index-aligned)
```

src/sampleworks/core/rewards/structure_factor.py
DorisMai: updated the docstring of `prepare` to describe the `sfc` object better

src/sampleworks/eval/synthetic_utils.py
DorisMai: done

src/sampleworks/eval/synthetic_utils.py
DorisMai: sounds good, fixed

src/sampleworks/eval/synthetic_utils.py
DorisMai: added

src/sampleworks/eval/synthetic_utils.py
DorisMai: added

src/sampleworks/eval/generate_synthetic_sf.py
DorisMai: opened

src/sampleworks/core/rewards/structure_factor.py
DorisMai: good advice, will be refactored

src/sampleworks/core/rewards/structure_factor.py
DorisMai: refactored

src/sampleworks/core/rewards/structure_factor.py
DorisMai: this is indeed a safer approach

manzuoni-astera (changes requested):
Requesting changes at current head `c3f689f`.

1. MTZ column resolution must require exactly an amplitude and matching sigma column with correct dtypes, and require explicit selection when multiple amplitude sets exist. The current union check permits malformed/reversed input while auto-selection makes the scored target depend on file order.
2. Caller-provided unit-cell and space-group corrections are not honored because SFC 0.3.3 overwrites parser metadata from the MTZ during initialization. Correct the dataset metadata before SFC initialization or remove/reject ineffective overrides.
3. `exclude_free_reflections=True` does not recognize MTZs generated here without manually overriding the free-flag column/value, and silently excludes nothing.
4. The per-conformer solvent path requires occupancy exactly `1 / ensemble_size`, but only broadcast equality is checked, producing inconsistent protein/solvent scaling for other values.

Please also rebase to resolve the current merge conflict and run the waiting GPU jobs. Several discussion replies describe refactors that are not present in the current head.

tests/conftest.py
DorisMai: moved to tests/reward/conftest.py (new file). scope is still session because it is consumed by four test modules (test_reward_inputs.py, test_reward_function_contract.py, test_real_space_density_reward.py, and test_structure_factor_reward.py)

src/sampleworks/core/rewards/structure_factor.py
DorisMai: no, but it shouldn't need to. this is intended to accept `model_atom_array`, and the purpose of this function is get topology info, which should be available from one AtomArray.

PR #274, opened 2026-06-29
feat: add protpardelle

marcuscollins:
This is a working version of Protpardelle-1c in SampleWorks. It may still require parameter tuning and other updates--in particular it isn't clear how to use self-conditioning. However we can generate structures with this version and so I'm making this PR as a milestone. 

Only a couple significant changes have been made outside the Protpardelle wrapper code itself. One is to make a features class that is unfrozen, so that we can pass self-conditioning input forward to the next Euler step during sampling. Another is that we define different sampling parameters for Protpardelle when instantiating the sampler. There is no CLI control for either change in this PR.

## Summary by CodeRabbit

* **New Features**
  * Added Protpardelle structure guidance support, including a new model option, Protpardelle-specific CLI parameters, and packaged sampling configuration.
* **Bug Fixes**
  * Guidance now logs the parsed configuration at startup.
* **Chores**
  * Updated development ignore rules for sample data and refined dependency/feature pins for Protpardelle-related installs.
* **Tests**
  * Added Protpardelle wrapper and CLI argument parsing tests; adjusted runner tests for more deterministic behavior.


src/sampleworks/utils/imports.py
```diff
@@ -172,6 +184,52 @@ def wrapper(*args: Any, **kwargs: Any) -> Any:
...
+                    import pytest
+
+                    pytest.skip(error_msg)
+                except ImportError:
+                    raise ImportError(error_msg) from None
+            return func(*args, **kwargs)
+
+        return wrapper  # type: ignore
+
+    return decorator
```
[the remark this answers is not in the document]
marcuscollins: @coderabbitai make an issue to track this as an issue, do not implement it. Let one of the humans implement it.

tests/models/protpardelle/conftest.py
```diff
@@ -0,0 +1,159 @@
...
+from pathlib import Path
+
+import pytest
+
+
+# Must be set before any `import protpardelle...` happens. Respect an
+# externally configured directory (e.g. when real weights are available).
+os.environ.setdefault(
+    "PROTPARDELLE_MODEL_PARAMS", tempfile.mkdtemp(prefix="protpardelle_model_params_")
+)
```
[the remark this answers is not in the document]
marcuscollins: @xraymemory https://github.com/diff-use/sampleworks/issues/286

tests/runs/test_runner.py
```diff
@@ -26,7 +26,8 @@ def test_argv_for_rf3_partial_matches_bash(monkeypatch: pytest.MonkeyPatch) -> N
     assert inv.log_path == Path("/results/rf3_run.log")
 
     argv = inv.argv
-    assert argv[:6] == ["pixi", "run", "-e", "rf3", "python", "/app/run_grid_search.py"]
+    assert argv[:5] == ["pixi", "run", "-e", "rf3", "python"]
+    assert "run_grid_search.py" in argv[5]  # the exact path varies by invocation
```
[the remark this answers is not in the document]
marcuscollins: @coderabbitai Please make an issue for this but do not implement it. Please mention @manzuoni-astera and @xraymemory and tag the issue "P1" and "engineering"

src/sampleworks/cli/guidance.py
```diff
@@ -3,18 +3,25 @@
...
+
+    from sampleworks.utils.guidance_script_utils import get_model_and_device, run_guidance
+
+    logger.info(f"Running guidance with config: {config}")
     device, model_wrapper = get_model_and_device(
         config.device,
         getattr(config, "model_checkpoint", None),
         config.model,
         method=getattr(config, "method", None),
+        protpardelle_config_path=getattr(config, "protpardelle_config_path", None),
```
k-chrispens: Is there a way we can remove this here, or at least make it more general rather than including a model specific arg here? I know there are a lot of models that use hydra YAML configs, so I expect we will probably need to generalize this at some point

src/sampleworks/core/samplers/edm.py
```diff
@@ -415,15 +415,17 @@ def step(
...
         noisy_state = torch.as_tensor(noisy_state).detach().requires_grad_(allow_gradients)
 
         # t_hat will be float if check_context didn't raise
         # Use no_grad when gradients aren't needed to avoid memory overhead from
         # gradient checkpointing holding intermediate activations
+        # TODO testing adding eps to signature for use with Protpardelle-1c, if successful,
+        #   I need to modify the Protocol itself. @Michael Anzuoni
         with torch.set_grad_enabled(allow_gradients):
-            x_hat_0 = model_wrapper.step(noisy_state, t_hat, features=features)
+            x_hat_0 = model_wrapper.step(noisy_state, t_hat, eps, features=features)
```
k-chrispens: I'm confused here, because this no longer appears to be needed in the model step function that is in this PR. So I think #283 should be closed.

src/sampleworks/models/protpardelle/wrapper.py
```diff
@@ -0,0 +1,766 @@
+"""Wrapper for Protpardelle-1c models.
+
+Follows the ``StructureModelWrapper`` protocol in
```
k-chrispens: This isn't correct - this shouldn't be following the StructureModelWrapper. It should be (and I believe is, though that would require an isinstance check) a FlowModelWrapper

src/sampleworks/models/protpardelle/wrapper.py
```diff
@@ -0,0 +1,766 @@
...
+``sampleworks.models.protocol`` so Protpardelle can be used interchangeably
+with other structure predictors in sampling pipelines.
+
+This wrapper targets the *sequence-conditioned* all-atom Protpardelle-1c
+models (``task: "ai-allatom"``, e.g. the model defined by ``cc89.yaml``).
+These models take a protein sequence as their only conditioning input and
+generate an all-atom structure by running the full reverse-diffusion
+trajectory internally. Because a single ``step`` call encapsulates the entire
```
k-chrispens: Also confused by this - the step call shouldn't be encapsulating the whole diffusion loop (and currently isn't)

src/sampleworks/models/protpardelle/wrapper.py
```diff
@@ -0,0 +1,766 @@
...
+        seq_mask = cond.seq_mask.to(device=self.device, dtype=x_t_atom37.dtype)
+        residue_index = cond.residue_index.to(device=self.device)
+        chain_index = cond.chain_index.to(device=self.device)
+        struct_self_cond = cond.x_self_conditioning
+        if struct_self_cond is not None:
+            struct_self_cond = struct_self_cond.to(device=self.device, dtype=x_t_atom37.dtype)
+
+        # To understand all these arguments better, you can study the (complicated)
+        # Protpardelle.sample method. Hints: partial diffusion is enabled and cc.enabled = False!
+        # https://github.com/ProteinDesignLab/protpardelle-1c/src/protpardelle/core/models.py
```
k-chrispens: This link doesn't exist. I am assuming you mean this? https://github.com/ProteinDesignLab/protpardelle-1c/blob/ee378400f25b801fa481028000f9060183d7fb4c/src/protpardelle/core/models.py#L740

k-chrispens: And https://github.com/ProteinDesignLab/protpardelle-1c/blob/ee378400f25b801fa481028000f9060183d7fb4c/src/protpardelle/core/models.py#L1766

src/sampleworks/models/protpardelle/wrapper.py
```diff
@@ -0,0 +1,766 @@
...
+            sse_cond=None,  # secondary structure conditioning, TODO maybe use?
+            adj_cond=None,  # adjacency conditioning, # TODO not sure what this is for exactly
+            seq_self_cond=None,  # we don't do sequence design.
+            seq_crop_cond=None,
+            run_mpnn_model=False,  # we don't want to do sequence design
+        )
+
+        # pass the self-conditioning to the next step by updating the features.
+        # Detach it since we don't want the gradient flowing back to a previous step.
+        # TODO: I wonder if we need to adjust this since we will apply additional guidance.
```
k-chrispens, line 718: Dru recommends not dealing with this

src/sampleworks/models/protpardelle/wrapper.py
```diff
@@ -0,0 +1,766 @@
...
+        # ai-allatom sampling, so select the correct atoms via the atom37 mask.
+        # The mask is identical across the batch (single input sequence), so a
+        # single 2-D mask flattens every ensemble member consistently.
+        atom_mask_2d = cond.atom_mask[0].to(device=x0.device).bool()  # [L, 37]
+        flat_coords = x0[:, atom_mask_2d]  # [batch, atoms, 3]
+
+        return flat_coords
+
+    # This just generates the input noise. When doing partial diffusion, the
+    # noise is added to the input coordinates with some weight..
```
k-chrispens, line 731: At some point (can't find the issue in the search here) you suggested refactoring out all these functions that are the same across diffusion models. In this case, the initialize_from_prior is basically the same across the 3 models, so perhaps we add a new issue to deal with this

k-chrispens: Though of course with protpardelle it will require using the atom37 representation...

src/sampleworks/utils/guidance_script_arguments.py
```diff
@@ -197,6 +197,15 @@ def validate_model_checkpoint(
     "msa_path",
     "disable_chiral_features",
     "track_chiral_features",
+    "protpardelle_config_path",
+    "protpardelle_s_churn",
```
k-chrispens, line 201: This s_churn parameter is related to gamma_min in the AF3EDMSampler - if we want to adjust it, we should maybe make a new EDMSampler that has all the parameters in Karras et al instead of having the specific renamings that AF3 uses.

src/sampleworks/utils/guidance_script_arguments.py
```diff
@@ -197,6 +197,15 @@ def validate_model_checkpoint(
...
+    "protpardelle_step_scale",
```
k-chrispens, line 202: also, step_scale is already in the sampler. We shouldn't use an additional config parameter for it

src/sampleworks/utils/guidance_script_arguments.py
```diff
@@ -197,6 +197,15 @@ def validate_model_checkpoint(
...
+    "protpardelle_sidechain_mode",
+    "protpardelle_skip_mpnn_proportion",
+    "protpardelle_jump_steps",
+    "protpardelle_uniform_steps",
+    "protpardelle_temperature",
+    "protpardelle_top_p",
```
k-chrispens, line 208: I think we need to be careful here. We should come up with a way to avoid adding a bunch of arguments to this file that aren't more general across models. As we add more models, it will make this file inscrutable. I imagine the first thing to do would be to maybe include an "arguments/knobs" module along with the wrapper.py and then import here?

src/sampleworks/utils/guidance_script_arguments.py
```diff
@@ -606,11 +615,77 @@ def add_rf3_specific_args(parser: argparse.ArgumentParser | GuidanceConfig):
...
+        default=True,
+        help="Use Protpardelle uniform-step sampling (default: enabled)",
+    )
+    parser.add_argument(
+        "--protpardelle-temperature",
+        type=float,
+        default=1.0,
+        help="Temperature applied to Protpardelle aatype logits",
+    )
+    parser.add_argument(
```
k-chrispens, line 675: We shouldn't need this since we will never be doing design, right?

src/sampleworks/utils/guidance_script_utils.py
```diff
@@ -482,6 +506,26 @@ def _run_guidance(args: GuidanceConfig, guidance_type: str, model_wrapper, devic
...
             recycling_steps=recycling_steps,
         )
+    elif "Protpardelle" in wrapper_class_name:
+        from sampleworks.models.protpardelle.wrapper import annotate_structure_for_protpardelle
+
+        structure = annotate_structure_for_protpardelle(
+            structure,
+            ensemble_size=args.ensemble_size,
+            num_steps=args.num_diffusion_steps,
+            s_churn=getattr(args, "protpardelle_s_churn", 40.0),
```
k-chrispens, line 533: I imagine several of these are not going to be adjusted - not that we want to restrict the number of knobs, but especially the design arguments don't seem relevant to include here. Also, several of these are not consumed anywhere aside from in the annotation on the structure.

src/sampleworks/utils/guidance_script_utils.py
```diff
@@ -692,8 +738,9 @@ def run_guidance_job_queue(job_queue_path: str) -> list[JobResult]:
     device, model_wrapper = get_model_and_device(
         str(template_job.device),
         template_job.model_checkpoint,
-        template_job.model,
+        template_job.model,  # this is not actually the model, it's the model name, e.g. boltz2
```
k-chrispens, line 762: should we just rename this to model_name?

src/sampleworks/models/protpardelle/wrapper.py
```diff
@@ -0,0 +1,766 @@
...
+        """
+        This runs just the forward pass of the Protpardelle model. See
+        protpardelle-1c/src/protpardelle/core/models.py:L1760
+        (commit ee378400f25b801fa481028000f9060183d7fb4c on branch main)
+
+        The returned tensor is the final all-atom prediction, flattened to the 
+        atoms implied by the input sequence (the ``atom_mask`` in the 
+        conditioning).
+
+        Parameters
```
k-chrispens, line 658: This docstring isn't correct

src/sampleworks/models/protpardelle/wrapper.py
```diff
@@ -0,0 +1,766 @@
...
+                f"Noise level shape {tuple(noise_level.shape)} must be scalar, "
+                f"batch-sized, or match seq_mask shape {tuple(seq_mask.shape)}."
+            )
+        return noise_level
+
+    def step(
+        self,
+        x_t: Float[Tensor, "batch atoms 3"],
+        t: Float[Tensor, "*batch"] | float,
+        sigma_float: float,
```
k-chrispens: unused parameter that should be removed

k-chrispens (changes requested):
I saw some of the recent issues as I left work today and got sucked into reading the rest of this PR. If this is ready for review, these are my thoughts. The most important thing is that it doesn't appear that the modelwrapper step function is aligned with how it is called in the sampler. It would also be good if the protpardelle wrapper got plugged into the other modelwrapper integration tests in tests/conftest.py - I think several of these tests would fail in the current implementation

src/sampleworks/cli/guidance.py
marcuscollins: Good callout. I will do the same with the "method" argument too since it is Boltz2-specific.

src/sampleworks/core/samplers/edm.py
marcuscollins: You are correct, it isn't needed (although I am a little worried that it very slightly changes the alignment transform, which ppdl does using the coordinates without eps.) This has already been changed, it was a carry-over from a merge I executed inexpertly. :)

src/sampleworks/models/protpardelle/wrapper.py
marcuscollins: You are correct, thanks, fixed.

src/sampleworks/models/protpardelle/wrapper.py
marcuscollins: Leftover from Claude's first--and as you point out, incorrect--implementation that I then had to fix.

src/sampleworks/utils/guidance_script_arguments.py
marcuscollins: I think I'm going to leave these in. We aren't doing design now, but personally I think we will want to push for designing ensembles in the future. Maybe we won't do it, but I can see lots of reasons to keep it an option.

src/sampleworks/utils/guidance_script_utils.py
marcuscollins: Yes, but I think it has a significant "blast radius", so let's make an issue for it and we can PR that separately.

@coderabbitai please make an issue to change the argument/attribute GuidanceConfig.model to GuidanceConfig.model_name, tag it "P1" and "engineering". Make sure to note in the description to thoroughly test this. There are places where `model` can refer to the model name (a string or one of StructurePredictor's enum values) and other places (e.g. in the Boltz2Wrapper) where it refers to an actual ModelWrapper object.

src/sampleworks/models/protpardelle/wrapper.py
marcuscollins: @k-chrispens what isn't correct? I've updated to try to be more precise.

src/sampleworks/models/protpardelle/wrapper.py
marcuscollins: Done.

src/sampleworks/models/protpardelle/wrapper.py
k-chrispens: At the time, it didn't match the signature, but I think Michael's PR fixed this

marcuscollins:
commit history got messed up. Closing this and re-opening on mdc/protpardelle

PR #277, opened 2026-06-29
Pass checkpoint image properly

denis-zaitsev:
Fix Docker workflow output masking.

A value stored in GitHub Secrets matches part of the Docker image path, so GitHub masks that part of the string. Because we were passing the full image path between jobs as an output, the value could be dropped/emptied.

This changes the workflow to pass only the checkpoint digest between jobs and rebuild the full image path inside the build job.

## Summary by CodeRabbit

* **Chores**
  * Improved the way checkpoint image references are passed between build steps, using digest information to construct the final image reference more reliably.
  * Added validation to ensure the checkpoint digest is present and correctly formatted before builds continue.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -34,7 +34,7 @@ jobs:
     permissions:
       contents: read
     outputs:
-      checkpoints_image: ${{ steps.verify_checkpoints.outputs.checkpoints_image }}
+      checkpoints_digest: ${{ steps.verify_checkpoints.outputs.checkpoints_digest }}
 
     steps:
       - name: Copy checkpoint image to Docker Hub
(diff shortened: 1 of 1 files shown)


Abdelsalam-Abbas (approved):
Approving. Verified against the latest `main` run: `sync_checkpoints` succeeds, `public` fails at the empty-input check, `astera` is skipped — exactly the secret-in-job-output emptying this fixes. Passing the bare `sha256:` digest across the job boundary and reconstructing the ref inside the consuming job (a step output, not subject to the same emptying) is the right mechanism, and the added `sha256:` prefix check is a nice guard.

Non-blocking follow-up: the `public` -> `astera` `image-ref` job output (docker.yml:111/179, consumed at :230) embeds the same `docker.io/<namespace>/...` path and is the same kind of cross-job output. If the masking secret is the Docker Hub namespace, `astera` will hit the identical emptying once `public` starts succeeding (it's been skipped until now). Worth applying the same pattern there — pass only the digest/short-sha and rebuild the full ref inside `astera`.

PR #280, opened 2026-06-29
ci(docker): avoid secret-masked image outputs

manzuoni-astera:
## Summary
- pass only the checkpoint digest between workflow jobs so GitHub secret masking does not suppress the `sync_checkpoints` output
- reconstruct the digest-pinned Docker Hub checkpoint ref inside the public image job before invoking buildx
- pass the public image digest to the Astera overlay job and build the overlay from that exact digest instead of a SHA tag output
- update Docker docs to describe the digest-based handoff

## Why
The Docker workflow currently mirrors and verifies checkpoints successfully, then GitHub skips the `checkpoints_image` job output because the full image ref appears to contain a secret-matched value. The public image job receives an empty `CHECKPOINTS_IMAGE` and exits before any image build starts.

This keeps cross-job outputs to bare `sha256:...` digests, then reconstructs image refs inside the consumer job where they are used.

## Validation
- `git diff --check`
- `ruby -e 'require "yaml"; YAML.load_file(".github/workflows/docker.yml")'`\n- `actionlint .github/workflows/docker.yml`

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated Docker image handling to use digest-pinned references across workflow steps, reducing the chance of mismatched or outdated images.
  * Improved validation and propagation of checkpoint image values between build jobs.
* **Documentation**
  * Clarified the Docker setup notes to reflect that the private overlay image is built from the exact public image digest generated in the same run.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -3,7 +3,7 @@ name: Build and Push Docker Images
 # Two-image split:
 #   1. Public pixi-with-checkpoints image from Dockerfile -> public registry.
 #   2. Private Astera overlay from Dockerfile.astera -> Harbor, based on the
-#      exact public image SHA tag built in step 1.
+#      exact public image digest built in step 1.
 
 on:
   push:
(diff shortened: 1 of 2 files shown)


denis-zaitsev:
This thing is already fixed in main

PR #281, opened 2026-06-30
Sampleworks container build fully fixed

denis-zaitsev:
- Use the public image digest from the Docker build output instead of rebuilding a sha-* tag manually.
- Pass that digest-pinned public image into the Astera EXT image build.
- Add a small validation step so the Astera build fails early if the public image digest is missing or invalid.
- Login to the public registry before the Astera build, because it needs to pull the public image by digest.
- Make Docker build cache upload best-effort with ignore-error=true, so cache push failures do not fail the actual image release.
- Verified that the new Harbor image was pushed and contains ext plus the ext shell hook.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved container image handling to use a verified digest for handoff between build stages, reducing the risk of pulling the wrong image version.
  * Added validation to fail builds when the expected image digest is missing or not properly pinned.
  * Made build cache uploads more resilient so cache-related issues won’t fail the workflow.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,6 +35,9 @@ concurrency:
   group: ci-${{ github.ref }}
   cancel-in-progress: true
 
+env:
+  CONDA_OVERRIDE_CUDA: "12"
+
 jobs:
   lint:
     runs-on: ubuntu-latest
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -210,8 +210,7 @@ jobs:
             type=semver,pattern={{version}}
             type=semver,pattern=v{{version}}
 
-      - name: Resolve public image input
-        id: public-ref
+      - name: Validate public image digest
         env:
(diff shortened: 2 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -35,6 +35,9 @@ concurrency:
   group: ci-${{ github.ref }}
   cancel-in-progress: true
 
+env:
+  CONDA_OVERRIDE_CUDA: "12"
```
denis-zaitsev, line 39: We added CONDA_OVERRIDE_CUDA=12 because CI runs on CPU-only GitHub runners, but our Pixi environments depend on CUDA 12 packages. Newer Pixi checks this more strictly and fails if CUDA is not detected, even before tests start. This variable tells Pixi to resolve the environment as CUDA 12 without requiring a real GPU on the CI runner.

mag-astera (approved):

PR #292, opened 2026-07-01
feat: support extended (12-char) PDB IDs in output CIF patching

smallfishabc:
RCSB now issues extended PDB IDs (`pdb_########`, e.g. `pdb_00004hhb`) alongside legacy 4-char codes, ahead of the July 2027 archive cutover. Make the CIF-patching pipeline accept both forms so it keeps working through the transition, with 4-char behavior unchanged.

- patch_output_cif_files.py: widen the RCSB-id regex to `(pdb_[A-Za-z0-9]{8}|<legacy>)` and route extraction through a new `extract_rcsb_id` helper. The helper validates the captured token against the real PDB-id grammar (legacy = leading digit; extended = `pdb_` + 8 alnum), so a stray folder like `TEST/` is rejected instead of silently fetched, and returns the id verbatim.
- analyses/{all,analyze_grid_search,external_tools}.toml: PATCH_RCSB_PATTERN accepts both forms.
- tests: add tests/eval/test_patch_output_cif_files.py (legacy, extended, and rejected inputs) and update the runner pattern assertion.

Extended IDs already resolve on RCSB production, so fetch works today with no endpoint change.

## Summary by CodeRabbit

* **New Features**
  * Expanded RCSB identifier recognition to support both legacy 4-character IDs and newer `pdb_`-prefixed 8-character formats across evaluation and CIF patching workflows.
* **Bug Fixes**
  * Improved CIF path parsing to reject ambiguous/partial matches and distinguish “no match” from structurally invalid captures.
* **Documentation**
  * Updated the patching command example to use the broadened `--rcsb-pattern` regex.
* **Tests**
  * Added/updated unit tests for extraction behavior (legacy/extended formats, no-match, invalid-capture, and regex misconfiguration), plus a test helper to load standalone scripts.


analyses/all.toml
```diff
@@ -10,7 +10,7 @@ PATCH_CIF_PATTERN = "refined.cif"
 GRID_SEARCH_DEPTH = "4"
 PATCH_DEPTH = "${GRID_SEARCH_DEPTH}"
 PATCH_INPUT_PDB_PATTERN = "processed/{pdb_id}/{pdb_id}_single_001_density_input.cif"
-PATCH_RCSB_PATTERN = "${GRID_SEARCH_RESULTS_DIR}/([A-Za-z0-9]{4})"
+PATCH_RCSB_PATTERN = "${GRID_SEARCH_RESULTS_DIR}/(pdb_[A-Za-z0-9]{8}|[A-Za-z0-9]{4})"
```
marcuscollins: As you point out in the description, we could change the original regex to `[0-9][A-Za-z0-9]{3}`.

scripts/eval/EVALUATION.md
```diff
@@ -46,7 +46,7 @@ You can run the following command, which assumes:
 ```shell
 pixi run -e analysis python scripts/patch_output_cif_files.py \
   --input-dir /home/ubuntu/grid_search_results \
-  --rcsb-pattern 'grid_search_results/(.{4})/...' \
+  --rcsb-pattern 'grid_search_results/(pdb_[A-Za-z0-9]{8}|.{4})/...' \
```
marcuscollins: Let's update this so it handles current-style 4-character PDB ids more precisely.

scripts/patch_output_cif_files.py
```diff
@@ -66,7 +66,7 @@ def parse_args():
     )
     parser.add_argument(
         "--rcsb-pattern",
-        default="grid_search_results/(.{4})",
+        default="grid_search_results/(pdb_[A-Za-z0-9]{8}|.{4})",
```
marcuscollins: ditto here, let's use the updated pattern that handles 4-character IDs better.

scripts/patch_output_cif_files.py
```diff
@@ -137,19 +137,51 @@ def main(
...
+    into the wrong template without erroring.
+    """
+    m = re.search(rcsb_regex, str(cif_path))
+    if not m:
+        return None
+    if m.re.groups != 1:
+        raise ValueError(
+            f"--rcsb-pattern must have exactly one capturing group, "
+            f"got {m.re.groups}: {rcsb_regex!r}"
+        )
```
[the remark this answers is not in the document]
marcuscollins: @smallfishabc this is a good callout from copilot, please make this change.

scripts/patch_output_cif_files.py
```diff
@@ -137,19 +137,51 @@ def main(
...
+    token = m.group(1)
+    if not _VALID_RCSB_ID.fullmatch(token):
```
marcuscollins, line 214: Is it easier to check that `_VALID_RCSB_ID` is contained in `rcsb_regex` first? Then you could raise an `ValueError` that the user-provided regex is incorrect and telling the user they need to include a valid PDB ID regex in their pattern.

scripts/patch_output_cif_files.py
```diff
@@ -137,19 +137,51 @@ def main(
...
     cif_path = Path(cif_file)
-    m = re.search(rcsb_regex, str(cif_path))
-    rcsb_id = m.group(1) if m else None
-    if not m:
+    rcsb_id = extract_rcsb_id(cif_path, rcsb_regex)
+    if rcsb_id is None:
         msg = (
-            f"Unable to parse an RCSB structure id: from path {cif_file} with pattern {rcsb_regex}"
+            f"Unable to parse a valid RCSB structure id from path {cif_file} "
+            f"with pattern {rcsb_regex}"
```
marcuscollins: See previous comment, but this leaves open the possibility that the user-provided `rcsb_regex` is incompatible with `_VALID_RCSB_ID`, so we should check that somewhere (probably in extract_rcsb_id) and tell the user if that might be the problem.

tests/eval/test_patch_output_cif_files.py
```diff
@@ -0,0 +1,84 @@
...
+    spec = importlib.util.spec_from_file_location("patch_output_cif_files_script", _SCRIPT_PATH)
+    assert spec is not None and spec.loader is not None
+    mod = importlib.util.module_from_spec(spec)
+    spec.loader.exec_module(mod)
+    return mod
+
+
+@pytest.fixture(scope="module")
+def script():
+    return _load_script()
```
marcuscollins: Why not just define the whole function here?

tests/eval/test_patch_output_cif_files.py
```diff
@@ -0,0 +1,84 @@
...
+
+from __future__ import annotations
+
+import importlib.util
+from pathlib import Path
+
+import pytest
+
+
+_SCRIPT_PATH = Path(__file__).resolve().parents[2] / "scripts" / "patch_output_cif_files.py"
```
marcuscollins: You might consider using `files` https://docs.python.org/3/library/importlib.resources.html#importlib.resources.files instead. There's nothing strictly wrong with the approach you have here, but `files` makes it a little cleaner I think.

tests/eval/test_patch_output_cif_files.py
marcuscollins: Also, we do exactly the same thing (with a different script) in `tests/eval/test_rscc_grid_search_script.py`. We should make this a utility function that can load any script, and then you can point it to the specific script you want in a specific test file like this one.

tests/eval/test_patch_output_cif_files.py
```diff
@@ -0,0 +1,84 @@
...
+
+
+@pytest.fixture(scope="module")
+def script():
+    return _load_script()
+
+
+# The script-side default pattern. The TOML presets use ``[A-Za-z0-9]{4}`` for the legacy
+# branch instead of ``.{4}``; both are exercised (see ``test_matches_toml_legacy_class``).
+_DEFAULT_REGEX = r"grid_search_results/(pdb_[A-Za-z0-9]{8}|.{4})"
```
marcuscollins: Is this defined anywhere else? I would watch out for things like this that should be constants defined in just one place.

tests/eval/test_patch_output_cif_files.py
```diff
@@ -0,0 +1,84 @@
...
+
+
+@pytest.mark.parametrize(
+    ("folder", "expected"),
+    [
+        ("4hhb", "4hhb"),  # legacy, lowercase
```
marcuscollins, line 37: Should we (or do we) handle examples like "protease_pdb_1000abcd" where we'd extract "pdb_1000abcd"? It might be good to include some more complex examples here.

marcuscollins (changes requested):

tests/eval/test_patch_output_cif_files.py
smallfishabc: I am writing a separate util file for this function and use it. I think if that looks good we can change other script as well.

scripts/patch_output_cif_files.py
smallfishabc: I think I will generate a separate warning for the regex

tests/eval/test_patch_output_cif_files.py
smallfishabc: I defined it in the patch_output_cif_files.py but in the test script it only show up once . We can incorporate this with the script loader later if it shows up multiple times.

tests/eval/test_patch_output_cif_files.py
smallfishabc: I think I will keep it as is for now.

tests/eval/test_patch_output_cif_files.py
smallfishabc: I fix the logic so it will send out warnings about protease_pdb like input.

manzuoni-astera (approved):
Reviewed current head `99e6f9e`. The prior functional blockers are resolved: capture-group validation precedes matching, IDs must occupy a complete normalized path component, invalid captures get distinct diagnostics, and `_VALID_RCSB_ID` is defined before the generated default pattern. CPU tests, type checks, lint, and CodeRabbit pass. The GPU jobs are waiting for protected-environment approval rather than failing.

Two non-blocking follow-ups: update the guidance that suggests custom patterns can extract IDs embedded inside larger folder names, because boundary validation intentionally rejects those; and add regression assertions covering `DEFAULT_RCSB_PATTERN` plus a permissive custom regex that exercises the left-boundary check.

Approving; merge after required checks complete.

marcuscollins (approved):
Michael has approved the latest version. It looks like my concerns have been addressed, thanks.

PR #295, opened 2026-07-02
refactor(msa): remove unused Protenix a3m writes in _compute_msa

### LINKED ISSUE #179: Check if output for Protenix from _compute_msa is used, remove if not.
It appears that the code fragment below from the end of `_compute_msa` in `msa.py` returns output that isn't used, based on the fact that it gets overwritten for every new call to the colabfold API, and that this doesn't break Protenix. I believe Protenix MSAs are produced separatel. We should confirm that this output is unused, and if so, remove the code and the comments above it. 

The code in question is at the very end of `_compute_msa`:
```
        # TODO I don't think these are actually used anywhere. We can probably remove them.

        msa_idx_idr = msa_dir / f"{idx}"
        msa_idx_idr.mkdir(exist_ok=True, parents=True)
        logger.info(f"Writing MSA for target {target_id} sequence {idx} to {msa_idx_idr.resolve()}")
        msa_idx_idr.joinpath("non_pairing.a3m").write_text(unpaired_msa[idx])
        if paired:
            msa_idx_idr.joinpath("pairing.a3m").write_text(paired_msas[idx])

    return outputs
```

manzuoni-astera:
## Summary

Fixes #179.

`_compute_msa` wrote per-sequence `non_pairing.a3m` / `pairing.a3m` files under `msa_dir/{idx}/`, guarded by a `# TODO I don't think these are actually used anywhere` comment. They aren't:

- The **boltz/rf3** path consumes the flat `{hash}_{n}.a3m` / `.csv` files validated by `_validate_msa_cache_contents`.
- The **Protenix** path reads `non_pairing.a3m` / `pairing.a3m` from a *different* directory — `msa_dir/protenix/<hash>/<idx>/` — produced by `protenix_msa_search`, never from the `msa_dir/{idx}/` location this block wrote.
- No test references these files (`grep tests/` for `non_pairing` / `pairing.a3m` → nothing).

Removed the dead writes plus the stale stray format-doc string that only described them. No consumers to update.

Closes #179.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved MSA handling for Protenix requests so results are cached in the expected Protenix-specific location.
  * Prevented duplicate Protenix searches for repeated identical requests by reusing cached results.
  * Ensured standard MSA outputs remain separate from Protenix-specific cached files.

### DIFF
--- a/src/sampleworks/utils/msa.py
+++ b/src/sampleworks/utils/msa.py
@@ -229,46 +229,6 @@ def _compute_msa(
 
         outputs[name] = msa_path
 
-        """
-        a3m files for Protenix.
-        In case we weren't having enough fun, we need to write it all out in yet a third
-        # format, for use by Protenix. Protenix expects a JSON blob like so:
-        [
-          {
-            "sequences": [
-              {"proteinChain":
-                {
-                  "sequence": "ACDE...",
-                  "msa":
-                    {
-                      "precomputed_msa_dir:": "/path/to/msa_directory"
-                      "pairing_db": "uniref100"
-                    },
-                  ... other fields
-                }
-              }, ...
-            ]
-          }, ...
-        ]
-
-        It expects /path/to/msa_directory to look like this:
-        0:  # index corresponds to position in the list "sequences"
-            non_pairing.a3m
-            pairing.a3m  # only present if pairing was used.
-        1:
-            non_pairing.a3m
-            pairing.a3m
-        etc...
-        """
-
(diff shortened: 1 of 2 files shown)


marcuscollins (commented):
I'm still a little nervous that this will backfire--can we do some functional testing (or maybe add unit tests with good mocking for HTTPS endpoints) to make sure that we will properly populate and use the cache for these protenix MSA files especially?

marcuscollins (approved):
Thanks for the quick update.

PR #296, opened 2026-07-02
docs(models): standardize model_atom_array docstring across wrappers

### LINKED ISSUE #109: Clarify naming of `model_atom_array` in existing ModelWrapper conditioning features
When we talk about `model_atom_array`, it isn't super clear what is meant there, which is that this is the AtomArray shape that will be accepted by the ModelWrapper given the current sequence.

manzuoni-astera:
## Summary

Fixes #109.

`model_atom_array` was documented inconsistently across the conditioning dataclasses, and `ProtenixConditioning` omitted it entirely. Standardized on one description across RF3 / Protenix / Boltz:

> The AtomArray of the atoms the model actually operates on for the current sequence. May differ from `true_atom_array` (padding, added/missing atoms, element differences, etc.). Used for atom reconciliation against the true structure.

Also resolves the `# TODO: figure out a standardized way of handling all these` note on `BoltzConditioning.true_atom_array`, since the naming/semantics are now documented consistently. Docs-only, no behavior change.

Closes #109.

### DIFF
--- a/src/sampleworks/models/boltz/wrapper.py
+++ b/src/sampleworks/models/boltz/wrapper.py
@@ -288,8 +288,10 @@ class BoltzConditioning:
     true_atom_array : AtomArray | None
         Original structure's atom array (occupancy-filtered) for reward computation.
         Boltz's internal atom representation may differ from this due to padding, etc.
-        # TODO: figure out a standardized way of handling all these! related to several
-        todos elsewhere
+    model_atom_array : AtomArray | None
+        The AtomArray of the atoms the model actually operates on for the current
+        sequence. May differ from ``true_atom_array`` (padding, added/missing atoms,
+        element differences, etc.). Used for atom reconciliation against the true structure.
     """
 
     s: Tensor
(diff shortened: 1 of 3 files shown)


src/sampleworks/models/boltz/wrapper.py
```diff
@@ -288,8 +288,10 @@ class BoltzConditioning:
     true_atom_array : AtomArray | None
         Original structure's atom array (occupancy-filtered) for reward computation.
         Boltz's internal atom representation may differ from this due to padding, etc.
-        # TODO: figure out a standardized way of handling all these! related to several
-        todos elsewhere
+    model_atom_array : AtomArray | None
+        The AtomArray of the atoms the model actually operates on for the current
```
marcuscollins, line 292: Nit: I'm an old-fashioned grammar curmudgeon and favor "atoms on which the model actually operates" to "atoms the model actually operates on"

marcuscollins (approved):
Could we change the grammar a little? It's just a nit, but I'm old-fashioned.

Feel free to merge when ready. Hopefully it won't force you to run GPU tests for this, lmk if needed and I can override.

PR #297, opened 2026-07-02
fix(imports,tests): support bare require_* decorators; clean up protpardelle temp dirs

### LINKED ISSUE #285: require_protpardelle decorator does not support bare @decorator usage
Human note: I requested code rabbit to create this issue. I don't quite understand the structure of this wrapper, but it does seem like it will incorrectly interpret the to-be-wrapped method as a "message" if the bare form is used. --Marcus

## Summary
In `src/sampleworks/utils/imports.py`, the `require_protpardelle` decorator factory (added around lines 187-230) only supports the parenthesized form `@require_protpardelle("message")`. When used as a bare decorator, `@require_protpardelle`, the decorated function object is passed in as the `message` argument instead of being wrapped, so the function is never actually wrapped/protected. The docstring's first example (`@require_protpardelle`) is therefore misleading and does not work as documented.

## Rationale
The same pattern is used for the other `require_*` decorators (`require_boltz`, `require_protenix`, `require_rf3`) in the same file, so they likely share this issue, but this report focuses on `require_protpardelle` per the originating review comment.

## Affected area
- File: `src/sampleworks/utils/imports.py`
- Function: `require_protpardelle` (and docstring examples)

## Suggested fix
Either:
1. Update `require_protpardelle` to detect when the first positional argument is a callable (bare decorator usage) vs. a string message, and branch accordingly to support both `@require_protpardelle` and `@require_protpardelle("message")`, or
2. Remove/correct the bare-decorator example in the docstring so only the parenthesized form `@require_protpardelle()` is documented/supported.

## Acceptance criteria
- [ ] `require_protpardelle` works correctly when used as `@require_protpardelle` (bare) or docstring is updated to remove that example.
- [ ] `require_protpardelle("custom message")` continues to work as before.
- [ ] Docstring examples accurately reflect supported usage.

## References
- PR: https://github.com/diff-use/sampleworks/pull/274
- Originating review comment: https://github.com/diff-use/sampleworks/pull/274#discussion_r3493412905
- Requested by: @marcuscollins

### LINKED ISSUE #286: Make sure temporary directories are cleaned up in tests
This got flagged by code rabbit in PR #274 
```
# Must be set before any `import protpardelle...` happens. Respect an
# externally configured directory (e.g. when real weights are available).
os.environ.setdefault(
    "PROTPARDELLE_MODEL_PARAMS", tempfile.mkdtemp(prefix="protpardelle_model_params_")
)
```
and again in test_protpardelle_wrapper.py.

To resolve this issue, make any necessary changes to make sure the temp directories are cleaned up after tests are completed. If no changes are needed, just add that to the comments that the directories are cleaned up, and how that is known (something in pytest, somewhere else in the code)  just so this doesn't get flagged again.

manzuoni-astera:
## Summary

Two P1 quick wins that live on this branch (both flagged by CodeRabbit on #274).

### #285 — `require_*` decorators don't support bare `@decorator` usage
The `require_protpardelle` factory (and its siblings `require_boltz` / `require_protenix` / `require_rf3` / `require_any_model`) only worked as `@require_x("msg")`. Used bare (`@require_x`), the decorated function was passed in as `message`, so the function was never wrapped/guarded — silently unprotected. Added a `callable(message)` check that re-dispatches through the factory, so **all three** forms now work:

```python
@require_protpardelle              # bare
@require_protpardelle()            # empty parens
@require_protpardelle("custom")    # with message
```

The docstrings' bare examples are now accurate. Every existing call site uses `@require_x()`, so no behavior change there.

### #286 — temp dirs leaked in protpardelle tests
`tests/models/protpardelle/conftest.py` and `test_protpardelle_wrapper.py` set the model-params dir with `os.environ.setdefault("PROTPARDELLE_MODEL_PARAMS", tempfile.mkdtemp(...))`. `setdefault` evaluates its default eagerly, so `mkdtemp()` ran — and leaked a dir — on **every** session even when the var was already set (e.g. real weights configured, or the other module already set it). Now guard on the var being unset, then register `atexit.register(shutil.rmtree, ..., ignore_errors=True)` so the throwaway dir is removed at process exit. Whichever module runs first owns creation+cleanup; the other no-ops.

Targets `mdc/add-protpardelle` since the affected code isn't on `main` yet.

Fixes #285. Fixes #286.

### DIFF
--- a/tests/models/protpardelle/conftest.py
+++ b/tests/models/protpardelle/conftest.py
@@ -6,7 +6,9 @@ directory to exist. The tests don't need real weights, so point
 imported, then build a small randomly-initialized ``ai-allatom`` model.
 """
 
+import atexit
 import os
+import shutil
 import tempfile
 import textwrap
 from pathlib import Path
@@ -14,11 +16,15 @@ from pathlib import Path
 import pytest
 
 
-# Must be set before any `import protpardelle...` happens. Respect an
-# externally configured directory (e.g. when real weights are available).
-os.environ.setdefault(
-    "PROTPARDELLE_MODEL_PARAMS", tempfile.mkdtemp(prefix="protpardelle_model_params_")
-)
+# Must be set before any `import protpardelle...` happens. Respect an externally
+# configured directory (e.g. when real weights are available); otherwise create a
+# throwaway dir and remove it at process exit so tests don't leak temp dirs.
+# `setdefault` can't own the created dir (it eagerly evaluates mkdtemp even when the
+# var is already set), so guard explicitly and register cleanup only when we created it.
+if "PROTPARDELLE_MODEL_PARAMS" not in os.environ:
+    _model_params_dir = tempfile.mkdtemp(prefix="protpardelle_model_params_")
+    os.environ["PROTPARDELLE_MODEL_PARAMS"] = _model_params_dir
+    atexit.register(shutil.rmtree, _model_params_dir, ignore_errors=True)
 
 protpardelle_models = pytest.importorskip(
     "protpardelle.core.models", reason="Protpardelle not installed in this environment"
--- a/src/sampleworks/utils/imports.py
+++ b/src/sampleworks/utils/imports.py
@@ -76,6 +76,11 @@ def require_boltz(message: str | None = None) -> Callable[[F], F]:
     ... def custom_function():
     ...     pass
     """
+    if callable(message):
+        # Bare ``@require_boltz`` usage: the decorated function arrives as ``message``.
+        # Re-dispatch so both ``@require_boltz`` and ``@require_boltz("msg")`` work.
+        return require_boltz()(message)
+
     default_message = "Boltz model wrapper is not available. Install with: pixi install -e boltz"
 
     def decorator(func: F) -> F:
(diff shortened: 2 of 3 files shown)


tests/models/protpardelle/conftest.py
```diff
@@ -6,19 +6,25 @@
...
-)
+# Must be set before any `import protpardelle...` happens. Respect an externally
+# configured directory (e.g. when real weights are available); otherwise create a
+# throwaway dir and remove it at process exit so tests don't leak temp dirs.
+# `setdefault` can't own the created dir (it eagerly evaluates mkdtemp even when the
+# var is already set), so guard explicitly and register cleanup only when we created it.
+if "PROTPARDELLE_MODEL_PARAMS" not in os.environ:
+    _model_params_dir = tempfile.mkdtemp(prefix="protpardelle_model_params_")
+    os.environ["PROTPARDELLE_MODEL_PARAMS"] = _model_params_dir
+    atexit.register(shutil.rmtree, _model_params_dir, ignore_errors=True)
```
marcuscollins, line 27: oooh, handy! I didn't know of this.

marcuscollins (approved):

PR #270, opened 2026-06-22
fix(protpardelle): pipe sampling options through guidance CLI

xraymemory:
## Summary
- add Protpardelle-specific CLI flags for config path and sampling controls
- forward config path into ProtpardelleWrapper construction
- forward sampling options into annotate_structure_for_protpardelle
- keep guidance --help lightweight by lazy-loading runtime dependencies

Fixes #258

## Validation
- uvx ty check src/sampleworks/utils/guidance_script_arguments.py src/sampleworks/utils/guidance_script_utils.py src/sampleworks/cli/guidance.py tests/cli/test_guidance_cli.py tests/models/protpardelle/test_protpardelle_wrapper.py
- uvx ruff check src/sampleworks/utils/guidance_script_arguments.py src/sampleworks/utils/guidance_script_utils.py src/sampleworks/cli/guidance.py tests/cli/test_guidance_cli.py tests/models/protpardelle/test_protpardelle_wrapper.py
- PYTHONPATH=src uvx --python 3.12 --with pytest pytest --noconftest tests/cli/test_guidance_cli.py -q
- PYTHONPATH=src uvx --python 3.12 --with pytest pytest --noconftest tests/models/protpardelle/test_protpardelle_wrapper.py -q (skipped locally: Protpardelle not installed)

PR #300, opened 2026-07-02
feat(add protpardelle): Add the Protpardelle-1c cc89 protein generative model.

### LINKED ISSUE #258: Pipe protpardelle options through to the model
Currently we only pass the ensemble size through to the actual Protpardelle model in the `_run_guidance` method (guidance_script_utils.py). We need to pipe more of the options through to control protpardelle more fully. Alternatively, some options may need to be passed through to the model when it is initialized as well.

### LINKED ISSUE #286: Make sure temporary directories are cleaned up in tests

marcuscollins:
@coderabbitai ignore

Adding Protpardelle, attempt 2 (see PR #274)

Closes #258 — protpardelle sampling options were piped through the guidance CLI in #270 (merged into `mdc/add-protpardelle`); lands in `main` with this PR.
Closes #286 — protpardelle test temp-dir cleanup was done in #297 (merged into `mdc/add-protpardelle`); lands in `main` with this PR.

Original description:

This is a working version of Protpardelle-1c in SampleWorks. It may still require parameter tuning and other updates--in particular it isn't clear how to use self-conditioning. However we can generate structures with this version and so I'm making this PR as a milestone.

Only a couple significant changes have been made outside the Protpardelle wrapper code itself. One is to make a features class that is unfrozen, so that we can pass self-conditioning input forward to the next Euler step during sampling. Another is that we define different sampling parameters for Protpardelle when instantiating the sampler. There is no CLI control for either change in this PR.

## Summary by CodeRabbit

* **New Features**
  * Added support for the Protpardelle-1c structure prediction model, including sequence-conditioned generation.
  * Introduced Protpardelle-specific CLI options and a packaged default configuration.
  * Added new installation environments/features and updated dependency sets for Protpardelle support.
* **Improvements**
  * Renamed model configuration fields from `model` to `model_name` across guidance, grid-search, and related outputs.
  * Enhanced guidance behavior to support Protpardelle-specific settings and updated sampling invocation.
* **Documentation**
  * Updated grid-search troubleshooting examples to use `model_name`.
* **Chores**
  * Adjusted `.actlignore` to ignore `data/` by default while keeping sample data included.


src/sampleworks/data/cc89_epoch415.yaml
```diff
@@ -0,0 +1,101 @@
...
+  auto_calc_sigma_data: true
+  chain_residx_gap: 200
+  dummy_fill_mode: zero
+  fixed_size: 512
+  mixing_ratios:
+  - 1.0
+  n_aatype_tokens: 21
+  n_examples_for_sigma_data: 500
+  pdb_paths:
+  - /scratch/users/tianyulu/augmented_ingraham_cath_bugfree
```
k-chrispens, line 11: Is this a required part of the hydra config? Might be worth letting an agent loose on this to figure out what's the minimal config we can keep alongside protpardelle

src/sampleworks/utils/guidance_script_arguments.py
```diff
@@ -197,6 +197,15 @@ def validate_model_checkpoint(
     "msa_path",
     "disable_chiral_features",
     "track_chiral_features",
+    "protpardelle_config_path",
+    "protpardelle_s_churn",
```
k-chrispens: Same comment as before here - I think several of these parameters don't get passed through and are instead general sampler params

src/sampleworks/utils/guidance_script_arguments.py
```diff
@@ -197,6 +197,15 @@ def validate_model_checkpoint(
...
+    "protpardelle_step_scale",
+    "protpardelle_sidechain_mode",
+    "protpardelle_skip_mpnn_proportion",
```
k-chrispens: And these parameters we likely won't need given we aren't doing design

src/sampleworks/utils/guidance_script_utils.py
```diff
@@ -482,6 +523,30 @@ def _run_guidance(args: GuidanceConfig, guidance_type: str, model_wrapper, devic
...
             recycling_steps=recycling_steps,
         )
+    elif "Protpardelle" in wrapper_class_name:
+        from sampleworks.models.protpardelle.wrapper import annotate_structure_for_protpardelle
+
+        structure = annotate_structure_for_protpardelle(
+            structure,
+            ensemble_size=args.ensemble_size,
+            num_steps=args.num_diffusion_steps or 500,
+            s_churn=getattr(args, "protpardelle_s_churn", 40.0),
```
k-chrispens: Similar parameter comments as earlier - are these necessary for initializing the model?

tests/models/protpardelle/test_protpardelle_wrapper.py
```diff
@@ -0,0 +1,367 @@
+"""Tests for the Protpardelle structure-model wrapper.
```
k-chrispens: Here is another "structure model wrapper" mention - should be flow model wrapper

src/sampleworks/models/protpardelle/wrapper.py
```diff
@@ -0,0 +1,770 @@
...
+    of ``configs/running/sampling_partial_diffusion_allatom.yaml`` in the
+    Protpardelle-1c repository).
+
+    Attributes
+    ----------
+    ensemble_size : int
+        Number of structures to sample for the input sequence (batch dim). Default 8.
+    num_steps : int
+        Number of denoising (ODE discretization) steps. Defaults to 200.
+    s_churn : float
```
k-chrispens: We should set the gamma value, step_scale, etc. from this when we create a sampler if we are going to keep these parameters

src/sampleworks/models/protpardelle/wrapper.py
```diff
@@ -0,0 +1,770 @@
...
+    skip_mpnn_proportion: float = 1.0
+    jump_steps: bool = False
+    uniform_steps: bool = True
+    temperature: float = 1.0
+    top_p: float = 1.0
+    extra_sampling_kwargs: dict[str, Any] = field(default_factory=dict)
+
+
+# TODO: I need to check whether these defaults make sense. In fact I don't think
+#  we will use most of these (e.g. skip_mpnn_proportion, since we don't do seq design.
```
k-chrispens: Should we make a separate issue on this or address in this PR?

src/sampleworks/models/protpardelle/wrapper.py
```diff
@@ -0,0 +1,770 @@
...
+            atom_mask=atom_mask,
+            atom37_residue_index=atom37_residue_index,
+            atom37_atom_index=atom37_atom_index,
+            sampling_kwargs=sampling_kwargs,
+            sequences=tuple(sequences),
+            x_self_conditioning=None,
+        )
+
+        # x_init is a shape-compatible reference drawn from a Gaussian prior.
+        # TODO: Claude put this in. I don't think we really need it. Our sampler
```
k-chrispens: Yes I agree that it shouldn't be necessary, Claude may have done this because of the reference structure handling I do in the other modelwrappers

k-chrispens (commented):
Looks good!! A few downstream issues we should track. Another main concern is that we don't have a GPU test for this, which will be important for integration and regressions down the line. We should also add this to the list of modelwrappers in tests/conftest.py so it gets passed through the current integration tests

tests/runs/test_runner.py
```diff
@@ -166,7 +169,7 @@ def test_rf3_partial_chiral_off_flag_present(monkeypatch: pytest.MonkeyPatch) ->
 def test_build_invocations_records_output_dir(monkeypatch: pytest.MonkeyPatch) -> None:
     """`run_grid_search.py` assumes its --output-dir exists; the runner must mkdir it."""
     monkeypatch.setenv("HOME", "/home/test")
-    preset = loader.load_preset("rf3_partial")
+    preset = loader.load_preset("rf3_partial", overrides=["jobs.0.gpu_count=1"])
```
smallfishabc, line 172: Just a quick question. Why do we need to override the GPU count in the test?

src/sampleworks/cli/guidance.py
```diff
@@ -5,16 +5,21 @@
...
 from sampleworks.utils.guidance_script_arguments import GuidanceConfig
-from sampleworks.utils.guidance_script_utils import get_model_and_device, run_guidance
 
 
 def main(argv: list[str] | None = None) -> int:
     config = GuidanceConfig.from_cli(argv)
+
+    from loguru import logger
+
+    from sampleworks.utils.guidance_script_utils import get_model_and_device, run_guidance
```
smallfishabc, line 15: A quick question, why should we import it here not at the starting of the file?

src/sampleworks/utils/imports.py
```diff
@@ -172,6 +199,56 @@ def wrapper(*args: Any, **kwargs: Any) -> Any:
...
+        "Protpardelle model wrapper is not available. Install with: pixi install -e protpardelle"
+    )
+
+    def decorator(func: F) -> F:
+        @functools.wraps(func)
+        def wrapper(*args: Any, **kwargs: Any) -> Any:
+            if not PROTPARDELLE_AVAILABLE:
+                error_msg = message or default_message
+                try:
+                    import pytest
```
smallfishabc, line 258: Will this be triggered if the pytest is accidentally installed in the users environment?

pyproject.toml
```diff
@@ -24,16 +24,52 @@ pypi-dependencies = {"boltz" = "*", "cuequivariance-torch" = "*", "cuequivarianc
 platforms = ["linux-64"]
 
 [tool.pixi.feature.boltz-osx]
-pypi-dependencies = {"boltz" = "*", "cuequivariance-torch" = "*", "rdkit" = ">=2025.3.6"}
+pypi-dependencies = {"boltz" = "*", "cuequivariance-torch" = "*", "rdkit" = ">=2025.3.6", "joblib" = "*", "sfcalculator-torch" = ">=0.3.2"}
```
DorisMai, line 27: sfc should be >=0.3.3

pyproject.toml
```diff
@@ -24,16 +24,52 @@ pypi-dependencies = {"boltz" = "*", "cuequivariance-torch" = "*", "cuequivarianc
...
+biotite = "*"
+dm-tree = "*"
+jaxtyping = "*"
+ml_collections = "*"
+modelcif = "*"
+transformers = "*"
+wandb = "*"
+
+[tool.pixi.feature.protpardelle.pypi-options.dependency-overrides]
+pandas = "==2.3.1"
```
DorisMai: do you still need to separately do this despite line 186 `[tool.pixi.pypi-options.dependency-overrides]`?

pyproject.toml
```diff
@@ -24,16 +24,52 @@ pypi-dependencies = {"boltz" = "*", "cuequivariance-torch" = "*", "cuequivarianc
...
+prody = "*"
+pyyaml = "*"
+scipy = "*"
+pytorch = ">=2.6.0,<2.8"
+tqdm = "*"
+typer = "*"
+
+[tool.pixi.feature.protpardelle.pypi-dependencies]
+protpardelle = {git = "https://github.com/ProteinDesignLab/protpardelle-1c.git"}
```
DorisMai: might want to pin a tag/version?
I am also a bit surprised to see the large diff here -- some deps (e.g. wandb, jupyter probably should be dev/analysis? and some deps (e.g. hydra-core, omegaconf, jaxtyping) are already declared in [project.dependencies] below?

pyproject.toml
```diff
@@ -24,16 +24,52 @@ pypi-dependencies = {"boltz" = "*", "cuequivariance-torch" = "*", "cuequivarianc
...
+huggingface_hub = "*"
+hydra-core = "*"
+jupyter = "*"
+numpy = ">=1.25.0,<2.0"
+omegaconf = "*"
+pandas = "==2.3.1"
+prody = "*"
+pyyaml = "*"
+scipy = "*"
+pytorch = ">=2.6.0,<2.8"
```
DorisMai: this resolved to cpu torch for me, moving under [tool.pixi.feature.protpardelle.pypi-dependencies] as `torch = ">=2.6.0,<2.8"` instead fixes it.

src/sampleworks/models/protpardelle/wrapper.py
```diff
@@ -0,0 +1,770 @@
...
+        # layout. The mapping is derived from the input structure's atom names.
+        atom_array = get_asym_unit_from_structure(structure, atom_array_index=0)
+        atom37_residue_index, atom37_atom_index = self._atom37_indices_from_atom_array(atom_array)
+
+        # This should be the way to make the atom mask:
+        #   atom_mask = atom37_mask_from_aatype(aatype, seq_mask) but it doesn't handle OXT, so:
+        atom_mask = torch.zeros(
+            (padded_len, ATOM37_NUM_ATOMS), dtype=torch.float, device=self.device
+        )
+        atom_mask[atom37_residue_index, atom37_atom_index] = 1
```
DorisMai, line 545: What happens if the input reference structure has missing atoms? it seems that the num_atom count from this `atom_mask` would be different from that using other models?

src/sampleworks/models/protpardelle/wrapper.py
DorisMai: it seems that this x_init is just in general not used, even for other models? should the protocol be updated then?

src/sampleworks/models/protpardelle/wrapper.py
```diff
@@ -0,0 +1,770 @@
...
+
+        atom_slots = np.array(
+            [residue_constants.atom_order[name] for name in atom37_names], dtype=np.int64
+        )
+
+        # Contiguous residue ordinal (0..L-1) per atom, in atom order across chains.
+        residue_starts = struc.get_residue_starts(atom_array)
+        is_start = np.zeros(len(atom_names), dtype=np.int64)
+        is_start[residue_starts] = 1
+        residue_ordinals = np.cumsum(is_start) - 1
```
DorisMai: same concern as before (line 418-423), if there are missing residues (flexible loop), would this numbering be off then? might be good to add a real structure test in addition to the existing toy/minimal ones.

src/sampleworks/models/protpardelle/wrapper.py
```diff
@@ -0,0 +1,770 @@
...
+        :meth:`ProtpardelleWrapper._convert_to_atom37`). Shared across the batch
+        (a single input structure), so it carries no batch dimension.
+    atom37_atom_index : Int[Tensor, "atoms"]
+        For each of the ``N`` flat atoms, the atom37 slot ``0..36`` it occupies,
+        i.e. ``residue_constants.atom_order[atom_name]``.
+    sampling_kwargs : dict[str, Any]
+        Keyword arguments forwarded to ``Protpardelle.sample``.
+    sequences : tuple[str, ...]
+        The input one-letter sequences, one per protein chain (for reference).
+    x_self_conditioning: Float[Tensor, "batch L 37"]
```
DorisMai: is this coordinate? should it be "batch L 37 3"?

src/sampleworks/utils/guidance_script_utils.py
```diff
@@ -482,6 +523,30 @@ def _run_guidance(args: GuidanceConfig, guidance_type: str, model_wrapper, devic
...
+            top_p=getattr(args, "protpardelle_top_p", 1.0),
+        )
+        edm_sampler_kwargs = {
+            "s_max": 80,
+            "s_min": 0.001,
+            "gamma_0": 0.08,
+            "gamma_min": 0.00,
+            "sigma_data": 10.3,
+            "step_scale": 1.0,
+        }
```
DorisMai: I don’t see anywhere protpardelle’s own sample() is directly called? What’s the intended usage of ProtpardelleWrapper_build_sampling_kwargs() then? It gets a bit more confusing as those (seemingly unused args) are exposed CLI, but the hard coded sampler args here are not exposed.

src/sampleworks/models/protpardelle/wrapper.py
```diff
@@ -0,0 +1,770 @@
...
+
+# Number of aatype tokens for the 21-token alphabet (20 canonical + X).
+# Matches ``data.n_aatype_tokens`` in the Protpardelle-1c configs (e.g. cc89.yaml).
+NUM_AATYPE_TOKENS = 21
+
+# Atomworks keeps selenomethionine atoms as ``SE`` while Protpardelle's atom37
+# representation uses the canonical methionine sulfur slot, ``SD``.
+ATOM37_ATOM_NAME_ALIASES = {"SE": "SD"}
+
+# get default model configurations from protpardelle-1c/cc89
```
DorisMai: I initially thought you were referring to `data/cc89_epoch415.yaml`. a comment like "get the default all-atom sampling protocol from the protpardelle package" might be more helpful, since this sampling config should be irrelevant to which specific protpardelle model?

DorisMai (commented):
I have some confusion points about ProtpardelleWrapper, the main one is probably regarding the topology, as other things might get sorted out in future prs.

pyproject.toml
marcuscollins: Done

pyproject.toml
marcuscollins: Done

pyproject.toml
marcuscollins: Added ref = "ee378400f25b801fa481028000f9060183d7fb4c"

I was being conservative, but I will remove duplicates at least.

pyproject.toml
marcuscollins: Good catch. No, this isn't necessary.

src/sampleworks/cli/guidance.py
marcuscollins: @smallfishabc Good question, I think something must have gotten messed up with the commit history, like a rebase was missed somewhere. I shouldn't need to change this file at all.

In general, imports inside functions are made to speed things up--e.g. if I only needed to import something when one particular function was called, and it was a slow import, I might guard it like this.

marcuscollins: Oh, I see what's going on.... Hmmm. This is a change I should have made separately, but it had to do with reconciling arguments across the different models (protenix, protpardelle, rf3, boltz1/2)

src/sampleworks/data/cc89_epoch415.yaml
marcuscollins: That's unclear, so I've left it in for now. I'll have the code rabbit agent create an issue and then reconcile this after this is merged.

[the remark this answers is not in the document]
marcuscollins: @coderabbitai create an issue to address this. I think we want to use safe empty default paths wherever possible, and remove sections that we do not appear to need. For instance I suspect the "data" section is only used during training and we won't need it as part of our package. Once the current PR is merged, we'll work on that change.

src/sampleworks/models/protpardelle/wrapper.py
marcuscollins: I'll edit the comment. We _do_ require cc89_epoch415.yaml. If you trace everything through, that ultimately requires the file referred to here in order to do sampling.

src/sampleworks/models/protpardelle/wrapper.py
marcuscollins: Good catch. Fixed.

src/sampleworks/models/protpardelle/wrapper.py
marcuscollins: It turns out these are never used--we do set all of this in the sampler. This is vestigial code from my first attempt with Claude, which used Protpardelle.sample() in the .step() method. I rewired everything myself afterwards but forgot to remove this.

marcuscollins: I'll remove this and associated tests.

src/sampleworks/models/protpardelle/wrapper.py
marcuscollins: I decided to address here. I had Claude remove all these extra arguments that really belong to the sampler.

src/sampleworks/models/protpardelle/wrapper.py
marcuscollins: I think Doris is correct.

@coderabbitai please make an issue to verify our belief that this x_init is unused and if so remove x_init from GenerativeModelInput and clean up the related code.

src/sampleworks/models/protpardelle/wrapper.py
marcuscollins: @DorisMai do you mean what happens if we try to align this structure to the reference? We only convert to (and from) the atom37 layout inside `ProtpardelleWrapper.step`, the alignment is done inside `AF3EDMSampler.step`, the same for all models, and the `reconciler` Karson created handles this by aligning only on atoms the two have in common. 

It turns out `atom_mask`'s only use is to convert back from atom37 to "normal" coordinates.

src/sampleworks/models/protpardelle/wrapper.py
marcuscollins: This is only used to map from an AtomArray{Stack} to the atom37 format, so it's a contiguous tensor index, not an actual residue index. Now, there is a question about whether those missing residues would be in our input to begin with. The models we use generally require a complete sequence, so for the coordinates the model is operating on, there shouldn't be any missing residues.

src/sampleworks/utils/guidance_script_arguments.py
marcuscollins: removed.

src/sampleworks/utils/guidance_script_utils.py
marcuscollins: Yep, I've stripped all of this out.

src/sampleworks/utils/guidance_script_utils.py
marcuscollins: It's vestigial. When I first wrote this, I had Claude frame it out for me, and that did call Protpardelle.sample inside of ProtpardelleWrapper.step (which it should not have done, and which makes no sense). I rewrote the step method to call .forward directly, but forget to take all of this stuff out.

src/sampleworks/utils/imports.py
marcuscollins: Yes, in which case you'd end up with a less clear error message, although it would eventually raise ImportError from somewhere else.

tests/runs/test_runner.py
marcuscollins: Our test runners only have one GPU, but the job configuration uses 2 by default.

manzuoni-astera (changes requested):
Blocking issues remain in the Protpardelle integration at current head `a88eb4d`.

1. Atom37 output is flattened in canonical mask order instead of being mapped back through explicit residue/atom indices to the input AtomArray order. Common PDB atom ordering therefore associates generated coordinates with the wrong atoms.
2. Residue positions are derived from only residues present in the AtomArray while conditioning uses full canonical sequences. Missing loops shift every subsequent mapping.
3. Non-protein chains are skipped in sequence conditioning but remain in atom37 conversion, causing unknown atoms or out-of-range/misaligned residue indices.
4. The published/Astera image does not install the Protpardelle environment or provide checkpoint/model-parameter resolution, and CI never runs `protpardelle-dev`; the new tests therefore skip in existing jobs.

Please use explicit atom index maps/reconciliation for realistic incomplete and heterogeneous structures, provision the image runtime, and add a Protpardelle CI smoke test with realistically ordered atoms.

pyproject.toml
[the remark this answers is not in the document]
marcuscollins: Done

run_grid_search.py
```diff
@@ -587,7 +587,7 @@ def save_results(
     for existing_run in existing_runs:
         key = (
             existing_run.get("protein"),
-            existing_run.get("model"),
+            existing_run.get("model_name"),
```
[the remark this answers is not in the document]
marcuscollins: Ignoring. We have decided not to keep backwards compatibility.

xraymemory:
Diagnosed the `lock file not up-to-date` failure. **It's not a version skew** — I pinned CI to the exact pixi that generated the lock (`pixi-version: v0.73.0`, commit 97c877c) and it *still* fails, so #300's committed `pixi.lock` genuinely doesn't match `pyproject.toml`.

**Why `pixi update` didn't fix it:** the lock can't be regenerated on macOS. This workspace has `system-requirements.cuda = 12` plus git/editable deps (`protpardelle` git, `rc-foundry` editable, `sampleworks` editable), so a re-solve has to *build* those for `linux-64`, which a Mac can't do:
```
build dispatch initialization failed: the environment 'boltz-analysis' does not support 'osx-arm64'
linux-64-cuda-12 (subdir linux-64): requires subdir 'linux-64', which this machine can't run
```
That's almost certainly why your `pixi update` on macOS produced a lock CI rejects.

**Fix — regenerate on linux-64 with pixi 0.73.0:**
```bash
# on a linux box / ACTL pod (which already has CUDA):
pixi self-update --version 0.73.0
pixi lock
git add pyproject.toml pixi.lock && git commit -m 'chore: relock on linux' && git push
```
The `pixi-version` pin I added means CI will then validate with the same 0.73.0 you locked with. I can also add a `workflow_dispatch` 'relock' CI job that regenerates the lock on ubuntu and commits it, if you'd rather not do it by hand — say the word.

**Two unrelated things #300 still needs for the protpardelle tests to actually run** (separate from the lock): (1) add `protpardelle-dev` to the **cpu-tests** matrix (it's only in typecheck right now), and (2) update `test_at_least_one_available` to include `PROTPARDELLE_AVAILABLE` — it currently asserts only boltz/protenix/rf3, so it'll fail in a protpardelle-only env.

xraymemory:
✅ **Lock fixed.** Regenerated `pixi.lock` on Linux via the Relock workflow (`282e32d`). It had to run on ubuntu with `CONDA_OVERRIDE_CUDA=12` — a GPU-less runner is missing the `__cuda>=12` virtual package that `system-requirements.cuda` needs, and macOS can't build the linux-64 deps at all. Key deps are unchanged (numpy 1.26.4, boltz 2.2.1 both before/after), so it was stale-lock cleanup, not a version bump. The `lock file not up-to-date` error is gone and CI now installs cleanly.

Also pushed: `protpardelle-dev` added to the cpu-tests matrix, and `test_at_least_one_available` now counts `PROTPARDELLE_AVAILABLE`.

**What's now failing is #300's own code** (was hidden behind the lock error the whole time — CI never got past install before):
- `TypeError: MismatchCaseWrapper.step() takes 3 positional arguments but 4 ... were given` at `src/sampleworks/core/samplers/edm.py:429` — the sampler calls `.step()` with a signature the wrapper doesn't accept. This is the step()/Protocol reconciliation being incomplete; it fails across **all** envs (breaks `test_mismatch_integration` wholesale).
- `ty`: `ProtpardelleWrapper.__init__` called without `checkpoint_path`/`config_path`/`device`, and a missing `model_name` argument.
- `ruff` lint findings.

These are real and worth fixing before merge, but they're separate from the infra work. Happy to take a pass at the `step()` signature mismatch if useful — it looks like the highest-leverage one since it takes out the whole integration suite.

The reusable one-click Relock lives on #334 (pin + workflow), which needs a review to merge to main.

xraymemory (approved):
Verified the atom37 round-trip now gathers through the explicit residue/atom index maps (fixes the atom-order swap), non-protein chains are filtered before mapping, and the `step()` signature is reconciled. CI is green across every env including `protpardelle-dev`. Residue-vs-canonical mapping stays the known, acknowledged risk for the 40-structure run. LGTM — approving.

PR #301, opened 2026-07-03
feat, fix(metada carrying): fix metadata dropping through the pipeline (atomworks, predictors), eliminate the post-fact patchign script.

rtviii:
Sampleworks predictors emit CIF files with inconsistent sequence numbering and chain ids, and
without the categories a strict consumer needs — so evaluators like PDB-REDO's `tortoize` refuse
to process them, and atom selections on the input and output don't line up. Until now that was
patched after the fact by `patch_output_cif_files.py`, a re-fetch-and-graft script run on every
output.

This branch moves that work into the pipeline. Sampleworks now writes a complete, self-consistent,
provenance-stamped CIF at generation time, reproducible from public RCSB data, and the patch script
is deleted.

Validation runs against three independent tools throughout: `mmcif-validator` (format integrity),
`tortoize` (backbone geometry — the strict one), and `phenix.clashscore`.

## TL;DR

- Inputs are derived from an RCSB original plus a declared, hashed transform, and cached.
- On write-out we carry the depositor's real `_entity` / `_entity_poly` / `_entity_poly_seq` onto
  the prediction, reconcile the numbering back to the input, and add `_struct_asym`.
- Trial metadata is embedded as `_sampleworks` / `_sampleworks_result` and survives atomworks and
  biotite round-trips.
- Multi-entity structures work: entities are matched by sequence, so the carry survives a predictor
  relabeling every chain to A.
- `patch_output_cif_files.py` is gone; analysis presets read `refined.cif` directly.

## How the data flow changed

**Single chain — before vs after**

fig1_flow_before

fig2_flow_after

**mmCIF categories at each stage — before vs after** (solid = real category, hatched = stub,
red strike = missing → the file fails validation)

fig3_categories_before

fig4_categories_after

**Multi-chain (the entity-carry fix), simplified — before vs after**

fig5_multichain_before

fig6_multichain_after

## What the old patch script got wrong, and how it's fixed

Each shortcoming the patch script carried, and how the pipeline handles it now. Code lives in
`src/sampleworks/utils/{input_prep,cif_utils,guidance_script_utils}.py`.

| Problem in the post-hoc script | How it works now | Where | Status |
|---|---|---|---|
| PDB id recovered by regex on the file path | provenance is written into the file; the carry runs at generation with the job's known pdb id | `_carry_real_entity_categories`, `_sampleworks` | fixed |
| more than one entity was silently skipped | per-entity carry loop; each output entity matched by residue-name sequence, so it survives chain relabeling | `carry_entity_categories`, `_match_polymer_entity` | fixed |
| entity id created without verifying it | the match asserts agreement, otherwise degrades to a typeless stub rather than inventing an id | `_match_polymer_entity` | fixed |
| `_entity_poly_seq.num` never checked against `_atom_site.label_seq_id` | reconcile shifts the numbering per entity so the two agree | `carry_entity_categories` | fixed |
| `_entity` / `_entity_poly` could describe more than the modeled atoms | carried subset to the kept chains; `_chem_comp` extended to unmodeled canonical residues | `carry_entity_categories` | fixed |
| output indistinguishable from an experimental (XRD/MX) deposit | `_sampleworks` provenance stamped in — but we don't yet rewrite the depositor's method fields | `add_category_to_cif` | partial |
| the whole crawl-and-patch pass | retired — everything above happens inline at write time | `save_everything` | fixed |

### Reproducible inputs
The input CIF is derived from an RCSB original plus a declared transform (`CarveSpec` + `spec_hash`),
applied by a pure `carve()` and cached by `prepare_input()`, driven from a registry of specs. It
reproduces the existing hand-made inputs (2YL0 / 9BN8 / 1vme) residue-for-residue. Shipped as a prep
step (`scripts/prepare_inputs.py`); generation reads the cached input path from the proteins CSV, so
`run_grid_search.py` / `ProteinInput.from_csv` are untouched. Code: `input_prep.py`.

### Complete and consistent output
Everywhere Sampleworks controls the structure, the representation is now complete and matches the
input:

- residue ids match across representations; sequences are trivially alignable — end deletions and
  unmodeled-gap insertions are allowed, substitutions are rejected
- predictor renumbering/relettering is reset back to the input (match by sequence, then reconcile);
  a predictor that resets everything to chain A and renumbers from 1 is mapped back correctly
- the carried `_entity` / `_entity_poly` / `_entity_poly_seq` are checked for agreement with the
  `_atom_site` records
- the output aligns to the original — chain B stays chain B, same numbering and sequence
- input, output, and in-memory CIFs all pass `mmcif-validator`

Code: `cif_utils.py` (`carry_entity_categories`, `add_completeness_categories`, `_match_polymer_entity`).

### Metadata that sticks
Trial metadata is written into the CIF as `_sampleworks` (configuration) and `_sampleworks_result`
(outcome), with carve provenance on the cached input. It doesn't break `tortoize` or
`phenix.clashscore`, and it survives an atomworks/biotite load-modify-save round-trip. Captured: the
JobConfig / JobResult / GuidanceConfig fields plus the sequence and chain transforms. Not yet
captured: de-containerized real input-file paths (a candidate for the DiffUSE Hub). Code:
`guidance_script_utils.save_everything`, `cif_utils.add_category_to_cif`.

## Deliberate deviations from the literal spec

- The input carries completeness stubs, not the full entity sequence. The real `_entity_poly_seq`
  lives on the output, and atomworks `parse` ignores entity categories on the input anyway.
- Three numbering policies (`preserve_auth` | `preserve_label` | `from_one`), because the hand-made
  knowns are inconsistent — 2YL0/9BN8 use author numbering, but 1vme uses canonical numbering (9–410).
- Per-residue altloc resolution rather than a flat "keep A" — 1vme residue 356 is altloc-B-only, and
  a flat filter dropped it and fragmented the chain.

## Not in this PR

- Rewriting the depositor's experimental-method fields so the output no longer looks like an XRD/MX
  deposit.
- A full three-tool validation run on a multi-entity deposit — the end-to-end runs below were
  single-chain, and a multi-entity structure whose modeled residues don't form one contiguous run
  currently degrades to a typeless `_entity` stub instead of a full carry.
- De-containerized real input-file paths / DiffUSE Hub migration.
- Having generation invoke the carve inline — it is a decoupled prep step by design.

## Validation

All three tools green end-to-end on 1vme and 6b8x (pod runs): `mmcif-validator` zero errors,
`tortoize` real z-scores, `phenix.clashscore` real per-model scores. Multi-entity carry is
unit-tested (heterodimer / homodimer / per-entity reconcile) and verified on Fab 1fvc locally.

MSE footnote: `phenix.clashscore` flags an intra-residue CE–CG pair on selenomethionines. The
residue is written correctly (SE present and correctly named); this is a cosmetic MolProbity
artifact, not a defect.

Tests: `tests/utils/test_input_prep.py` (~38) and `tests/utils/test_cif_utils.py` (~63), plus a
per-model `tortoize` harness test.

## Summary by CodeRabbit

* **New Features**
  * Added a reproducible input-preparation workflow for generating cached input CIFs from a registry.
  * Evaluation workflows now run directly on `refined.cif` outputs, with improved support for multi-model structures and preserved CIF metadata.
  * Added more reliable structure-saving behavior that keeps evaluation and provenance details in output files.

* **Bug Fixes**
  * Updated evaluation result scanning to better handle metadata in CIF files and fall back safely when needed.
  * Improved atom-index reconciliation so saved coordinates align more consistently with the input structure.


rtviii:
Holy moly, i've never felt more "AI-native" and "agent pilled".

scripts/eval/EVALUATION.md
```diff
@@ -16,6 +16,11 @@ running, raising an error if it is not available.
 
 # Running the evaluations
 ## Preparing the output CIF files
+
+> Obsolete. Sampleworks now writes eval-ready `refined.cif` directly — the generation pipeline
+> carries the depositor's real entity categories and the input residue numbering onto the output, so
+> the metric scripts run against `refined.cif` with no preparation. The `patch_output_cif_files.py`
+> step described below has been retired and removed; run the scripts against `refined.cif`.
 As of this writing, Sampleworks outputs CIF files that primarily contain the output atomic
```
marcuscollins, line 24: Probably should remove the bit about requiring patching, since Sampleworks now outputs CiF files that should be complete.

README.md
```diff
@@ -256,9 +256,10 @@ rebuilt `pixi-with-checkpoints:sampleworks` image instead.
 
 `run_analysis` uses the same TOML runner as `run_experiments`, but loads presets
 from `analyses/*.toml` and runs the scripts under `scripts/eval/`.
-The `analyze_grid_search`, `all`, and `external_tools` presets first run a sequential
-`patch_outputs` pre-job, which creates `refined-patched.cif` files from each
-`refined.cif` before the evaluation jobs start.
+The `analyze_grid_search`, `all`, and `external_tools` presets run their evaluation jobs
```
marcuscollins, line 259: It may confuse users who aren't familiar with the package to say no CIF-patching is needed, since they won't understand it was previously required. You might say "... so the previously-required CIF-patching step is no longer required"

src/sampleworks/eval/grid_search_eval_utils.py
```diff
@@ -109,6 +110,138 @@ def parse_trial_dir(trial_dir: Path) -> dict[str, int | float | None]:
     }
 
 
+def _none_if_null(value: str | None) -> str | None:
+    """Normalize CIF nulls / blanks (``?`` / ``.`` / empty) to None; else the stripped string."""
+    if value is None:
+        return None
+    s = str(value).strip()
```
marcuscollins, line 117: Nit: Because `str(None)` evaluates to the string `"None"`, you could do this more cleanly by adding `"None"` to the set of "blank" strings. I prefer not to have separate paths if one path can suffice.

src/sampleworks/eval/grid_search_eval_utils.py
```diff
@@ -109,6 +110,138 @@ def parse_trial_dir(trial_dir: Path) -> dict[str, int | float | None]:
...
+    return None if s in {"?", ".", ""} else s
+
+
+def _to_int(value: str | None) -> int | None:
```
marcuscollins, line 121: Add a docstring to explain when this is needed.

scripts/eval/run_and_process_tortoize.py
```diff
@@ -133,12 +136,17 @@ def get_protein_level_z_scores(tortoize_json: dict[str, Any]) -> pd.DataFrame:
 
 def get_stats_for_single_path(path: Path) -> tuple[DataFrame, DataFrame]:
     """
-    Run tortoize on a single CIF file and extract residue/protein-level stats.
+    Run tortoize per model on a single CIF file and extract residue/protein-level stats.
+
+    tortoize is single-model: it loads zero atoms from a stacked ensemble. Each model is extracted
```
marcuscollins, line 141: I recall differently. I am 99.9% sure tortoize outputs scores separately for each model. I know this because I have output that shows different values for each model, e.g., /mnt/diffuse-shared/raw/sampleworks/combined_occ_sweep_results/tortoize_residues.csv. You can also see that tortoize iterates over models here: https://github.com/PDB-REDO/tortoize/blob/trunk/src/tortoize.cpp#L1087

You should probably revert this particular change unless you can show that tortoize does not behave as advertised.

scripts/prepare_inputs.py
```diff
@@ -0,0 +1,80 @@
...
+RCSB deposit, applies the declared carve, and writes the cached input CIF at
+``<cache-dir>/<pdb_id>__<spec_hash>.cif`` -- the file the generation pipeline's ``structure``
+column already expects. One ``<pdb_id>\t<path>`` line is printed per prepared input.
+
+This is a decoupled prep step: it *produces* the input file; ``run_grid_search.py`` and
+``ProteinInput.from_csv`` are unchanged. Point a proteins-CSV ``structure`` column at a printed
+path -- use the absolute path, since ``from_csv`` resolves relative paths against the CSV's own
+directory, not the working directory.
+
+Run with sampleworks importable (e.g. ``PYTHONPATH=src`` or inside the pixi env). The RCSB deposit
```
marcuscollins, line 13: remove comment about PYTHONPATH. We should encourage everyone to use pixi, or if they are experienced, they will know they can modify PYTHONPATH as needed. Setting `PYTHONPATH=src` is not something an average user should do.

scripts/prepare_inputs.py
```diff
@@ -0,0 +1,80 @@
...
+    python scripts/prepare_inputs.py                       # every pdb id in the default registry
+    python scripts/prepare_inputs.py 1vme 2yl0             # just these two
+    python scripts/prepare_inputs.py --cache-dir /tmp/in --registry my_registry.json
+"""
+
+from __future__ import annotations
+
+import argparse
+import sys
+from importlib.resources import files
```
marcuscollins, line 27: Did you run `ruff` on this code? I thought it demanded a blank line between `import` and `from` statements?

scripts/prepare_inputs.py
```diff
@@ -0,0 +1,80 @@
...
+    parser.add_argument(
+        "pdb_ids",
+        nargs="*",
+        help="PDB ids to prepare (default: every id in the registry).",
+    )
+    parser.add_argument(
+        "--registry",
+        type=Path,
+        default=_DEFAULT_REGISTRY,
+        help="Carve registry JSON keyed by pdb id (default: sampleworks/data/input_registry.json).",
```
marcuscollins, line 49: Is it necessary to use a JSON? I'm trying to think about end-users here who are perhaps less familiar with that format. Perhaps you can take a CSV or JSON?

scripts/prepare_inputs.py
```diff
@@ -0,0 +1,80 @@
...
+    parser.add_argument(
+        "--cache-dir",
+        type=Path,
+        default=None,
+        help="Output cache directory (default: ~/.sampleworks/inputs).",
+    )
+    parser.add_argument(
+        "--rcsb-cache",
+        type=Path,
+        default=None,
```
marcuscollins, line 60: Set up a default in the `$HOME/.sampleworks` directory used by other parts of the package.

scripts/prepare_inputs.py
```diff
@@ -0,0 +1,80 @@
...
+        help="Directory of <pdb_id>.cif deposit originals, for offline reproduction.",
```
marcuscollins, line 61: It shouldn't just be offline. We don't want to hit any server more than once for the same information.

src/sampleworks/utils/input_prep.py
```diff
@@ -0,0 +1,487 @@
+"""Reproducible preparation of model-input CIFs from RCSB deposits.
+
+A sampleworks input CIF is a *carve* of an RCSB deposit: one or more polymer chains, with
+waters / ligands dropped and a chosen residue numbering, written as ``_atom_site`` plus the
+carried unit cell / connectivity and the minimal completeness categories that keep it
+mmcif-validator clean. (The depositor's real entity sequence is carried onto the prediction
```
marcuscollins, line 6: By "entity sequence" here, do you mean the actual CIF category _entity_sequence or something else? I would be clear about this because for the less knowledgeable it could be confusing and make people think our input doesn't have a sequence that matches the original PDB entry (which should not be the case.)

src/sampleworks/utils/cif_utils.py
```diff
@@ -330,3 +337,684 @@ def _normalize_nulls(value: Any) -> Any:
...
+    """Download (and cache) the original mmCIF deposit for ``pdb_id`` from RCSB.
+
+    Reuses the on-disk cache the retired ``patch_output_cif_files.py`` used. biotite's ``fetch``
+    only hits the network when the file is not already present, so once the cache is warm this is
+    offline-friendly and cheap to call per generation.
+
+    Parameters
+    ----------
+    pdb_id : str
+        The 4-character PDB id (case-insensitive), e.g. ``"1vme"``.
```
marcuscollins, line 581: Feng has an open PR to use the new pdb_00001ABC style PDB IDs. Please make sure this is compatible.

src/sampleworks/utils/input_prep.py
```diff
@@ -0,0 +1,487 @@
...
+
+from sampleworks.utils.cif_utils import (
+    add_category_to_cif,
+    add_completeness_categories,
+    fetch_rcsb_cif,
+)
+
+
+# Altloc ids that mark an atom as having no alternate conformer (always kept).
+_BLANK_ALTLOCS = (".", "?", "")
```
marcuscollins, line 49: Create an issue to gather constants that get re-used in multiple places into one constants file. We already have sampleworks.eval.constants but we should further collect them. This one for instance is essentially the same as `sampleworks.utils.atom_array_utils.BLANK_ALTLOC_IDS`. 

Tag the new issue "engineering" and "P1".

src/sampleworks/utils/input_prep.py
```diff
@@ -0,0 +1,487 @@
...
+_VALID_DROP_TOKENS = frozenset({"water", "ligand"})
+_DEFAULT_DROP = frozenset({"water", "ligand"})
+
+_NUMBERING_POLICIES = ("preserve_auth", "preserve_label", "from_one")
+_DEFAULT_INPUT_CACHE = Path("~/.sampleworks/inputs")
+
+# The standard _atom_site coordinate columns an input CIF carries (the shape the existing inputs
+# and the output writer use). Deposit-specific extras (e.g. pdbx_tls_group_id, which references a
+# refinement category the input does not carry) are dropped so the file stays mmcif-validator clean.
+_STANDARD_ATOM_SITE_COLUMNS = (
```
marcuscollins, line 63: This seems like it belongs in cif_utils.py

src/sampleworks/utils/input_prep.py
```diff
@@ -0,0 +1,487 @@
...
+class CarveSpec:
+    """A declared transform from an RCSB deposit to a sampleworks input CIF.
+
+    Parameters
+    ----------
+    chains : list[str]
+        Author chain ids (``auth_asym_id``) to keep, in order. Each becomes its own output entity.
+    drop : frozenset[str]
+        Entity classes to drop. ``"water"`` drops water entities; ``"ligand"`` drops
+        non-polymer / branched / macrolide entities. Polymer chains are never dropped, so modified
```
marcuscollins, line 98: What about RNA/DNA that is bound? Do we handle that by chain selection? I would comment on this specifically because (I think) not all our structure models will handle nucleic acid polymers.

src/sampleworks/utils/input_prep.py
```diff
@@ -0,0 +1,487 @@
...
+    "pdbx_formal_charge",
+    "auth_seq_id",
+    "auth_comp_id",
+    "auth_asym_id",
+    "auth_atom_id",
+    "pdbx_PDB_model_num",
+)
+
+
+@dataclass
```
marcuscollins, line 88: Should it use `frozen=True`?

src/sampleworks/utils/input_prep.py
```diff
@@ -0,0 +1,487 @@
...
+# Altloc ids that mark an atom as having no alternate conformer (always kept).
+_BLANK_ALTLOCS = (".", "?", "")
+
+# Recognised ``drop`` tokens. This cut keeps polymer entities and drops everything else, so the
+# only supported configuration is dropping both; the tokens remain a declared, hashable record of
+# intent and the anchor for a future richer drop vocabulary (e.g. retaining a ligand).
+_VALID_DROP_TOKENS = frozenset({"water", "ligand"})
+_DEFAULT_DROP = frozenset({"water", "ligand"})
+
+_NUMBERING_POLICIES = ("preserve_auth", "preserve_label", "from_one")
```
marcuscollins, line 57: why use tuples here and frozenset elsewhere? I favor consistency to avoid any confusion.

src/sampleworks/utils/input_prep.py
```diff
@@ -0,0 +1,487 @@
...
+
+    chains: list[str]
+    drop: frozenset[str] = field(default_factory=lambda: frozenset(_DEFAULT_DROP))
+    numbering: str = "preserve_auth"
+    selection: str | None = None
+
+    def __post_init__(self) -> None:
+        if not self.chains:
+            raise ValueError("CarveSpec.chains must name at least one chain")
+        self.chains = [str(c) for c in self.chains]
```
marcuscollins, line 117: Is this necessary?

src/sampleworks/utils/input_prep.py
```diff
@@ -0,0 +1,487 @@
...
+        self.drop = frozenset(self.drop)
+        unknown = self.drop - _VALID_DROP_TOKENS
+        if unknown:
```
marcuscollins, line 120: I recommend using the new throwaway variable syntax: `if unknown := self.drop - _VALID_DROP_TOKENS:` which will only realize the variable if the block below is to be evaluated.

src/sampleworks/utils/input_prep.py
```diff
@@ -0,0 +1,487 @@
...
+            raise ValueError(f"CarveSpec.drop has unknown tokens {sorted(unknown)}")
+        if self.numbering not in _NUMBERING_POLICIES:
+            raise ValueError(
+                f"CarveSpec.numbering must be one of {_NUMBERING_POLICIES}, got {self.numbering!r}"
+            )
+
+    def to_dict(self) -> dict:
```
marcuscollins, line 127: I'm pretty sure you don't need this and can use the built-in `asdict` method of dataclasses.

src/sampleworks/utils/input_prep.py
```diff
@@ -0,0 +1,487 @@
...
+        Entity classes to drop. ``"water"`` drops water entities; ``"ligand"`` drops
+        non-polymer / branched / macrolide entities. Polymer chains are never dropped, so modified
+        polymer residues (HETATM that belong to a polymer entity) are kept. Default
+        ``{"water", "ligand"}``.
+    numbering : str
+        How the kept residues are numbered (output ``label_seq_id == auth_seq_id``):
+        ``"preserve_auth"`` uses the deposit author numbering; ``"preserve_label"`` uses the deposit
+        canonical (``label_seq_id``) numbering; ``"from_one"`` renumbers ``1..N`` per chain.
+    selection : str | None
+        Reserved for an atom-selection refinement; not yet implemented (raises if set).
```
marcuscollins, line 106: It does not appear to raise an error if the value is something other than None.

src/sampleworks/utils/input_prep.py
```diff
@@ -0,0 +1,487 @@
...
+        )
+
+    def spec_hash(self) -> str:
+        """A stable 12-hex-char digest of the spec.
+
+        Canonical: keys sorted, ``chains`` and ``drop`` sorted, so the hash is independent of the
+        order chains were listed in. The cache key for a prepared input is ``<pdb_id>__<hash>``.
+        """
+        canonical = {
+            "chains": sorted(self.chains),
```
marcuscollins, line 153: If in `post_init` you sorted chains and drop, then you could just use `asdict` here to get the dictionary, which would be a little more robust.

src/sampleworks/utils/input_prep.py
```diff
@@ -0,0 +1,487 @@
...
+    raw = json.loads(Path(path).expanduser().read_text())
+    return {pdb_id.lower(): CarveSpec.from_dict(spec) for pdb_id, spec in raw.items()}
+
+
+def _longest_consecutive(nums: list[int]) -> list[int]:
+    """Longest run of consecutive integers in a sorted unique list."""
+    if not nums:
+        return []
+    best = cur = [nums[0]]
+    for value in nums[1:]:
```
marcuscollins, line 173: You can push this method down to numpy to make it potentially faster:
```
nums_np = np.asarray(nums)
starts = np.where(np.append(-1, np.diff(nums_np)) != 1)[0]
lengths = np.diff(np.append(starts, len(nums_np))) 
idx_max = np.argmax(lengths)
best = nums_np[starts[idx_max]:starts[idx_max] + lengths[idx_max]]
```

This is at least 10x faster for lists of a few thousand items. It maybe doesn't matter here (a few ms vs. 10s of us), but it is good to look for vectorized ways of doing things.

src/sampleworks/utils/input_prep.py
```diff
@@ -0,0 +1,487 @@
...
+def _sole_block(cif: CIFFile) -> CIFBlock:
+    return cif[list(cif.keys())[0]]
+
+
+def _entity_types(block: CIFBlock) -> dict[str, str]:
+    """Map ``_entity.id -> _entity.type`` from a deposit block (empty if no ``_entity``)."""
+    if "entity" not in block:
+        return {}
+    entity = block["entity"]
+    ids = [str(i) for i in entity["id"].as_array(str)]
```
marcuscollins, line 189: This looks like you convert to str twice, which should not be necessary.

src/sampleworks/utils/input_prep.py
```diff
@@ -0,0 +1,487 @@
...
+    if "type" in entity:
```
marcuscollins, line 190: Add a comment as to why you're making this distinction would help.

src/sampleworks/utils/input_prep.py
```diff
@@ -0,0 +1,487 @@
...
+
+
+def _resolve_altlocs(
+    candidate: np.ndarray,
+    chains: np.ndarray,
+    auth_seq: np.ndarray,
+    alt: np.ndarray,
+    n: int,
+) -> np.ndarray:
+    """Pick one conformer per atom among ``candidate`` atoms.
```
marcuscollins, line 208: docstring should explain what each input argument is. Also, use longer more descriptive names. I _guess_ that `n` is... well I can't tell what `n` is. I guess `alt` is the array of altloc ids from an AtomArray or AtomArrayStack? This begs the question, why not just pass an AtomArray/Stack?

src/sampleworks/utils/input_prep.py
```diff
@@ -0,0 +1,487 @@
...
+
+    Keeps blank (``.``/``?``/empty) altlocs plus, for atoms that carry an altloc, the residue's
+    preferred conformer -- ``"A"`` when present, otherwise the lowest altloc id in that residue.
```
marcuscollins, line 211: What does "preferred" mean here? Highest occupancy?

src/sampleworks/utils/input_prep.py
```diff
@@ -0,0 +1,487 @@
...
+    Resolving per residue (rather than a flat "keep altloc A") means a residue whose only conformer
+    is labeled (e.g.) ``"B"`` is kept rather than dropped, which would otherwise fragment the chain.
+    """
+    preferred: dict[tuple[str, str], str] = {}
+    for i in np.nonzero(candidate)[0]:
```
marcuscollins, line 216: again I would try to vectorize this. It can actually make your code more clear, and is almost always much faster. 
```
idx = np.where((candidate != 0) & !np.isin(alt, _BLANK_ALTLOCS)[0]
a = alt.astype(str)[idx]
...

File: src/sampleworks/utils/input_prep.py [hunk 29]
```diff
@@ -0,0 +1,487 @@
...
+    block_name: str,
+) -> None:
+    """Carry ``_cell`` (whole) and ``_struct_conn`` (subset to kept residues) from the deposit.
+
+    ``_struct_conn`` rows survive only when *both* partners reference a kept chain and a kept
+    residue (so disulfides among kept residues are retained, conns to dropped ligands/waters are
+    not). Partner sequence ids are remapped through ``numbering_map`` so they match the carved
+    ``_atom_site`` numbering (an identity map under ``preserve_auth``).
+    """
+    if "cell" in src_block:
```
marcuscollins, line 244: make "cell" and similar strings a constant.

src/sampleworks/utils/input_prep.py
```diff
@@ -0,0 +1,487 @@
...
+
+
+def _carry_cell_and_struct_conn(
+    dst: CIFFile,
+    src_block: CIFBlock,
+    numbering_map: dict[tuple[str, str], str],
+    kept_by_chain: dict[str, set[str]],
+    block_name: str,
+) -> None:
+    """Carry ``_cell`` (whole) and ``_struct_conn`` (subset to kept residues) from the deposit.
```
marcuscollins, line 237: document input arguments and use better names. In particular I'm confused about the exact contents of `numbering_map` and `kept_by_chain`.

src/sampleworks/utils/input_prep.py
```diff
@@ -0,0 +1,487 @@
...
+
+    ``_struct_conn`` rows survive only when *both* partners reference a kept chain and a kept
+    residue (so disulfides among kept residues are retained, conns to dropped ligands/waters are
+    not). Partner sequence ids are remapped through ``numbering_map`` so they match the carved
+    ``_atom_site`` numbering (an identity map under ``preserve_auth``).
+    """
+    if "cell" in src_block:
+        cell = src_block["cell"]
+        # Drop the deposit's entry_id so add_completeness_categories can supply one consistent with
+        # the output _entry.id (the block name); keeping the deposit's would dangle the foreign key.
```
marcuscollins, line 247: dangle what foreign key?

src/sampleworks/utils/input_prep.py
```diff
@@ -0,0 +1,487 @@
...
+    for i in np.nonzero(candidate)[0]:
+        a = str(alt[i])
+        keep[i] = a in _BLANK_ALTLOCS or a == preferred.get((str(chains[i]), str(auth_seq[i])))
+    return keep
+
+
+def _carry_cell_and_struct_conn(
+    dst: CIFFile,
+    src_block: CIFBlock,
+    numbering_map: dict[tuple[str, str], str],
```
marcuscollins, line 233: I guess we keep the original chain ids? So this doesn't need to be a map from <chain_id_old>, <res_id_old> -> <chain_id_new>, <res_id_new> ?

src/sampleworks/utils/input_prep.py
```diff
@@ -0,0 +1,487 @@
...
+            {key: list(cell[key].as_array(str)) for key in cell if key != "entry_id"},
+            "cell",
+            overwrite=True,
+            block_name=block_name,
+        )
+
+    if "struct_conn" not in src_block:
+        return
+    struct_conn = src_block["struct_conn"]
+    cols = {key: list(struct_conn[key].as_array(str)) for key in struct_conn}
```
marcuscollins, line 259: Is it necessary to realize this here? I get that the evaluation is lazy, so `struct_conn` isn't completely ready, but most of the keys are never used. You could below in `_col` check `if key in struct_conn:` and only then get the category data.

src/sampleworks/utils/input_prep.py
```diff
@@ -0,0 +1,487 @@
...
+        # the output _entry.id (the block name); keeping the deposit's would dangle the foreign key.
+        add_category_to_cif(
+            dst,
+            {key: list(cell[key].as_array(str)) for key in cell if key != "entry_id"},
+            "cell",
+            overwrite=True,
+            block_name=block_name,
+        )
+
+    if "struct_conn" not in src_block:
```
marcuscollins, line 256: I think it might make sense to put this block of code in its own method, so it can be more clearly documented and tested.

src/sampleworks/utils/input_prep.py
```diff
@@ -0,0 +1,487 @@
...
+        return
+    struct_conn = src_block["struct_conn"]
+    cols = {key: list(struct_conn[key].as_array(str)) for key in struct_conn}
+    nrows = len(next(iter(cols.values()))) if cols else 0
+
+    def _col(*keys: str) -> list[str]:
```
marcuscollins, line 262: You have a near-exact duplicate of this method below. Instead of in-lining this method, make a version that takes `cols` as an argument, and use it in both places.

src/sampleworks/utils/input_prep.py
```diff
@@ -0,0 +1,487 @@
...
+        for key in keys:
+            if key in cols:
+                return cols[key]
```
marcuscollins, line 265: This will return on the first key that it finds, is that what you want? Is the idea that this is checking for columns in some fallback order? Don't you want to use the preferred numbering system (e.g. "preserve_auth")?

src/sampleworks/utils/input_prep.py
```diff
@@ -0,0 +1,487 @@
...
+            keep.append(i)
+    if not keep:
+        return
+
+    new_cols = {key: [values[i] for i in keep] for key, values in cols.items()}
+
+    # Rewrite each partner onto the carved atom_site: identify the residue by its deposit author
+    # (chain, seq), then set both the label and auth asym/seq ids to the carved chain and number
+    # (label == auth in the carve) so the connectivity resolves against the new _atom_site.
+    def _new_col(*keys: str) -> list[str] | None:
```
marcuscollins, line 290: Ditto comment on L262

src/sampleworks/utils/input_prep.py
```diff
@@ -0,0 +1,487 @@
...
+    keep = []
+    for i in range(nrows):
+        c1, s1 = _partner("ptnr1", i)
+        c2, s2 = _partner("ptnr2", i)
+        if _kept(c1, s1) and _kept(c2, s2):
+            keep.append(i)
+    if not keep:
+        return
+
+    new_cols = {key: [values[i] for i in keep] for key, values in cols.items()}
```
marcuscollins, line 285: Similar to the comment above--do you need to create this new copy of `cols`, or could you just re-use `cols` and modify it instead?

src/sampleworks/utils/input_prep.py
```diff
@@ -0,0 +1,487 @@
...
+    # (chain, seq), then set both the label and auth asym/seq ids to the carved chain and number
+    # (label == auth in the carve) so the connectivity resolves against the new _atom_site.
+    def _new_col(*keys: str) -> list[str] | None:
+        for key in keys:
+            if key in new_cols:
+                return new_cols[key]
+        return None
+
+    for prefix in ("ptnr1", "ptnr2"):
+        chains_here = _new_col(f"{prefix}_auth_asym_id", f"{prefix}_label_asym_id")
```
marcuscollins, line 297: Give these better names, e.g., `new_asym_id_column` or `new_chain_id_column`.

src/sampleworks/utils/input_prep.py
marcuscollins: Also, why use strings for sequence numbers?

src/sampleworks/utils/input_prep.py
```diff
@@ -0,0 +1,487 @@
...
+    the input -- the input needs only the validator's stub parents.)
+
+    Parameters
+    ----------
+    original : str | Path | CIFFile
+        The RCSB deposit (path or already-open ``CIFFile``).
+    spec : CarveSpec
+        The declared transform.
+    source : str
+        Provenance label (typically the PDB id); also used as the output block / entry name.
```
marcuscollins, line 328: if it's typically the PDB id, why not set the default to None and re-use what comes from `original`?

src/sampleworks/utils/input_prep.py
```diff
@@ -0,0 +1,487 @@
...
+    completeness categories so the file is mmcif-validator clean, and stamps ``_sampleworks`` carve
+    provenance. (The depositor's real entity *sequence* is carried onto the prediction output, not
+    the input -- the input needs only the validator's stub parents.)
+
+    Parameters
+    ----------
+    original : str | Path | CIFFile
+        The RCSB deposit (path or already-open ``CIFFile``).
+    spec : CarveSpec
+        The declared transform.
```
marcuscollins, line 326: This is not a particularly meaningful description.

src/sampleworks/utils/input_prep.py
```diff
@@ -0,0 +1,487 @@
...
+    src_block = _sole_block(src)
+    atom_site = src_block["atom_site"]
+    cols = {key: np.asarray(atom_site[key].as_array(str)) for key in atom_site}
+    n = len(next(iter(cols.values())))
+    if "label_entity_id" not in cols:
+        raise ValueError("deposit _atom_site lacks label_entity_id; cannot apply entity-aware drop")
+
+    entity_types = _entity_types(src_block)
+    chain_col = "auth_asym_id" if "auth_asym_id" in atom_site else "label_asym_id"
+    chains = cols[chain_col]
```
marcuscollins, line 350: Here's another place where you could re-use your `_cols` method from above.

src/sampleworks/utils/input_prep.py
```diff
@@ -0,0 +1,487 @@
...
+    entity_ids = cols["label_entity_id"]
+    alt = cols.get("label_alt_id", np.array(["."] * n))
+    auth_seq = cols["auth_seq_id"]
+    label_seq = cols["label_seq_id"]
+
+    def _entity_kept(entity_id: str) -> bool:
```
marcuscollins, line 356: you don't need this method, just inline it below:

```
type_keep = np.array([entity_types.get(str(e), "polymer") == "polymer" for e in entity_ids])
```

src/sampleworks/utils/input_prep.py
```diff
@@ -0,0 +1,487 @@
...
+        # Keep polymer entities (which carry modified residues like MSE/PCA as HETATM); drop water
+        # and non-polymer. Unknown entity ids default to kept.
+        return entity_types.get(entity_id, "polymer") == "polymer"
+
+    type_keep = np.array([_entity_kept(str(e)) for e in entity_ids])
```
marcuscollins, line 361: give `type_keep` a better name.

src/sampleworks/utils/input_prep.py
```diff
@@ -0,0 +1,487 @@
...
+    atom_site = src_block["atom_site"]
+    cols = {key: np.asarray(atom_site[key].as_array(str)) for key in atom_site}
+    n = len(next(iter(cols.values())))
+    if "label_entity_id" not in cols:
+        raise ValueError("deposit _atom_site lacks label_entity_id; cannot apply entity-aware drop")
+
+    entity_types = _entity_types(src_block)
+    chain_col = "auth_asym_id" if "auth_asym_id" in atom_site else "label_asym_id"
+    chains = cols[chain_col]
+    entity_ids = cols["label_entity_id"]
```
marcuscollins, line 351: lots of re-used string keys--these should all be constants to avoid mistakes.

src/sampleworks/utils/input_prep.py
```diff
@@ -0,0 +1,487 @@
...
+
+    def _entity_kept(entity_id: str) -> bool:
+        # Keep polymer entities (which carry modified residues like MSE/PCA as HETATM); drop water
+        # and non-polymer. Unknown entity ids default to kept.
+        return entity_types.get(entity_id, "polymer") == "polymer"
+
+    type_keep = np.array([_entity_kept(str(e)) for e in entity_ids])
+    candidate = type_keep & np.isin(chains, spec.chains)
+    if "pdbx_PDB_model_num" in cols:
+        first_model = next(iter(dict.fromkeys(cols["pdbx_PDB_model_num"].tolist())))
```
marcuscollins, line 364: Okay, this is a _weird_ construction. Are you sure it isn't equivalent to
```
first_model = cols["pdbx_PDB_model_num"].tolist()[0]
```
?

src/sampleworks/utils/input_prep.py
```diff
@@ -0,0 +1,487 @@
...
+    for entity_index, chain in enumerate(spec.chains, start=1):
+        mask = base & (chains == chain)
+        if not mask.any():
+            raise ValueError(f"no kept polymer atoms for chain {chain!r} in {source}")
+        present = sorted({int(r) for r in auth_seq[mask] if _is_int(r)})
+        run = set(_longest_consecutive(present))
+        mask &= np.array([_is_int(r) and int(r) in run for r in auth_seq])
+
+        # Map each kept residue's deposit auth id -> its output sequence id under the policy.
+        if spec.numbering == "from_one":
```
marcuscollins, line 383: use constants for re-used strings.

src/sampleworks/utils/input_prep.py
```diff
@@ -0,0 +1,487 @@
...
+            ordered = sorted({int(r) for r in auth_seq[mask]})
+            remap = {str(orig): str(rank) for rank, orig in enumerate(ordered, start=1)}
+        elif spec.numbering == "preserve_label":
+            remap = {str(auth_seq[i]): str(label_seq[i]) for i in np.nonzero(mask)[0]}
+        else:  # preserve_auth
+            remap = {str(orig): str(orig) for orig in {int(r) for r in auth_seq[mask]}}
+
+        total |= mask
```
marcuscollins, line 391: Give total a better name, e.g., `final_atom_site_mask`

src/sampleworks/utils/input_prep.py
```diff
@@ -0,0 +1,487 @@
...
+        mask = base & (chains == chain)
+        if not mask.any():
+            raise ValueError(f"no kept polymer atoms for chain {chain!r} in {source}")
+        present = sorted({int(r) for r in auth_seq[mask] if _is_int(r)})
+        run = set(_longest_consecutive(present))
+        mask &= np.array([_is_int(r) and int(r) in run for r in auth_seq])
+
+        # Map each kept residue's deposit auth id -> its output sequence id under the policy.
+        if spec.numbering == "from_one":
+            ordered = sorted({int(r) for r in auth_seq[mask]})
```
marcuscollins, line 384: Vectorize. Use `auth_seq.as_array`, set the type (if it isn't already an int, which it should be) and then use np.sort if needed.

marcuscollins: Incidentally, why wouldn't it be sorted? A PDB file should be in order, is there some edge case where it isn't?

src/sampleworks/utils/input_prep.py
```diff
@@ -0,0 +1,487 @@
...
+            remap = {str(orig): str(rank) for rank, orig in enumerate(ordered, start=1)}
+        elif spec.numbering == "preserve_label":
+            remap = {str(auth_seq[i]): str(label_seq[i]) for i in np.nonzero(mask)[0]}
+        else:  # preserve_auth
+            remap = {str(orig): str(orig) for orig in {int(r) for r in auth_seq[mask]}}
+
+        total |= mask
+        entity_out[mask] = str(entity_index)
+        kept_by_chain[chain] = set()
+        for i in np.nonzero(mask)[0]:
```
marcuscollins, line 394: There has to be a way to vectorize this.

marcuscollins (changes requested):
Okay, I've hit 52 comments, I think it is time to pass this back to you. Please have a look at the comments and work them into the entire PR, even the files I haven't looked at yet (I've looked at 11 files so far).

One meta theme: this could be broken into smaller PRs that would be easier to review. Think about the parts that could change independently. That lowers the cognitive load for reviewers and that helps improve code overall.

Main themes:
1. Avoid for loops where you can use `numpy` vector operations. You use `for` loops _a lot_ where there are opportunities to speed up your code _and_ make it more readable. In Python, numpy (or torch, or even Jax) operations are key to efficient code.

2. There are some opportunities to further factor out functions, including some that you've inlined, to reduce the code volume and make it more testable. 

3. bare strings are an opportunity for mistakes. I strongly suggest using constants for any key where you have default behavior if the key is missing, since if you mistype it anywhere, it will silently change the behavior of the code rather than raise an error. 

4. Even "private" methods need good docstrings. Corollary: look for opportunities to make private methods public, and re-use/merge with other methods wherever possible. Private methods give you control and they feel safe because in principle no one will modify them. But they lead to divergences in behavior across a module that is ultimately more of a headache. Instead, create good tests for your methods to constrain/indicate their behavior.

manzuoni-astera (changes requested):
The direct-to-`refined.cif` path is not ready to replace patching at current head `f5a0b9e`.

1. `guidance_script_utils.py` passes occupancy-suffixed protein names directly to `fetch_rcsb_cif()`. Fetch failure is swallowed, leaving a typeless entity stub and defeating the direct Tortoize-use guarantee.
2. `input_prep.py` sorts chains for its cache key but uses original chain order for entity assignment, so different requests can collide and return the wrong cached output.
3. Residue remapping ignores insertion codes, and label-only `_struct_conn` identifiers are mixed with auth-based maps; both can corrupt numbering/connectivity.
4. Rewriting `refined.cif` in place should be replaced with a same-directory temporary file plus atomic replacement.
5. `scripts/eval/EVALUATION.md` still documents the deleted patching workflow.

Please address these correctness issues and resolve the outstanding maintainer threads. CPU/type/lint checks pass, but the three GPU jobs remain waiting.

PR #313, opened 2026-07-13
feat(it-opt): per step latent space optimization

smallfishabc:
## What

Integrates **inference-time latent optimization (IT-opt)** into Sampleworks as a first-class guidance type, ported from the `it_opt/` reference forks (Protenix/Boltz/AF3).

Treats a frozen structure predictor as a differentiable sampler and optimizes its cached post-trunk latents — the single representation `s` and the pair representation `z` — directly against an experimental density reward on the *denoised* structure. No model weights are trained. Supersedes this PR's original scaffold: the `LatentAdaptedWrapper` affine `k*s + b` transform was never wired in, and was replaced by the direct-leaf approach below.

## Design

- **Direct-leaf optimization.** The `LatentOptimization` scaler extracts `(s, z)` once, clones them into optimizable leaves, and runs one Adam over them across the diffusion trajectory. Each step is one differentiable `denoise → x̂₀`, then `loss = reward(x̂₀) + anchor + bond_geometry`, backward, per-latent grad-clip, `adam.step()`; the coordinate state is detached, so the gradient is a greedy per-step `∂reward/∂latent`, not backprop-through-sampling.
- **Latent-only (v1).** Coordinate-space guidance is disabled (the attached step-scaler returns a zero coordinate direction) — all steering comes from the evolving latents.
- **Regularized.** `LatentAnchor` (mean-squared drift from the baseline latents) tethers the optimization; `BondGeometryReward` (bond-length + steric-clash hinges on the denoised coords) stops `z` from buying density fit with broken geometry.
- **Model-agnostic.** The only model-specific knowledge is which conditioning attribute holds each representation — two strings via `AttrLatentIO` (`s`/`z` for Boltz, `s_trunk`/`z_trunk` for Protenix/RF3). No model packages are imported; a 4th/5th model is one more pair of strings, not new code.
- **Ensemble in the scaler.** Per-member latents with an ensemble-averaged loss, consistent with #330 — not the model wrapper.
- **Protenix gradient flow.** `step()` keeps an injected latent leaf attached automatically (`detach_unless_leaf`: detach cached latents under grad unless `requires_grad=True`), and disables the diffusion shared-vars cache for `z` (else the `z` gradient is silently zero).

## Included

- `src/sampleworks/core/scalers/latent_optimization.py` — `LatentOptimization` scaler + `LatentAnchor`
- `src/sampleworks/core/rewards/geometry.py` — `BondGeometryReward` (bond-length + collision penalties)
- `src/sampleworks/models/latent_adapter.py` — `AttrLatentIO` / `LatentIO` (+ `tests/models/test_latent_adapter.py`, `test_latent_optimization.py`)
- `src/sampleworks/models/protenix/wrapper.py` — `detach_unless_leaf` gradient routing
- `src/sampleworks/utils/` — `GuidanceType.LATENT_OPT`, `add_latent_opt_args` (`--which-latent`, `--learning-rate`, `--outer-steps`, `--anchor-weight`, `--max-grad-norm`, `--bond-length-weight`), and the `LATENT_OPT` dispatch in `_run_guidance`

## Status / next steps

Draft — the loop runs end to end for both latents and is reachable from the CLI (`--guidance-type latent_opt`) and `run_guidance`. Under the density objective `s` is effectively inert while `z` is the lever; unregularized `z` raises the paper's altloc RSCC but drifts off-manifold and adds a few clashes (hence the anchor + bond-geometry terms). Coordinate guidance is the modest, safe baseline it must beat. No mode crosses the paper's 0.8 RSCC bar yet, so efficacy is still open.

> Note: full end-to-end Protenix `step()` runs need the model + weights (pixi/pod env). The scaler logic and gradient routing are verified in isolation, but a real single-latent run is what proves efficacy before marking ready for review.

🤖 Generated with Claude Code

## Summary by CodeRabbit

* **New Features**
  * Added `latent_opt` inference-time latent optimization with configurable latent selection, scheduling, anchoring, gradient clipping, and optional bond-geometry penalties.
  * Added bond-length and steric-clash regularizers for improved structural plausibility.
  * Added CLI controls and trajectory output support.
* **Bug Fixes**
  * Preserved gradient flow for optimizable cached latent representations during inference.
* **Tests**
  * Added coverage for latent adapters, optimization components, and CLI parsing.
* **Documentation**
  * Added IT-opt design and testing guides, plus a codebase walkthrough.
* **Chores**
  * Updated ignore rules for additional local tooling artifacts.


manzuoni-astera (commented):
Thanks for keeping this as a narrowly scoped step-1 scaffold. Treating the optimizer, sampler feature hook, Protenix detach handling, Boltz2 cache recomputation, and real experimental transform as intentional follow-up work, the basic read/transform/write seam looks reasonable.

Before this is marked ready, I think the scaffold contract needs a few corrections:

1. `LatentAdaptedWrapper` is not currently drop-in for the actual guidance entry point. `_run_guidance` dispatches preprocessing from `model_wrapper.__class__.__name__` and will reject `LatentAdaptedWrapper` as unknown (`guidance_script_utils.py:444-486`). Pure guidance also reads `model_wrapper.model` directly at line 541, while diagnostics and MSA reporting use other concrete-wrapper attributes. The one-line factory wrapping shown in the diagram would therefore fail before sampling. Please either make orchestration decorator-aware or deliberately proxy/unwrap the required model identity and attributes, with a pipeline-level test.

2. A wrong `single_attr` silently becomes a no-op (`latent_adapter.py:85-88, 172-174`), and the test currently enshrines that behavior. Because this string is required configuration, I would prefer a clear error, or an explicit opt-in `allow_missing` mode, so field drift cannot masquerade as successful adaptation.

3. The mock `step()` ignores conditioning, so the tests prove tensor replacement but not that rewritten conditioning affects the delegated step. A conditioning-sensitive mock should cover identity equivalence and non-identity propagation.

CI unit tests and typechecks pass across all environments, but lint currently fails with 39 introduced Ruff errors. GPU jobs are awaiting environment approval. With those scaffold-level issues addressed, the deferred model-specific work can reasonably remain in follow-ups.

docs/CODEBASE_GUIDE.md
```diff
@@ -0,0 +1,442 @@
...
+        ├─ {DataSpace,NoiseSpace,No}DPSScaler     → per-step guidance rule
+        └─ {PureGuidance | FKSteering}.sample(structure, model, sampler, scaler, reward)
+               │
+               └─ FOR each diffusion step:
+                    sampler.step(coords, model, context, scaler)
+                        ├─ model.step(noisy, t)          → x̂₀  (denoised prediction)
+                        ├─ reconciler.align(x̂₀, ref)     → into experimental frame
+                        └─ scaler.scale(x̂₀, context)     → reward(x̂₀), ∇reward → guidance
+        ▼
+  save_everything()  → refined.cif, trajectory/*.cif, losses.txt, job_metadata.json
```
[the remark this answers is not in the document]
marcuscollins: @smallfishabc it would make sense to update this as mentioned.

docs/CODEBASE_GUIDE.md
```diff
@@ -0,0 +1,445 @@
...
+
+### 2.1 The data objects that flow between them
+
+Understanding these five dataclasses is enough to read any part of the sampling loop.
+
+- **`GenerativeModelInput`** ([models/protocol.py:25](../src/sampleworks/models/protocol.py))
+  — `{x_init, conditioning}`. Produced by `featurize()`. `x_init` is the batched starting
+  coordinates (reference or prior noise, shape `(ensemble, atoms, 3)`); `conditioning` is a
+  model-specific object holding the cached trunk/pairformer output.
+  > **Update (#330).** `x_init` has since been removed — `GenerativeModelInput` now carries only
```
marcuscollins: Probably just remove any mention of x_init from the GenerativeModelInput section.

docs/CODEBASE_GUIDE.md
```diff
@@ -0,0 +1,445 @@
...
+
+- **`GuidanceOutput`** ([core/scalers/protocol.py:40](../src/sampleworks/core/scalers/protocol.py))
+  — `{structure, final_state, trajectory, losses, metadata}`. What a trajectory scaler
+  returns; `metadata` carries `trajectory_denoised` and `model_atom_array`.
+
+---
+
+## 3. The spine: `run_guidance()` step by step
+
+Both CLI paths converge here. File:
```
marcuscollins: I don't know why it says this. The analysis code _shouldn't_ ever call the guidance script utilities. What are the two CLI paths that converge here?

docs/CODEBASE_GUIDE.md
```diff
@@ -0,0 +1,445 @@
...
+| `ProtenixWrapper` | [protenix/wrapper.py](../src/sampleworks/models/protenix/wrapper.py) | AF3 reimpl. `structure_processing.py` builds the Protenix JSON (entities, modifications, covalent bonds). Optional diffusion-shared-vars cache (`pair_z, p_lm, c_l`). |
+| `RF3Wrapper` | [rf3/wrapper.py](../src/sampleworks/models/rf3/wrapper.py) | Baker AF3 replica. Uses an inference engine + trunk-with-recycling generator. Optional chiral-feature tracking/disabling (writes `chiral_grad_stats.json`). |
+| `LatentAdaptedWrapper` *(archived — not in the build)* | [latent_adapter/archived_injector_family.md](latent_adapter/archived_injector_family.md) | An alternative injector/decorator approach (affine/delta transform at `featurize`). Never wired in; superseded by the direct-leaf `LatentOptimization` scaler and archived. IT-opt now reads/writes latents via `AttrLatentIO` in [latent_adapter.py](../src/sampleworks/models/latent_adapter.py). |
```
marcuscollins: If it is unused, we should just remove it from this PR. You can keep it in a separate branch if you want to keep it for some future work.

docs/IT_OPT_DESIGN.md
```diff
@@ -0,0 +1,176 @@
...
+**v1 is latent-only.** Coordinate-space guidance is *not* applied — the attached step-scaler
+returns a zero coordinate direction (§4, `_GradEnablingScaler`), so the only steering comes from the
+evolving latents.
+
+## 2. The as-built loop
+
+```
+extract (s, z) from the trunk once  ->  clone into optimizable leaves (requires_grad=True)
+for each outer round (fresh prior noise each round):
+    optimizer = Adam([s, z])                      # ONE persistent Adam per round
```
marcuscollins, line 35: How do you define a "round" here? Is it a complete denoising trajectory?

docs/IT_OPT_DESIGN.md
```diff
@@ -0,0 +1,176 @@
...
+    for each diffusion step:
+        x̂₀   = differentiable denoise(x_t, s, z)  # ONE forward, run under autograd
+        loss  = reward(x̂₀) + anchor(s, z) + bond_geometry(x̂₀)
```
marcuscollins, line 38: is the anchor some kind of regularizer to keep s and z from drifting too far? Is there a knob to control how strong this?

docs/IT_OPT_DESIGN.md
```diff
@@ -0,0 +1,176 @@
...
+
+Private helpers, for orientation: `_leaf_latents` (promote s/z to leaves), `_optimize_one_round`
+(one outer round), `_latent_adam_step` (score → backward → clip → step), `_sample_with_frozen_latents`
+(final clean pass).
+
+### Constructor contract
+
+```python
+LatentOptimization(
+    ensemble_size=1, num_steps=200, guidance_t_start=0.0, *,
```
marcuscollins, line 69: Are these settings primarily for RF3/Boltz/Protenix? Might be good to start looking into how/whether this will work for Protpardelle.

docs/IT_OPT_DESIGN.md
```diff
@@ -0,0 +1,176 @@
...
+    outer_steps=1, learning_rate=0.05, max_grad_norm=1.0,
+    optimize_single=True, optimize_pair=True,
+    anchor_weight_single=0.0, anchor_weight_pair=0.0,
+    bond_length_weight=0.0, single_attr="s", pair_attr="z",
+)
+```
+
+- `ensemble_size` structures are sampled in parallel and **share** the latents (per-member latents
+  are a documented follow-up).
+- `guidance_t_start` is a fraction in `[0,1]`; stored as `guidance_start = int(guidance_t_start*num_steps)`.
```
marcuscollins, line 79: This is something to watch out for; in the rest of the package this is given in number of steps. We should probably change that (there's an old TODO to change it) but we haven't yet.

docs/IT_OPT_DESIGN.md
```diff
@@ -0,0 +1,176 @@
...
+  Steps before it are plain frozen-latent diffusion.
```
marcuscollins, line 80: are --> is

docs/IT_OPT_DESIGN.md
```diff
@@ -0,0 +1,176 @@
...
+  **not** emit the siblings' `"trajectory_denoised"` — treat scaler-specific keys as optional.
+
+## 4. The gradient gate
+
+The load-bearing mechanism. `AF3EDMSampler.step` reads `getattr(scaler, "requires_gradients", False)`;
+if true it runs the denoiser under `torch.set_grad_enabled(True)` and returns a `denoised` (`x̂₀`)
+that still carries a graph back to the latent leaves:
+
+```python
+class _GradEnablingScaler:
```
marcuscollins, line 96: Is this a new class you've created? Where does it live?

docs/IT_OPT_DESIGN.md
```diff
@@ -0,0 +1,176 @@
...
+  Steps before it are plain frozen-latent diffusion.
+- `outer_steps` resample rounds, fresh prior noise each. **Constructor default `1`; CLI default `2`.**
+- `learning_rate` Adam LR; one persistent Adam is built **once per round**.
+- `max_grad_norm` clips `s` and `z` **independently** (see §6 — a joint clip starves `s`).
+- `single_attr`/`pair_attr` default to Boltz names; the wiring always passes model-resolved names.
+- The `metadata` on the returned `GuidanceOutput` carries `"optimization_losses"` (per-round,
+  per-step data losses) and `"latent_drift"` (per-round relative L2 drift of each latent). It does
+  **not** emit the siblings' `"trajectory_denoised"` — treat scaler-specific keys as optional.
+
+## 4. The gradient gate
```
marcuscollins, line 89: This section is confusing.

docs/IT_OPT_DESIGN.md
```diff
@@ -0,0 +1,176 @@
...
+The sampler does **not** re-detach `state`/`denoised` — the scaler loop detaches between iterations.
+
+The denoised `x̂₀` is frame-aligned to the experimental reference by the sampler (via the
+`AtomReconciler` + `alignment_reference`) *before* the reward sees it, so reward functions can assume
+pre-aligned coordinates.
+
+## 5. Per-model gradient readiness
+
+The only model-specific concern is whether the cached latents can receive a gradient. The optimized
+attributes are `s`/`z` for Boltz and `s_trunk`/`z_trunk` for Protenix/RF3
```
marcuscollins, line 113: Definitely start thinking about how you can apply this to Protpardelle

docs/IT_OPT_DESIGN.md
```diff
@@ -0,0 +1,176 @@
...
+(`DEFAULT_SINGLE_REP_ATTR` / `DEFAULT_PAIR_REP_ATTR` in `latent_adapter.py`).
+
+| Model | Status | Note |
+|---|---|---|
+| **Boltz1** | Clean | Trunk latents reach the diffusion module directly. |
+| **RF3** | Clean | Same. |
+| **Protenix** | Works, with two conditions | (a) `step()` keeps an injected leaf attached automatically via the `detach_unless_leaf` helper — a cached latent detaches under grad (avoids double-backward), but a leaf with `requires_grad=True` is kept attached. No manual edit needed. (b) For `z`, the diffusion shared-vars cache must be **off** (`pair_z`/`p_lm`/`c_l` are z-derived; a stale cache silently zeroes / corrupts the `z` gradient). `_run_guidance` disables it for `LATENT_OPT`. |
+| **Boltz2** | Not yet | `z` (and part of `s`) are baked into a cached `diffusion_conditioning` at featurize time; needs the conditioning recomputed from the live latents. |
```
marcuscollins, line 121: so does ITO not work for Boltz2 yet?

docs/IT_OPT_DESIGN.md
```diff
@@ -0,0 +1,176 @@
...
+All four models' conditioning is a `@dataclass(frozen=True, slots=True)`, swapped via
+`dataclasses.replace` — hence `AttrLatentIO`.
+
+## 6. Design choices and invariants
+
+**Kept from the reference (the algorithm):** `s`/`z` as the optimized variables with `s_inputs`
+frozen; one Adam step per diffusion step on the denoised `x̂₀`; latents persist across steps and
+rounds; an outer resample loop; an anchor to the trunk baseline; a final clean sampling pass.
+
+**Fixed from the reference:**
```
marcuscollins, line 132: I don't think you ever list what the reference code is. Should add a GitHub link and a link to the paper too.

docs/IT_OPT_DESIGN.md
```diff
@@ -0,0 +1,176 @@
...
+- Bounded geometry hinges instead of the reference's exploding `exp(relu(...))` clash term; no
+  hardcoded save-step.
+
+**Deliberate deviations:** the objective is a pluggable `RewardFunctionProtocol` (v1:
+`RealSpaceRewardFunction`, density fit) rather than backbone-RMSD; the anchor is mean-squared rather
+than the reference's Frobenius norm (shape-agnostic, so `w_s`/`w_z` are comparable); latents are
+shared across the ensemble in v1.
+
+**Invariants worth stating:**
+- **Optimization is not sampling.** An optimized `(s, z)` is a point estimate; reporting it as a
```
marcuscollins, line 151: This might be a future interesting direction to go in.

docs/IT_OPT_TESTING.md
```diff
@@ -0,0 +1,206 @@
...
+capture — the clustering silhouette stays flat and ~80% of ensembles still collapse to one
+conformation, consistent with the paper's thesis. `coord_guidance` is inert on every metric here. The
+default bond weight `5e-5` is the smallest that restores mean clash to baseline while keeping the full
+density gain and the diversity; `1e-3` over-constrains and erodes diversity. Default
+`bond_length_weight` is `5e-5` (`LatentOptimization.__init__` + `--bond-length-weight`); `0` disables
+the penalty.
+
+## 6. Open problems
+
+1. **`coord_guidance` is inert vs. the paper.** The paper reports coordinate guidance taking RSCC ≥
```
marcuscollins: Which paper?

docs/IT_OPT_TESTING.md
```diff
@@ -0,0 +1,206 @@
...
+   0.8 from 45.4% to 96.0%; here it is indistinguishable from baseline on every metric, and it is the
```
marcuscollins: This seems like an overstatement of the sampleworks paper on The Stacks. Is this accurate??

src/sampleworks/core/rewards/geometry.py
```diff
@@ -0,0 +1,215 @@
...
+    """Covalent radius of an element (Å), or 0 for an unknown/blank symbol.
+
+    Model atom arrays are almost always clean single-symbol elements, but density inputs sometimes
+    carry a placeholder element ('?'); we return 0 rather than raise so one odd atom can't abort the
+    whole run.
+    """
+    try:
+        return float(gemmi.Element(element).covalent_r)
+    except Exception:  # noqa: BLE001 -- a bad element symbol just gets a zero radius
+        return 0.0
```
marcuscollins: zero seems like an odd choice. Will this accidentally penalize atoms for having non-zero bond lengths? Or does zero mean that atom won't record a collision? I guess either way, add a comment here about what zero will mean for the calculations.

src/sampleworks/core/rewards/geometry.py
```diff
@@ -0,0 +1,215 @@
...
+class BondGeometryReward:
+    """Penalize distorted bonds and steric clashes in the denoised structure.
+
+    The topology (which atoms are bonded, their ideal bond lengths, the per-atom-pair collision
+    distances, and which pairs to score for clashes) is read once from ``atom_array`` at
+    construction, so each step is just cheap tensor ops on the coords. ``atom_array`` MUST be the
+    atom set whose ordering matches the coordinate tensor the reward sees -- the *model* atom array,
+    not the deposited structure -- or the bond indices will not line up with the coordinates.
+
+    Despite the ``Reward`` name (the module's convention), this is a penalty: it is added to,
```
marcuscollins, line 71: This is important: we might want to use it in combination with other rewards in the future, so we probably do want to formulate it as a reward (to be _maximized_).

src/sampleworks/core/rewards/geometry.py
```diff
@@ -0,0 +1,215 @@
...
+        # Collect each bond once as a sorted (low, high) index pair, so it is not listed as both
+        # i-j and j-i.
+        pairs = set()
+        for atom_i, partners in enumerate(partners_per_atom):
+            for atom_j in partners:
+                if atom_j == -1:  # padding entry, not a real partner
+                    continue
+                bond = (min(atom_i, int(atom_j)), max(atom_i, int(atom_j)))
+                pairs.add(bond)
+        pair_array = np.array(sorted(pairs)) if pairs else np.empty((0, 2), dtype=int)
```
marcuscollins: Can you look into whether AtomArray has a method for computing all the bond pairs as an array? I thought once the AtomArray had bond information there was something other than .get_all_bonds that would provide them in a more useful format. It would be faster than this double loop.

src/sampleworks/core/rewards/geometry.py
```diff
@@ -0,0 +1,215 @@
...
+        Entry (i, j) is the sum of atom i's and atom j's covalent radii; two non-bonded atoms closer
+        than this (plus ``clash_padding``) are overlapping.
+        """
+        radii = np.array([_covalent_radius(atom_array[i].element) for i in range(len(atom_array))])
+        r = torch.tensor(radii, dtype=torch.float32, device=device)
+
+        # radius_i + radius_j for every pair: a column plus a row vector broadcasts to [n, n].
+        radius_column = r.unsqueeze(1)  # [n, 1]
+        radius_row = r.unsqueeze(0)  # [1, n]
+        return radius_column + radius_row
```
marcuscollins, line 158: You're calculating some of these twice--any bonded pair is already calculated by the previous method. Perhaps you can just look up the values needed there by indexing into the matrix you have here?

src/sampleworks/core/rewards/geometry.py
```diff
@@ -0,0 +1,215 @@
...
+        if self._bond_lengths.numel() == 0:
+            return coords.new_zeros(())
+
+        # Current length of every bond, for every ensemble member: [ensemble, n_bonds].
+        pos_a = coords[:, self._bond_atom_a]
+        pos_b = coords[:, self._bond_atom_b]
+        lengths = (pos_a - pos_b).norm(dim=-1)
+
+        deviation = (lengths - self._bond_lengths).abs()  # distance from the ideal length
+        excess = (deviation - self.bond_tolerance).clamp(min=0)  # part beyond the tolerance
```
marcuscollins: Do you only want to penalize positive deviations? If so, why? This seems appropriate for clashes, but not for bond lengths.

src/sampleworks/core/rewards/geometry.py
```diff
@@ -0,0 +1,215 @@
...
+        # pinned torch (2.7): its gradient is finite and matches the broadcast form even for
+        # coincident atoms, so the historical cdist zero-distance NaN does not apply here.
+        distances = torch.cdist(coords, coords)
+
+        # How far each pair is inside its allowed distance (positive = overlapping).
+        min_distance = self._collision_distances + self.clash_padding
+        overlap_per_member = (min_distance - distances).clamp(min=0)
+
+        # Worst overlap across the ensemble for each pair, then drop self- and bonded pairs.
+        worst_overlap = overlap_per_member.max(dim=0).values  # [n, n]
```
marcuscollins: When this back propagates, presumably it will only affect the worst clashes--is that efficient? Why not apply to all of members of the ensemble?

src/sampleworks/core/rewards/geometry.py
```diff
@@ -0,0 +1,215 @@
...
+        scored_overlap = worst_overlap * self._scored_pairs  # zero out the unscored pairs
+        return scored_overlap.sum()
+
+    def __call__(self, coords: Tensor) -> Tensor:
+        """Weighted sum of the bond-length and collision penalties, added to the loss."""
+        return self.weight * (self.bond_length_loss(coords) + self.collision_loss(coords))
```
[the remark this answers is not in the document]
marcuscollins: In this one case, I wouldn't worry about it. It's hard to imagine us ever not doing both of these if we do one of them.

src/sampleworks/models/protenix/wrapper.py
```diff
@@ -637,16 +637,40 @@ def step(
...
+        # Safe for the guidance path too: after featurize (run under no_grad) these are
+        # requires_grad=False constants with no graph, so "not detaching" is a no-op there;
+        # it only matters when an optimizable latent leaf has been injected as s_trunk/z_trunk.
         grad_needed = torch.is_grad_enabled()
         s_inputs = cond.s_inputs.detach() if grad_needed else cond.s_inputs
-        s_trunk = cond.s_trunk.detach() if grad_needed else cond.s_trunk
-        z_trunk = cond.z_trunk.detach() if grad_needed else cond.z_trunk
+        s_trunk = cond.s_trunk  # keep attached: latent gradient must reach s_trunk
+        z_trunk = cond.z_trunk  # keep attached: latent gradient must reach z_trunk
+        # pair_z / p_lm / c_l are z-derived caches. When optimizing z_trunk, featurize with
```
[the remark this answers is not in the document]
marcuscollins: @smallfishabc this does make me think we need to validate that this change doesn't somehow break coordinate guidance. You should check that you get similar results for at least one of the models and guidance weights in the paper (say, protenix with weight 0.1, etc...), over our test set (the "initial 40")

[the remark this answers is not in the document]
marcuscollins: If we do not get very similar results to what's in the paper, that would be a problem.

It would be good for us to actually devise some "integration tests" like this to verify that big changes such as this PR don't cause regressions to other code paths. Please add an issue to add integration tests that check results against the results in the paper. This can be assigned to Michael and marked engineering/P0. Probably Karson will need to consult on it.

src/sampleworks/models/latent_adapter.py
```diff
@@ -0,0 +1,165 @@
...
+- :class:`AttrLatentIO` -- reads/writes ``s`` and ``z`` by attribute name. This is the piece the
+  IT-opt scaler actually uses: it reads the baseline latents once, holds its own optimizable copies,
+  and writes them back each step.
+- :class:`LatentIO` -- the protocol (read/write contract) that :class:`AttrLatentIO` satisfies.
+
+An earlier alternative -- training a small transform (``k*s + b`` affine / additive delta) at the
+``featurize`` boundary via a ``LatentAdaptedWrapper`` decorator -- was never wired in and was
+superseded by the direct-leaf scaler. It is archived, with the reasoning, in
+``docs/latent_adapter/archived_injector_family.md`` (revive it as the reference's structured
+``update_mode`` family -- low-rank / per-residue -- if regularized perturbations are ever needed).
```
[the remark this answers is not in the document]
marcuscollins: @smallfishabc Remove dead comments like this--it will just confuse readers.

src/sampleworks/models/latent_adapter.py
```diff
@@ -0,0 +1,165 @@
...
+
+Design goal
+-----------
+- **Minimal, model-agnostic.** The only model-specific knowledge is *which attribute of the
+  conditioning holds each representation* -- ``"s"``/``"z"`` for Boltz, ``"s_trunk"``/``"z_trunk"``
```
marcuscollins, line 20: @smallfishabc add an issue to work out how we will do ITO with other feature schemes.  Use protpardelle as an example. We will probably have to rework exactly how the reward function works, since it will take different features. You will need to abstract out which features from the "trunk" are used.

src/sampleworks/models/latent_adapter.py
```diff
@@ -0,0 +1,165 @@
...
+    "boltz2": "z",
+    "protenix": "z_trunk",
+    "rf3": "z_trunk",
+}
+
+
+# "conditioning" throughout this file is the model's conditioning object: a (frozen) dataclass
+# carrying the latents ``s``/``z`` (and other cached state) as named attributes. These two helpers
+# just get, or replace, one such attribute on it.
+def _read_attr(conditioning, attr: str) -> Tensor | None:
```
marcuscollins: I would remove this and the next function. We should assume that conditioning is not None (all the models will fail if it is None) and it can be assumed to be a dataclass. This ends up just being indirection that doesn't guard against any real scenario, hides situations that _should_ raise errors, and makes the code harder to follow.

src/sampleworks/models/latent_adapter.py
```diff
@@ -0,0 +1,165 @@
...
+        When ``None`` (default), the pair accessors are no-ops -- single-rep-only
+        behaviour, matching the original scaffold.
+    """
+
+    def __init__(self, single_attr: str, pair_attr: str | None = None):
+        self.single_attr = single_attr
+        self.pair_attr = pair_attr
+
+    def read_single(self, conditioning) -> Tensor | None:
+        return _read_attr(conditioning, self.single_attr)
```
marcuscollins: Again, _read_attr and _set_attr aren't really needed. I would just use `getattr(conditioning, self.single_attr)` here. If that raises an error because conditioning is None, that's fine.

src/sampleworks/models/latent_adapter.py
```diff
@@ -0,0 +1,165 @@
...
+# ============================ protocol (template / interface) ============================
+# A Protocol is just a checklist of methods a class must have -- think a C header, or an abstract
+# base class, but *structural*: any class that already has these methods counts as a LatentIO
+# without inheriting from anything. This definition does not run; it only documents the contract and
+# lets the ty checker flag a mismatch. @runtime_checkable also lets isinstance(obj, LatentIO) work,
+# which the unit tests use. The real implementation is above: AttrLatentIO is the LatentIO.
+
+
+@runtime_checkable
+class LatentIO(Protocol):
```
marcuscollins, line 120: Will you ever have other versions of LatentIO besides AttrLatentIO? If not, then I don't think the Protocol is needed.

src/sampleworks/utils/guidance_constants.py
marcuscollins, line 47: @smallfishabc add an issue, assign to Karson, to determine whether or not we actually need both TrajectoryScalers and GuidanceType. They seem to be the same thing.

src/sampleworks/utils/guidance_script_utils.py
```diff
@@ -500,7 +505,20 @@ def _run_guidance(args: GuidanceConfig, guidance_type: str, model_wrapper, devic
     if "Protenix" in wrapper_class_name:
         from sampleworks.models.protenix.wrapper import annotate_structure_for_protenix
 
-        structure = annotate_structure_for_protenix(structure, recycling_steps=recycling_steps)
+        structure = annotate_structure_for_protenix(
+            structure,
+            recycling_steps=recycling_steps,
+            # ensemble_size removed: #330 dropped it from the wrapper annotate signature; the
```
marcuscollins: This comment is much too long. Try to keep it to just 2-3 lines maximum. You don't need to mention dropping ensemble size at all.

src/sampleworks/utils/guidance_script_utils.py
```diff
@@ -633,6 +651,76 @@ def _run_guidance(args: GuidanceConfig, guidance_type: str, model_wrapper, devic
...
+        from sampleworks.models.latent_adapter import (
+            DEFAULT_PAIR_REP_ATTR,
+            DEFAULT_SINGLE_REP_ATTR,
+        )
+
+        # LatentOptimization expects guidance start as a fraction of the schedule, but the config
+        # carries it as an integer step count, so we convert it here and default to optimizing from
+        # the first step.
+        guidance_t_start = args.guidance_start / num_steps if args.guidance_start > 0 else 0.0
+        which_latent = getattr(args, "which_latent", "pair")  # This is "single", "pair", or "both".
```
marcuscollins: You can make sure only these values are used by using the `choices` kwarg to `add_argument` when you create the argument in `guidance_script_arguments.py`

src/sampleworks/utils/guidance_script_utils.py
```diff
@@ -633,6 +651,76 @@ def _run_guidance(args: GuidanceConfig, guidance_type: str, model_wrapper, devic
...
+        anchor_weight = getattr(args, "anchor_weight", 0.0)
+        bond_length_weight = getattr(args, "bond_length_weight", 5e-5)  # smallest good default
```
marcuscollins: I prefer to set defaults when creating the argument with `add_argument`. This does two things, it prevents you from getting a value if the argument wasn't properly created in the first place, and it makes the defaults more easily visible to an end-user. If these attributes don't exist or don't have a value by this point, it _should_ raise an error here.

This goes for _anywhere_ you're fetching argument values from GuidanceConfig or from an argument parser.

src/sampleworks/utils/guidance_script_utils.py
```diff
@@ -633,6 +651,76 @@ def _run_guidance(args: GuidanceConfig, guidance_type: str, model_wrapper, devic
...
+            raise ValueError(
+                "Latent optimization has no latent-attribute names registered for model "
+                f"{model_key!r}."
+            ) from e
+
+        guidance = LatentOptimization(
+            ensemble_size=args.ensemble_size,
+            num_steps=num_steps,
+            guidance_t_start=guidance_t_start,
+            outer_steps=getattr(args, "outer_steps", 2),
```
marcuscollins: again, don't set defaults here, this goes for lines 701-703

src/sampleworks/utils/guidance_script_utils.py
```diff
@@ -633,6 +651,76 @@ def _run_guidance(args: GuidanceConfig, guidance_type: str, model_wrapper, devic
...
+            learning_rate=getattr(args, "learning_rate", 0.05),
+            max_grad_norm=getattr(args, "max_grad_norm", 1.0),
+            optimize_single=which_latent in ("single", "both"),
```
marcuscollins, line 698: And again here, instead of using `which_latent` I would suggest using `args.which_latent in ("single", "both")`

src/sampleworks/core/scalers/latent_optimization.py
```diff
@@ -0,0 +1,554 @@
+"""Inference-time latent-space optimization (IT-opt).
+
+A Sampleworks port of the reference ``run_it_optimization``
+(``it_opt/protenix/it_optimization_manager.py:288``). It treats the frozen
```
marcuscollins: provide a GitHub link to the repo

src/sampleworks/core/scalers/latent_optimization.py
```diff
@@ -0,0 +1,554 @@
...
+#   model: FlowModelWrapper        model                    should be a FlowModelWrapper
+#   weights: Sequence[float]       weights                  should be a list/tuple of floats
+#   latents: Sequence[Tensor]      latents                  should be a list/tuple of Tensors
+#   losses: list[float]            losses                   should be a list of floats
+#   denoised: Tensor | None        denoised                 should be a Tensor, or None
+#   f(...) -> GuidanceOutput       f(...)                   f should return a GuidanceOutput
+# -------------------------------------------------------------------------------
+
+
+class _GradEnablingScaler:
```
marcuscollins: Is this basically a dummy scaler in place of something like NoiseSpaceDPSScaler? You can double check it is the same but there is already  `sampleworks.core.scalers.step_scalers.NoScalingScaler` that I think is identical. Please just use that one instead.

src/sampleworks/core/scalers/latent_optimization.py
```diff
@@ -0,0 +1,554 @@
...
+
+    def __init__(self, weights: Sequence[float]):
+        self.weights = list(weights)
+
+    def __call__(self, latents: Sequence[Tensor], baselines: Sequence[Tensor]) -> Tensor:
+        terms = [
+            w * torch.mean((lat - base) ** 2)
+            for w, lat, base in zip(self.weights, latents, baselines)
+        ]
+        return torch.stack(terms).sum()
```
marcuscollins, line 126: Is this basically regularization in the form of a reward? I think we will want to collect these kinds of things under the rewards code, so they are re-usable. I would move this to core.rewards just like the geometry.py file you created, and then import it here. That will help us when we go to redefine the rewards interface to make it easy to combine rewards in different ways.

src/sampleworks/core/scalers/latent_optimization.py
```diff
@@ -0,0 +1,554 @@
...
+        pair_attr: str = "z",
+    ):
+        """Initialize the latent-optimization trajectory scaler.
+
+        Parameters
+        ----------
+        ensemble_size
+            Number of structures sampled in parallel (they share the latents).
+        num_steps
+            Number of diffusion steps per pass.
```
marcuscollins: Is a "pass" the same as the term "round" you use elsewhere? It helps to keep terminology consistent throughout the code base, so if they are the same, I would change them to use one term (I would favor "pass")

src/sampleworks/core/scalers/latent_optimization.py
```diff
@@ -0,0 +1,554 @@
...
+        processed = process_structure_to_trajectory_input(
+            structure=structure,
+            coords_from_prior=coords,
+            features=features,
+            ensemble_size=self.ensemble_size,
+        )
+        reconciler = processed.reconciler.to(coords.device)
+        reward_inputs = processed.to_reward_inputs(device=coords.device)
+        schedule = sampler.compute_schedule(num_steps=self.num_steps)
+        grad_enabler = _GradEnablingScaler()
```
marcuscollins, line 260: As mentioned above, I think what you want to do here is use the NoScalingScaler and pass that as your step scaler.

src/sampleworks/core/scalers/latent_optimization.py
```diff
@@ -0,0 +1,554 @@
...
+
+        # --- optional coordinate-space geometry penalty -------------------------
+        # BondGeometryReward penalizes stretched bonds and steric clashes in the denoised structure,
```
marcuscollins, line 263: The first sentence is enough here.

src/sampleworks/core/scalers/latent_optimization.py
```diff
@@ -0,0 +1,554 @@
...
+        # curbing the overshoot where an aggressive latent update trades valid geometry for density
+        # fit. Its topology is read from the model atom array -- the atom set whose ordering matches
+        # the denoised coords and reward inputs -- so the bond indices line up. It stays None
+        # (disabled) unless a positive bond_length_weight was configured.
+        bond_geometry = None
+        if self.bond_length_weight > 0:
+            # Prefer the model atom array; fall back to the input atom array if it is absent.
+            geometry_atom_array = processed.model_atom_array or processed.atom_array
```
marcuscollins, line 267: Does that happen? If so, when/why? I'd be concerned that they don't have the same dimensions.

src/sampleworks/core/scalers/latent_optimization.py
```diff
@@ -0,0 +1,554 @@
...
+            bond_geometry = BondGeometryReward(
+                geometry_atom_array, self.bond_length_weight, coords.device
+            )
+
+        # --- optimize the latents (outer resample × inner diffusion steps) ------
+        # These two lists are diagnostics only -- reported in the returned metadata, never read by
+        # the loop. One entry is appended per round.
+        optimization_losses = []  # per round: the per-step data losses
+        latent_drift = []  # per round: how far each latent moved from its baseline (see below)
+        for outer in range(self.outer_steps):
```
marcuscollins: use tqdm so there's a progress bar. Do we want to have an option to optimize over only part of the trajectory? (Some kind of partial diffusion like approach)? If so, let's make an issue. E.g., in PureGuidance, we have `for i in tqdm(range(self.starting_step, self.num_steps)):`

src/sampleworks/models/latent_adapter.py
marcuscollins: Okay, reading how these are used, I see that you need a function. I'll make a comment around L350 there about this.

src/sampleworks/core/scalers/latent_optimization.py
```diff
@@ -0,0 +1,554 @@
...
+        # One row per optimizable latent: (optimize it?, how to read it off the conditioning, how to
+        # write it back, its anchor weight). We handle the single representation, then the pair.
+        specs = (
+            (self.optimize_single, io.read_single, io.write_single, self.anchor_weight_single),
+            (self.optimize_pair, io.read_pair, io.write_pair, self.anchor_weight_pair),
+        )
+        for enabled, read, write, anchor_weight in specs:
+            if not enabled:
+                continue
+            baseline = read(conditioning)
```
marcuscollins: Instead of this, I would just fetch the attribute you need here. Instead of passing the method, you can just pass, say, io.single_attr_name (I forget the exact attribute name) which will be more clear here what you're doing, you won't need as large tuples in `specs` (it would just be, e.g., `(self.optimize_single, io.single_attr_name, self.anchor_weight_single)`. Then you can use `getattr` and `setattr` here and below.

src/sampleworks/core/scalers/latent_optimization.py
```diff
@@ -0,0 +1,554 @@
...
+            if baseline is None:
+                continue
```
marcuscollins: when would this happen? I think only if conditioning is None or doesn't have the required attribute, right? In that case you _should_ be raising an error, rather than silently executing but doing nothing. Claude and other AI code agents favor code that runs no matter what. That's a defensive pattern used in customer facing applications where you want, say, a website to stay up and running no matter what. Here, we would _want_ the code to _crash_, not fail silently and make us think the method isn't working.

src/sampleworks/core/scalers/latent_optimization.py
```diff
@@ -0,0 +1,554 @@
...
+            conditioning = io.write_pair(conditioning, detached.pop(0))
+        frozen_features = GenerativeModelInput(conditioning=conditioning)  # #330: conditioning-only
+
+        coords = torch.as_tensor(
+            model.initialize_from_prior(batch_size=self.ensemble_size, features=frozen_features)
+        )
+        trajectory: list[Tensor] = []
+        losses: list[float | None] = []
+        steps = tqdm(range(self.num_steps), "IT-opt final sampling")
+        for i in steps:
```
marcuscollins, line 541: this is basically the same loop as in PureGuidance, right? Maybe make an issue to see if we can factor it out.

marcuscollins (changes requested):
Lots of little things to do, I haven't read through absolutely everything but let's start by addressing these comments, and then I'll do another pass. This is great work, just need to make it a little easier to understand and simplify a bit.

src/sampleworks/core/rewards/geometry.py
smallfishabc: I think the .abs() penalized both conditions.

src/sampleworks/models/protenix/wrapper.py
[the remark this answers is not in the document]
smallfishabc: I will do a detailed test later

src/sampleworks/models/latent_adapter.py
smallfishabc: I will leave it here just in case we need it for the protpardelle

src/sampleworks/core/scalers/latent_optimization.py
smallfishabc: I have proposed an issue to solve this later

docs/IT_OPT_DESIGN.md
smallfishabc: I introduced a more structured explanation. Basically a wrong is 200 steps.

docs/IT_OPT_DESIGN.md
smallfishabc: This will be discussed late. Since the Protpardelie has a different latent space representation.

src/sampleworks/core/scalers/latent_optimization.py
```diff
@@ -0,0 +1,566 @@
...
+    ``NoScalingScaler`` therefore leaves :attr:`SamplerStepOutput.denoised` detached, and IT-opt
+    would get no gradient back to the injected latent leaves.
+
+    Keeping guidance at zero leaves the diffusion *advance* unguided, realizing "coordinate
+    guidance disabled" without editing the sampler. The reward is recomputed by the trajectory
+    scaler on ``denoised``, so there is exactly one denoiser forward and one reward evaluation
+    per step (the reference's Tier-2 single-forward optimization, for free).
+    """
+
+    requires_gradients = True
```
k-chrispens, line 96: I think it would be worth moving this into the step scalers module, rather than having it here

docs/IT_OPT_DESIGN.md
k-chrispens: esp. w.r.t. the new BoltzPerturb thing

src/sampleworks/core/rewards/geometry.py
```diff
@@ -0,0 +1,242 @@
...
+#   coords: Tensor                coords                 should be a Tensor (~ a numpy array)
+#   bond_power: int = 2           bond_power = 2         should be an int
+#   bond_tolerance: float = 0.2   bond_tolerance = 0.2   should be a float
+#   device: torch.device | str    device                 should be a torch.device or a str
+#   f(...) -> float               f(...)                 f should return a float
+#   f(...) -> Tensor              f(...)                 f should return a Tensor
+# -------------------------------------------------------------------------------
+
+
+def _covalent_radius(element: str) -> float:
```
k-chrispens, line 47: This could go in some sort of util module instead of here, I imagine it would be broadly useful.

src/sampleworks/core/scalers/latent_optimization.py
```diff
@@ -0,0 +1,566 @@
...
+        self.anchor_weight_single = anchor_weight_single
+        self.anchor_weight_pair = anchor_weight_pair
+        self.bond_length_weight = bond_length_weight
+        self.single_attr = single_attr
+        self.pair_attr = pair_attr
+
+    def sample(
+        self,
+        structure: dict,
+        model: FlowModelWrapper,
```
k-chrispens, line 222: I think for now this is OK, but in the long term if we want to configure this so that e.g. something like ROCKET is implementable, we would want to expand this beyond a flow model wrapper. This might require handling the two differently here, but it would be useful to make this more general, rather than specific. For example, the latent targets might not be the pair and single rep, for some arbitrary model (e.g. protpardelle). This will require some redesign/refactoring of this code, as Marcus has indicated. I don't think it's a major roadblock for the current code, but in the next version of this I think it will be necessary.

The way I had envisioned this was that a FlowModelWrapper should be wrapped in an adapter class. In the step() function, it would take the diffusion conditioning as input rather than the denoised coordinates, if that makes sense. So this would be a FlowModelWrapper wrapped in a StructureModelWrapper, which would clarify the contract between the latent space optimization and the model

src/sampleworks/core/scalers/latent_optimization.py
```diff
@@ -0,0 +1,566 @@
...
+                    baselines=baselines,
+                    bond_geometry=bond_geometry,
+                )
+                losses.append(data_loss)
+                steps.set_postfix(loss=data_loss)
+
+            coords = step_output.state.detach()
+        return losses
+
+    def _latent_adam_step(
```
k-chrispens, line 455: Should we rename this? I think it’s currently not requiring that you use Adam, so it’s probably worth changing it to be _latent_optim_step instead of what it is now.

src/sampleworks/models/latent_adapter.py
k-chrispens: maybe to abstract out s and z from this process, we use some arbitrary Tensor containing dataclass as the "latent". I don't think we should have any issues with the reward function because the reward function is just taking coordinates as input.

src/sampleworks/models/latent_adapter.py
```diff
@@ -0,0 +1,144 @@
...
+    "boltz1": "z",
+    "boltz2": "z",
+    "protenix": "z_trunk",
+    "rf3": "z_trunk",
+}
+
+
+# "conditioning" throughout this file is the model's conditioning object: a (frozen) dataclass
+# carrying the latents ``s``/``z`` (and other cached state) as named attributes.
+class AttrLatentIO:
```
k-chrispens, line 65: I think this could be abstracted further to recursively identify any Tensors from a list of keys and then turn them into proper leaf tensors

src/sampleworks/models/latent_adapter.py
```diff
@@ -0,0 +1,144 @@
...
+# ============================ protocol (template / interface) ============================
+# A Protocol is just a checklist of methods a class must have -- think a C header, or an abstract
+# base class, but *structural*: any class that already has these methods counts as a LatentIO
+# without inheriting from anything. This definition does not run; it only documents the contract and
+# lets the ty checker flag a mismatch. @runtime_checkable also lets isinstance(obj, LatentIO) work,
+# which the unit tests use. The real implementation is above: AttrLatentIO is the LatentIO.
+
+
+@runtime_checkable
+class LatentIO(Protocol):
```
k-chrispens, line 120: This LatentIO protocol will need to be rewritten in order to deal with Protpardelle because right now it hard codes the dependency on single and pair representations

src/sampleworks/utils/guidance_script_utils.py
```diff
@@ -633,6 +645,76 @@ def _run_guidance(args: GuidanceConfig, guidance_type: str, model_wrapper, devic
...
+        try:
+            single_attr = DEFAULT_SINGLE_REP_ATTR[model_key]
+            pair_attr = DEFAULT_PAIR_REP_ATTR[model_key]
+        except KeyError as e:
+            raise ValueError(
+                "Latent optimization has no latent-attribute names registered for model "
+                f"{model_key!r}."
+            ) from e
+
+        guidance = LatentOptimization(
```
k-chrispens, line 691: I know it isn't currently like this, but the vision is that one day we shouldn't need all this branching and can just pass the guidance type into the constructor based on some configuration you pass into this guidance script! No problem with it now, just want to advocate for the vision

tests/models/test_latent_optimization.py
```diff
@@ -0,0 +1,139 @@
...
+import pytest
+import torch
+from sampleworks.core.scalers.latent_optimization import LatentAnchor, LatentOptimization
+from sampleworks.models.latent_adapter import AttrLatentIO
+from sampleworks.models.protocol import GenerativeModelInput
+from torch import Tensor
+
+
+@dataclass(frozen=True)
+class _Cond:
```
k-chrispens, line 27: For the next iteration, it would be worth creating a test dataclass with arbitrary latents that are not necessarily single and pair and also maybe contain some other information in other tensors that we don't care about. That way we can test that that isn't gumming up the process.

k-chrispens (changes requested):
This looks really great. Lots of good work here!! I think there are a few places that could be modified so that it's easier to expand to new models as they come in. This could be done by abstracting out the specific latents and the specific structure of the latents.

There are some design decisions that aren't necessarily ideal that we've carried over from the reference implementation. It would be worth trying to abstract out some of these things so that an ironing out exactly what the contract for latent optimization here is. E.g. what do we expect from a model wrapper? What do we expect from the features? How can we make it so that the sampler is specifically associated with a flow model wrapper when passed in, but that the sampler isn't required if we pass a different type of model in for something like ROCKET?

That will help inform then how we change this in the future so that we can accept arbitrary models that may not carry out a diffusion sampling loop, but still do some sort of iterative updating of the latents to produce a better structure

tests/models/test_latent_optimization.py
smallfishabc: Hi Krason, good suggestion. Even though all AF3 like model heavily rely on S and Z, the model like protpardelle do not use those. So I think this one will need to be considered with specific models.

src/sampleworks/models/latent_adapter.py
smallfishabc: Hi Karson, the current algorithm will not run on Protpardelle due to a very different architecture. So the fix will be a big Pr and will not be an intuitive one.

src/sampleworks/core/scalers/latent_optimization.py
smallfishabc: Hi Karsen, I have tested the SGD and other optimizers. Currently, we need to use Adam for performance reason. But we can discuss this later.

src/sampleworks/core/scalers/latent_optimization.py
smallfishabc: Hi Karson, the ROCKET is a one time optimization process on the MSA module. So we will implement that but I think those suggestion should be included in a different architecture design. The ITO method in this PR is strictly limited to the diffusion process.

src/sampleworks/core/rewards/geometry.py
smallfishabc: Good Call.

src/sampleworks/core/scalers/latent_optimization.py
smallfishabc: I think I opened some issue based on Marcus suggestion on how to combine those with existing step scalers module.

src/sampleworks/utils/guidance_script_utils.py
smallfishabc: Hi Karson, I do not think the branching itself is the problem. Those branches are not just picking a class, they pass disjoint arguments, RF3 needs the chiral flags and msa path, Boltz needs an out dir, Protpardelle sets the edm sampler kwargs, and the guidance type also feeds back into the model preprocessing, we disable the Protenix shared vars cache only for latent opt so the gradient can reach z_trunk. So a config would not remove those combinations, it only moves them out of Python where ty cannot check them and a bad one fails in the middle of a run instead of at construction. I think explicit is the right trade until the model and guidance axes are actually independent.

smallfishabc:
> This looks really great. Lots of good work here!! I think there are a few places that could be modified so that it's easier to expand to new models as they come in. This could be done by abstracting out the specific latents and the specific structure of the latents.
> 
> There are some design decisions that aren't necessarily ideal that we've carried over from the reference implementation. It would be worth trying to abstract out some of these things so that an ironing out exactly what the contract for latent optimization here is. E.g. what do we expect from a model wrapper? What do we expect from the features? How can we make it so that the sampler is specifically associated with a flow model wrapper when passed in, but that the sampler isn't required if we pass a different type of model in for something like ROCKET?
> 
> That will help inform then how we change this in the future so that we can accept arbitrary models that may not carry out a diffusion sampling loop, but still do some sort of iterative updating of the latents to produce a better structure

Hi Karson, thanks for the review. On the latents, s and z are a scope decision, not a carryover. All four models we support are AF3 like and use exactly those two, they differ only in the attribute name, which is why the adapter takes names instead of a generic container. 
ROCKET is a one time optimization on the MSA module and Protpardelle has a different architecture, so neither one fits the diffusion contract this PR implements. I think the model wrapper contract should be written when one of them actually lands, otherwise we are designing against a guess. I opened #366 and will file the rest so they are tracked outside this PR.

docs/IT_OPT_DESIGN.md
marcuscollins: Still need to address this.

docs/IT_OPT_DESIGN.md
marcuscollins: @smallfishabc can you make an issue to investigate this.

src/sampleworks/core/scalers/latent_optimization.py
marcuscollins: @smallfishabc please make sure that issue exists, and reference it here as a `# TODO` comment.

src/sampleworks/core/scalers/latent_optimization.py
marcuscollins: This still needs addressing.

src/sampleworks/core/scalers/latent_optimization.py
marcuscollins: 1. I'm not terribly interested rn in implementing ROCKET. 
2. It might still be worth re-factoring. @k-chrispens can you work with @smallfishabc to design this? I'm not a fan of changing the API of step(), but I might be convinced. In my view, that's the model contract. Perhaps we need to get together and sketch this out.

src/sampleworks/core/scalers/latent_optimization.py
marcuscollins: Needs to be addressed.

src/sampleworks/core/scalers/latent_optimization.py
marcuscollins: You should still take Adam out of the name, since it isn't required. Just call it `_latent_optimizer_step`

src/sampleworks/models/latent_adapter.py
marcuscollins: @smallfishabc if you have created the issue, please reference it here as a `# TODO`

src/sampleworks/models/latent_adapter.py
marcuscollins: @k-chrispens can you add an issue. I don't want to hold up the PR on this.

src/sampleworks/models/latent_adapter.py
marcuscollins: Make an issue, reference it here.

docs/CODEBASE_GUIDE.md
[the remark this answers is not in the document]
k-chrispens: IMO this entire doc should be a separate PR, and we should vet this more thoroughly. This appears largely LLM written, which is fine, but it uses a lot of terminology that we don't in practice and maps out in a lot more detail various parts of the codebase that may change, so then this doc also become a lot more brittle for agents than we would want.

marcuscollins (commented):
There are some unresolved comments that still need to be addressed. We also need to have a meeting with Karson and figure out how we'll refactor in the future to handle things like branching more seamlessly. This will enable us to do things like combine ITO and DPS approaches.

But we're close. Most of the remaining comments are just to create issues and note them as comment in the code. Let's get that done and then we can merge.

docs/IT_OPT_DESIGN.md
```diff
@@ -0,0 +1,179 @@
...
+flows through the coordinate recursion — it is a greedy, per-step latent gradient, **not**
+backprop-through-sampling. (Unrolling the full sampler is intractable: the EDM sampler re-noises and
+detaches each step, and a true unroll is O(steps·N²·d) memory.)
+
+## 3. Components and where they live
+
+| Component | File | What it is |
+|---|---|---|
+| `LatentOptimization` | [core/scalers/latent_optimization.py](../src/sampleworks/core/scalers/latent_optimization.py) | The scaler / entry point. Owns the loop, the ensemble, and the final sampling pass. |
+| `LatentAnchor` | same file | `Σ wᵢ·mean((latentᵢ − baselineᵢ)²)` — mean-squared drift from the baseline latents. Regularizer, added to the loss directly (not a reward). |
```
k-chrispens, line 57: Things like this could probably be rendered as latex instead of unicode if we want to make it pretty :)

docs/IT_OPT_DESIGN.md
```diff
@@ -0,0 +1,179 @@
...
+**Invariants worth stating:**
+- **Optimization is not sampling.** An optimized `(s, z)` is a point estimate; reporting it as a
+  single state collapses the Boltzmann/population weighting. If populations matter, sample the latent
+  rather than optimizing it. Ensembles here come from fresh prior noise per member, not from a
+  posterior over latents.
+- **On-manifold discipline.** Pushing `z` hard buys density fit with broken geometry; the anchor and
+  `BondGeometryReward` are the counter-pressure. Efficacy must be judged on held-out fit against
+  matched-compute baselines, not train-set loss (see [IT_OPT_TESTING.md](IT_OPT_TESTING.md)).
+
+## 7. Why `s`/`z` and not the MSA
```
k-chrispens, line 162: I think we could remove this section and put it under open problems or something like that instead

docs/IT_OPT_TESTING.md
```diff
@@ -0,0 +1,175 @@
...
+
+The runtime lives on the Astera pod; edit locally, sync, run in the Protenix env. Work up the levels
+and stop at the first that fails.
+
+```bash
+pixi run -e protenix-dev python your_it_opt_test.py            # needs the checkpoint + a map/structure
+pixi run -e protenix-dev python -m pytest tests/models/test_latent_adapter.py -q   # CPU, no checkpoint
+```
+
+### Level 0 — Does the gradient reach the latent? (the #1 failure mode)
```
k-chrispens, line 42: Is there a way to implement this as a test instead of as a script here?

src/sampleworks/core/rewards/geometry.py
k-chrispens: This appears still here rather than in utils?

src/sampleworks/core/scalers/latent_optimization.py
k-chrispens: I don't see the issue... can you create it? Or just move it over now?

src/sampleworks/core/scalers/latent_optimization.py
k-chrispens: I think we should still rename, since this takes any optimizer as input. Then we should just specify elsewhere that Adam is the most performant.

src/sampleworks/models/latent_adapter.py
k-chrispens: Yep. We should add an issue on this

src/sampleworks/utils/guidance_script_utils.py
k-chrispens: Yes, what I'm advocating for here is just that the model and guidance axes should continue to get less and less dependent until the point where we don't need all this conditional structure

tests/models/test_latent_optimization.py
k-chrispens: #417

tests/rewards/test_geometry.py
k-chrispens, line 1: This file is currently a stub and has a (I would argue) nearly useless test. I think delete this or expand it to test the full geometry loss

k-chrispens (commented):
There are a few lagging comments to clean up here still, but nothing substantial that is really holding me back from merging. I do think we ought to sit down and come up with some more full-pipeline testing for this (and the other guidance types for that matter) too, but that should come in a later PR.

smallfishabc:
> There are some unresolved comments that still need to be addressed. We also need to have a meeting with Karson and figure out how we'll refactor in the future to handle things like branching more seamlessly. This will enable us to do things like combine ITO and DPS approaches.
> 
> But we're close. Most of the remaining comments are just to create issues and note them as comment in the code. Let's get that done and then we can merge.

I have fixed two more comments including the rename and shape problem. The other issues are added with #TODO
Here is the table. 

comment | issue
-- | --
latent_optimization.py:96 — _GradEnablingScaler into step scalers | #364
latent_optimization.py:126 — LatentAnchor into core.rewards | #363
latent_adapter.py:20 — ITO with other feature schemes (Protpardelle) | #416
latent_adapter.py:65 — recursive tensor identification | #417 (filed by @k-chrispens)
latent_adapter.py:120 — what LatentIO should be | #418
IT_OPT_DESIGN.md:154 — sampling the latent rather than optimizing it | #419

<br class="Apple-interchange-newline">

smallfishabc:
> There are a few lagging comments to clean up here still, but nothing substantial that is really holding me back from merging. I do think we ought to sit down and come up with some more full-pipeline testing for this (and the other guidance types for that matter) too, but that should come in a later PR.

- `:561` — renamed to `_latent_optimizer_step`, docs updated too.
- `tests/rewards/test_geometry.py` — deleted. Filed #420 to cover `BondGeometryReward` properly.
- `docs/CODEBASE_GUIDE.md` — untracked, out of this PR.
- `IT_OPT_DESIGN.md:162` — §7 folded into §6 as an invariant.

**Issues filed, `# TODO` added at each site**

| comment | issue |
|---|---|
| `latent_optimization.py:97` — `_GradEnablingScaler` into step scalers | #364 |
| `latent_adapter.py:65` — recursive tensor identification | #417 |
| `latent_adapter.py:124` — `LatentIO` hardcodes single/pair | #418 |
| `test_latent_optimization.py:31` — test dataclass with arbitrary latents | #417 |
| `geometry.py:47` — `_covalent_radius` into a util module | #421 |
| `IT_OPT_TESTING.md:42` — make it a test | #422 |

**Not done**

- `IT_OPT_TESTING.md:42` — needs a checkpoint and a GPU, which we aren't set up for.
  Left as a manual recipe with a TODO pointing at #422.
- `IT_OPT_DESIGN.md:57` — LaTeX over unicode: GitHub tables don't render it reliably,
  so I left the unicode. Say if you'd rather I try.

PR #317, opened 2026-07-13
perf(rf3): reuse initialized model runtime

### LINKED ISSUE #291: Make sure we can pass instantiated model objects into all ModelWrapper classes for proper model re-use.
In Boltz1Wrapper, Boltz2Wrapper, and ProtenixWrapper we accept an argument `model` which corresponds to the already instantiated model object (e.g., for Boltz2Wrapper, a model of class Boltz2).  This allows us to re-use an existing, pre-loaded model to save time. The RF3Wrapper does not accept this argument for some reason, and it isn't used in guidance_script_utils.py:get_model_and_device.py. Following the logic in other classes, make it so we re-use pre-loaded RF3 models.

manzuoni-astera:
## Summary
- allow `RF3Wrapper` to reuse a model from an existing wrapper
- reuse the complete originating inference engine, trainer, pipeline, and Fabric context instead of loading a second checkpoint
- reject unsafe bare modules, checkpoint mismatches, and cross-device reuse
- forward pre-loaded RF3 models through `get_model_and_device`

Closes #291

## Why the engine is reused
RF3 preprocessing and model execution are coupled through Foundry's initialized `RF3InferenceEngine`. Replacing only `trainer.state["model"]` after initialization would still load the checkpoint and could attach the model to the wrong Fabric runtime. The originating engine is therefore retained with the model and reused as one runtime context.

## Validation
- `pytest tests/models/test_rf3_device_selection.py tests/utils/test_guidance_script_utils.py -q` (21 passed, 1 skipped)
- `ty check` on touched files in `rf3-dev`
- `ruff check` and `ruff format --check` on touched files
- `git diff --check`

## Summary by CodeRabbit

* **New Features**
  * RF3 wrappers can now reuse pre-initialized models and inference runtimes.
  * Added validation for checkpoint, device, and model/runtime compatibility when reusing models.
  * Preloaded RF3 models are now correctly passed through model and device setup utilities.

* **Bug Fixes**
  * Prevented unnecessary inference engine reinitialization for reused RF3 models.
  * Added clear validation for invalid or incomplete model reuse contexts.

### DIFF
--- a/src/sampleworks/models/rf3/wrapper.py
+++ b/src/sampleworks/models/rf3/wrapper.py
@@ -25,6 +25,9 @@ from sampleworks.utils.guidance_constants import StructurePredictor
 from sampleworks.utils.msa import MSAManager
 
 
+_INFERENCE_ENGINE_ATTR = "_sampleworks_rf3_inference_engine"
+
+
 @dataclass(frozen=True, slots=True)
 class RF3Conditioning:
     """Conditioning tensors from RF3 trunk forward pass.
@@ -218,6 +221,7 @@ class RF3Wrapper:
         checkpoint_path: str | Path,
         msa_manager: MSAManager | None = None,
         device: torch.device | str | None = None,
+        model: Any | None = None,
     ):
         """
         Parameters
@@ -232,6 +236,11 @@ class RF3Wrapper:
             parallel jobs that must target distinct GPUs — passing an ``int``
             to Fabric (the default) always resolves to GPU 0, which serialises
             otherwise-parallel workers onto a single device.
+        model: Any | None
+            Model previously obtained from another ``RF3Wrapper``. RF3 model
+            reuse also reuses the originating inference engine because its
+            trainer, preprocessing pipeline, Fabric strategy, and model are a
+            single initialized runtime context.
 
             References: https://lightning.ai/docs/fabric/stable/fundamentals/launch.html
               devices argument to fabric run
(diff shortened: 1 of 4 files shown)


src/sampleworks/models/rf3/wrapper.py
```diff
@@ -25,6 +25,9 @@
 from sampleworks.utils.msa import MSAManager
 
 
+_INFERENCE_ENGINE_ATTR = "_sampleworks_rf3_inference_engine"
```
marcuscollins, line 30: Maybe add a comment saying where this attribute is set/created?

marcuscollins (approved):
One small nit, just to add a comment on where to find how a particular attribute is set. If you can add that, go ahead and merge. Thanks.

src/sampleworks/models/rf3/wrapper.py
manzuoni-astera: Added a comment directly above `_INFERENCE_ENGINE_ATTR` explaining that it is attached to the trainer-owned model after `RF3InferenceEngine` initialization in `RF3Wrapper.__init__`, and that reused wrappers retrieve the complete runtime through it. I also added an identity check so a model cannot spoof an unrelated engine context.

manzuoni-astera:
Addressed the remaining nit in `1ab1c4c` by documenting exactly where `_INFERENCE_ENGINE_ATTR` is attached and how it is consumed. The later ownership-validation update also ensures a supplied model is exactly the trainer-owned model for that engine context.

PR #318, opened 2026-07-13
fix(tooling): align Ruff with Python 3.11

### LINKED ISSUE #311: Update to use newer type parameters
When working on bringing protpardelle into SW, I ran into the following type of listing error:
```
UP047 Generic function `require_boltz` should use type parameters
  --> src/sampleworks/utils/imports.py:57:5
   |
56 | @overload
57 | def require_boltz(message: F) -> F: ...
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^
58 | @overload
59 | def require_boltz(message: str | None = None) -> Callable[[F], F]: ...
   |
help: Use type parameters
```

F in this case is a `TypeVar`: `F = TypeVar("F", bound=Callable[..., Any])`. The newer way to do this is, I think:
```
def require_boltz[F: Callable[..., Any]](message: F) -> F
```
There are 11 such errors in imports.py, and presumably we can use this elsewhere as well. We should check that it is actually compatible with our version of Python.

For now I've just added `# ruff: noqa: UP047` to that file.

manzuoni-astera:
## Summary
- align Ruff's target version with the project's declared Python 3.11 minimum
- stop Ruff from recommending PEP 695 generic syntax that is a syntax error on a supported Python version
- retain the compatible `TypeVar` form until Python 3.11 support is intentionally dropped

Closes #311

## Validation
- `ruff check .`
- `ruff format --check .` (144 files already formatted)
- parsed `pyproject.toml` with Python `tomllib`
- `git diff --check`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -198,7 +198,7 @@ testpaths = ["tests"]
 extend-include = ["*.ipynb"]
 line-length = 100
 src = []
-target-version = "py312"
+target-version = "py311"
 
 [tool.ruff.lint]
 fixable = ["I001", "F401", "UP"]


marcuscollins (approved):
Nice.

Issue #305, opened 2026-07-09
Add a physical-plausibility reward (all-atom energy) to complement density-only guidance?


etherealsunshine:
### Summary
From reading the repo, the guidance currently centers on real-space density agreement with no explicit physical-plausibility term in the loop.  I wanted to get thoughts on adding an all-atom energy/plausibility reward as a RewardFunctionProtocol.

###  Current implementation 
What I'm seeing in the current code 
- The only steering reward is core/rewards/real_space_density.py (differentiable real-space density, via the qFit scattering machinery).
- The only "clash" logic is in metrics/sidechain_metrics.py, as a post-hoc evaluation metric, which isnt part of steering/selection.

### Proposed enhancement
- v1: a plausibility reward implementing RewardFunctionProtocol, backed by an all-atom score (e.g. Rosetta ref2015). Since it's non-differentiable, it'd suit FK/SMC steering (evaluate-and-resample) and/or post-hoc re-ranking, rather than gradient guidance.
- v2 : a differentiable, GPU-native learned energy (e.g. an energy-based transformer) so the same signal can be used for gradient guidance.

etherealsunshine:
Would be happy to implement this! Let me know what you think @marcuscollins

k-chrispens:
Would recommend checking out https://github.com/uw-ipd/tmol

etherealsunshine:
Seems perfect, thanks! The only caveat with them is tmol has no prebuilt wheel for Torch 2.7.1 (SampleWorks' version) where wheels cover 2.8–2.12, so a bump would greatly help user-friendliness. For now im going to source compile

marcuscollins:
@etherealsunshine this seems like a good improvement. Please go ahead and see what you can do!

PR #319, opened 2026-07-14
feat(rewards): add tmol plausibility and composite rewards

etherealsunshine:
## Summary
- Adds **tmol** (Rosetta `beta_nov2016` energy) as a differentiable physical-plausibility
reward for guided sampling, plus a **CompositeReward** that combines rewards (e.g. density
fit + plausibility) as a weighted sum. Both satisfy the existing `RewardFunctionProtocol`
and are selectable via `--reward-type`.
- Addresses #305 

## What's added
- `TmolPlausibilityReward` - differentiable beta2016 energy; maps model atom order  tmol's
  canonical per-residue grid via a scatter map built once in `prepare()`.
- `CompositeReward` - weighted sum of rewards; itself a `RewardFunctionProtocol`; forwards
  the optional `prepare()` hook to sub-rewards that need it.
- CLI/config: `--reward-type {density,plausibility,composite}`, `--tmol-weight`;
  `--density`/`--resolution` now required only in density mode.

## Validation
Tested FKS and DPS on [Modal](https://modal.com/) an A10G GPU, ensuring energy decreases monotonically and end-structures do not have any clashes.

## To be discussed: CUDA architectures
tmol compiles from source on every install, so `CMAKE_CUDA_ARCHITECTURES` is a
committed build setting that affects everyone. It's currently **`sm_80;86`** (A100 / A40 /
RTX 3090). Widening to tmol's own default **`80;86;89;90`** would also cover sm_89 (L40 /
RTX 4090) and sm_90 (H100), but makes every build slower and raises peak build memory (may
need a lower `CMAKE_BUILD_PARALLEL_LEVEL` to avoid OOM).  Let me know how to resolve this 
 

## Summary by CodeRabbit

* **New Features**
  * Added a tmol-based physical plausibility reward using the Rosetta beta2016 energy.
  * Introduced `CompositeReward` to combine multiple rewards via weighted sums.
  * Extended guidance configuration and CLI to support reward modes: density, plausibility, and composite (with configurable tmol weighting).
  * Added reward pre-setup support via a new preparation protocol, wired into guidance sampling.
  * Enabled tmol support across Pixi builds and environments (including CUDA builds).
* **Tests**
  * Added coverage for reward composition, gradients, validation, and preparation forwarding.


src/sampleworks/core/rewards/protocol.py
```diff
@@ -14,6 +14,18 @@
...
+@runtime_checkable
+class PreparableRewardProtocol(Protocol):
+    """A reward with an optional one-time ``prepare`` hook.
+
+    Rewards that need the concrete model atom ordering (e.g. tmol builds a scatter
+    map from it) implement ``prepare``; the sampler calls it once before sampling.
+    Rewards without it (e.g. density fit) simply don't match this protocol.
+    """
+
+    def prepare(self, atom_array: Any) -> Any: ...
```
marcuscollins, line 26: Several issues here:
1. What kind of object does `prepare` return? I think this needs to be more cleanly defined, and in fact I'd prefer if it returned None and only modified the "reward" object itself. 
2. atom_array needs a more specific type than `Any`. It is presumably at `AtomArray` or `AtomArrayStack`.

src/sampleworks/core/rewards/composite.py
```diff
@@ -0,0 +1,50 @@
...
+                inverse_indices,
+            )
+            total = term if total is None else total + term
+        return total
+
+    def prepare(self, atom_array):
+        for rewards in self.rewards:
+            if hasattr(rewards, "prepare"):
+                rewards.prepare(atom_array)  # prepare from TmolPlausibility
+        return self
```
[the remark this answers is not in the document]
marcuscollins: @etherealsunshine these are both good suggestions, please implement them.

src/sampleworks/core/rewards/composite.py
```diff
@@ -0,0 +1,50 @@
...
+
+    def __init__(
+        self,
+        rewards: list,
+        weights: list[float] | None = None,
+    ):
+
+        if weights is None:
+            weights = [1.0] * len(rewards)
+        if len(weights) != len(rewards):
```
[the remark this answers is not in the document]
marcuscollins: @etherealsunshine this is good practice: fail early with a clear explanation. Please implement.

src/sampleworks/core/rewards/composite.py
```diff
@@ -0,0 +1,50 @@
...
+
+    def __call__(
+        self,
+        coordinates: Float[torch.Tensor, "batch n_atoms 3"],
+        elements: Int[torch.Tensor, "batch n_atoms"],
+        b_factors: Float[torch.Tensor, "batch n_atoms"],
+        occupancies: Float[torch.Tensor, "batch n_atoms"],
+        unique_combinations: torch.Tensor | None = None,
+        inverse_indices: torch.Tensor | None = None,
+    ) -> Float[torch.Tensor, ""]:
```
marcuscollins, line 31: needs a good docstring. In particular you need to explain `unique_combinations` and `inverse_indices`.

src/sampleworks/core/rewards/composite.py
```diff
@@ -0,0 +1,50 @@
...
+        for reward, weight in zip(self.rewards, self.weights):
+            term = weight * reward(
+                coordinates,
+                elements,
+                b_factors,
+                occupancies,
+                unique_combinations,
+                inverse_indices,
+            )
+            total = term if total is None else total + term
```
marcuscollins, line 43: What type will `term` be? Presumably a torch.Float of size 1? This is a technically correct but unusual way to initialize `total` correctly, and it doesn't provide much insight into what `total` is. I would initialize `total` to be zeros that match what's expected from the rewards, and then just add to that.

src/sampleworks/core/rewards/composite.py
```diff
@@ -0,0 +1,50 @@
...
+        return total
```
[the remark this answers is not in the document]
marcuscollins: @etherealsunshine my above comment will help fix this problem too.

src/sampleworks/utils/guidance_script_arguments.py
```diff
@@ -413,7 +420,26 @@ def as_dict(self) -> dict[str, Any]:
 def add_generic_args(parser: argparse.ArgumentParser | GuidanceConfig):
     """Add CLI arguments shared by all models and guidance methods."""
     parser.add_argument("--structure", type=str, required=True, help="Input structure")
-    parser.add_argument("--density", type=str, required=True, help="Input density map")
+    parser.add_argument(
+        "--density",
+        type=str,
+        default=None,
+        help="Input density map (required for --reward-type density)",
```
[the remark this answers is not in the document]
marcuscollins: @etherealsunshine Definitely if this is now optional, you need to update the type annotation in GuidanceConfig, and either GuidanceConfig, or probably the density-based reward function itself, needs to raise a clear error if this argument is unset and `--reward-type` is "density". I think I prefer the error checking to be in the reward functions themselves, so that each reward function can check errors that make sense for it to check.

src/sampleworks/utils/guidance_script_arguments.py
```diff
@@ -413,7 +420,26 @@ def as_dict(self) -> dict[str, Any]:
...
+    )
+    parser.add_argument(
+        "--reward-type",
+        type=str,
+        default="density",
+        choices=["density", "plausibility", "composite"],
```
marcuscollins, line 433: Presumably you will check what these values are somewhere else when you want to instantiate your reward(s). Therefore make these all part of a string Enum (there may now be a StrEnum class in Python). That way you guard against spelling errors that can result in silent behavior changes. Then, anywhere these strings are used, check against the values in the Enum and use the Enum values directly instead of strings.

src/sampleworks/utils/guidance_script_arguments.py
```diff
@@ -413,7 +420,26 @@ def as_dict(self) -> dict[str, Any]:
...
+        help="Reward: experimental density fit, tmol beta2016 plausibility energy, "
+        "or a weighted combination",
+    )
+    parser.add_argument(
+        "--tmol-weight",
```
marcuscollins, line 438: this will quickly get unwieldy. I would look around for a better solution, or if necessary, pass in a json-parsable string that maps from reward function to weight. Another option would be to have a json or yaml file config, but it may be too soon for that.

src/sampleworks/utils/guidance_script_arguments.py
```diff
@@ -413,7 +420,26 @@ def as_dict(self) -> dict[str, Any]:
...
+        type=float,
+        default=1.0,
+        help="Multiplier on the tmol plausibility energy (--reward-type plausibility)",
```
marcuscollins, line 441: Use consistent names: if you use "plausibility" as the reward name, this should be named "--plausibility-weight" or vice versa.

src/sampleworks/core/rewards/composite.py
```diff
@@ -0,0 +1,50 @@
...
+    """Weighted sum of reward functions using the RewardFunctionProtocol"""
+
+    def __init__(
+        self,
+        rewards: list,
+        weights: list[float] | None = None,
+    ):
+
+        if weights is None:
+            weights = [1.0] * len(rewards)
```
marcuscollins, line 17: I would favor the default being 1/len(rewards). More than likely, the rewards will need to be scaled, but if we aren't scaling them relative to each other, I'd rather we not have these weights interact with the step size parameter.

src/sampleworks/core/scalers/fk_steering.py
```diff
@@ -112,6 +112,11 @@ def sample(
             ensemble_size=self.ensemble_size,
         )
 
+        # Optional structural hook (see PureGuidance): tmol-style rewards build their
+        # atom map here against the array to_reward_inputs uses; others skip.
+        if isinstance(reward, PreparableRewardProtocol):
+            reward.prepare(processed.model_atom_array or processed.atom_array)
+
```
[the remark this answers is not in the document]
marcuscollins: @etherealsunshine I've noticed that more and more people are using the `is not None` form, rather than relying on the "falseness" of None. It's more specific, and it avoids the problem that the "truthiness" of different classes can be hard to define, e.g., is an empty AtomArray True or False? (I'd say False, but not everyone will agree and we don't want to rely on biotite's developers to necessarily define that for us.)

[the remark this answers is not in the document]
marcuscollins: also @etherealsunshine remind me what the difference between these two is?

src/sampleworks/core/scalers/pure_guidance.py
```diff
@@ -87,6 +87,12 @@ def sample(
             ensemble_size=self.ensemble_size,
         )
 
+        # Optional structural hook (Protocol, not inheritance): rewards that need the
+        # concrete atom ordering of the coordinates -- e.g. tmol builds its scatter map --
+        # prepare against the same array to_reward_inputs uses. Rewards without prepare skip.
+        if isinstance(reward, PreparableRewardProtocol):
+            reward.prepare(processed_structure.model_atom_array or processed_structure.atom_array)
```
marcuscollins, line 94: same comment as above. Use `is not None` instead

src/sampleworks/utils/guidance_script_utils.py
```diff
@@ -269,6 +269,90 @@ def get_reward_function_and_structure(
...
+    -------
+    tuple[TmolPlausibilityReward, dict]
+        The reward function and the parsed atomworks structure dict.
+    """
+    import tmol
+    from tmol.io.canonical_ordering import (
+        default_canonical_ordering,
+        default_packed_block_types,
+    )
+
```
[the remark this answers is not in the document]
marcuscollins: I think the correct solution to this has two parts:
1. make sure all relevant environments have the tmol feature (or just make it default, is there a reason not to? It's possible to go too crazy paring down your environment to only what's needed for the current invocation of a command.)
2. Maybe add the suggested check that tmol is importable.

src/sampleworks/utils/guidance_script_utils.py
```diff
@@ -269,6 +269,90 @@ def get_reward_function_and_structure(
...
+    from sampleworks.core.rewards.tmol_plausibility import TmolPlausibilityReward
+
+    logger.debug(f"Loading structure from {structure_path}")
+    safe_structure_path = resolve_mixed_hetatm_atom_altlocs(Path(structure_path))
```
marcuscollins, line 307: 306-315 is copied from somewhere else, no? Make it a utility function rather than copy the code here. You'll have to modify some tests as well I think.

src/sampleworks/utils/guidance_script_utils.py
```diff
@@ -269,6 +269,90 @@ def get_reward_function_and_structure(
...
+        safe_structure_path,
+        hydrogen_policy="remove",
+        add_missing_atoms=False,
+        ccd_mirror_path=None,
+    )
+    if str(safe_structure_path) != str(structure_path):
+        safe_structure_path.unlink()  # delete the temporary file if one was created
+
+    logger.info("Creating tmol plausibility reward function")
+    co = default_canonical_ordering()
```
marcuscollins, line 318: Use meaningful variable names. Here, you can kind of get away with it because they only uses are close by, but in general it costs little to make the code more readable.

src/sampleworks/utils/guidance_script_utils.py
```diff
@@ -269,6 +269,90 @@ def get_reward_function_and_structure(
...
+    pbt = default_packed_block_types(device)
+    sfxn = tmol.beta2016_score_function(device)
```
marcuscollins, line 320: Can you describe this function in a comment? Also, add a link to where it is documented/described. Are there others in tmol that we might eventually want to pipe through?

src/sampleworks/utils/guidance_script_utils.py
```diff
@@ -269,6 +269,90 @@ def get_reward_function_and_structure(
...
+    setup; the density term needs a map(--density/--resolution)"""
+
+    from sampleworks.core.rewards.composite import CompositeReward
+
+    if args.density is None or args.resolution is None:
+        raise ValueError(
+            "--reward-type composite needs --density and --resolution for the density term."
+        )
+
+    density_reward, structure = get_reward_function_and_structure(
```
marcuscollins, line 341: possibly should split up `get_reward_function_and_structure` and in any event rename it to `get_density_reward_and_structure` to distinguish it from the more general cases that now exist.

src/sampleworks/utils/guidance_script_utils.py
```diff
@@ -269,6 +269,90 @@ def get_reward_function_and_structure(
...
+    sfxn = tmol.beta2016_score_function(device)
+    reward_function = TmolPlausibilityReward(co, pbt, sfxn, device, weight=weight)
+    return reward_function, structure
+
+
+def get_composite_reward_and_structure(
+    args,
+    device: torch.device,
+) -> tuple[Any, dict[str, Any]]:
+    """Build a CompositeReward = density fit + tmol plausibility, weighted.
```
marcuscollins, line 329: Make it more general, there's no reason to limit ourselves to just these two.

src/sampleworks/utils/guidance_script_utils.py
```diff
@@ -269,6 +269,90 @@ def get_reward_function_and_structure(
...
+    logger.info("Creating tmol plausibility reward function")
+    co = default_canonical_ordering()
+    pbt = default_packed_block_types(device)
+    sfxn = tmol.beta2016_score_function(device)
+    reward_function = TmolPlausibilityReward(co, pbt, sfxn, device, weight=weight)
+    return reward_function, structure
+
+
+def get_composite_reward_and_structure(
+    args,
```
marcuscollins, line 326: needs type annotation (is this a GuidanceConfig?) and you need a docstring documenting all your arguments. 

In general I prefer passing in specific arguments, rather than config objects, since it makes it easier to understand what's getting used and type check it.

src/sampleworks/utils/guidance_script_utils.py
```diff
@@ -432,14 +516,30 @@ def _three_state_resolver(value: str | bool | None, default: bool) -> bool:
...
-        args.density,  # str/path to a map file.
-        device,  # this needs to come from the global context, not the args object.
-        args.em,
-        args.loss_order,
-        args.resolution,
-        args.structure,  # path/string to a structure file.
-    )
+    # v1 reward selection (becomes a proper --reward flag later): "plausibility" uses the
+    # tmol beta2016 energy prior (no map needed); anything else falls back to density fit.
+    if getattr(args, "reward_type", "density") == "plausibility":
```
marcuscollins, line 521: I think if you just make _all_ reward functions composite, and use a configuration dictionary that maps between reward types and weights, the logic here becomes simpler and much easier to maintain (you won't have to keep adding extra conditionals here for every new possibility.

marcuscollins (changes requested):
@etherealsunshine lots of comments and things to improve. The code can be more modular, and simpler at the same time. LMK if you have questions or want to discuss.

marcuscollins:
@etherealsunshine another big issue here is how well we understand what this code will do in actual use: what should the weights be, how much impact will it have on guidance. This is no longer purely research code, it is a product that we want others to start to use, so it needs to be very usable. I think before we merge this we need to do experiments to establish best options for using it, and whether it has measurable positive impact on the outcomes.

etherealsunshine:
@marcuscollins that definitely makes sense. I was talking to Karson about this yesterday about this in terms of whether he had any specific priors for how to decide weights and reward annealing for the same. Once I clean up the highlighted comments, I'm happy to run a few jobs experimenting with how strongly the density (and tmol) reward should influence the trajectory as a function of timestep and track changes

manzuoni-astera (commented):
Thanks for taking on this integration; a differentiable physical prior is valuable, and the preparation hook is a reasonable direction. I do not think the current implementation is ready to merge, however.

The main blocker is the coordinate/topology map in `tmol_plausibility.py`. Lines 99–102 filter the `AtomArray`, but lines 119–134 retain indices in that filtered space and line 246 applies them to the original coordinate tensor. A zero-occupancy atom therefore shifts every subsequent mapping. Residues are also keyed only by `(chain_id, res_id)`, which merges insertion-code residues, while skipped nonstandard residues and other chain discontinuities are not represented through `res_not_connected`. tmol may consequently create peptide bonds across gaps and guide toward an incorrect topology.

There is currently no automated test of this reward itself. Before merging, I would want CPU/GPU tests covering finite score and gradients, zero-occupancy indexing, insertion codes, missing heavy atoms, unsupported-only inputs, chain breaks, batch behavior, and at least one sampler integration using actual model atom order.

The CLI/config change also needs completion. Default density mode now accepts missing density or resolution, contrary to the existing tests at `test_guidance_cli.py:488-518`, and `GuidanceConfig.as_dict()` serializes missing density as `"None"`. Conditional validation and plausibility/composite CLI tests should be added.

Finally, please regenerate and commit `pixi.lock` and pin a compatible tmol version. An unbounded, source-built CUDA dependency is not reproducible, and neither standard CI nor GPU CI has run yet. The previously raised empty-composite, typing, modularity, and reward-calibration comments should also be resolved.

marcuscollins:
Closing as stale. May revisit later.

PR #320, opened 2026-07-14
refactor(config): distinguish model names from objects

manzuoni-astera:
## Summary
- rename structure-predictor name fields to `model_name` across guidance, grid-search jobs, and job results
- preserve the public `--model` CLI flag with an explicit argparse destination
- keep wrapper constructor `model` parameters unchanged because those hold loaded model objects
- intentionally make the serialized schema change cleanly without legacy pickle/JSON migration shims

Implements #289. This PR is intentionally stacked on #300 because it updates the Protpardelle-integrated guidance files; #289 should remain open until the stack reaches `main`.

The README CLI examples remain valid because `--model` is unchanged.

## Validation
- `pytest tests/cli/test_guidance_cli.py tests/utils/test_guidance_script_arguments.py tests/utils/test_guidance_script_utils.py -q` (110 passed)
- targeted `ty check` passes
- `ruff check` and `ruff format --check` on touched Python files
- `git diff --check`


src/sampleworks/utils/guidance_script_arguments.py
```diff
@@ -268,6 +268,7 @@ def from_cli(
             if not model_preset:
                 pre.add_argument(
                     "--model",
+                    dest="model_name",
```
marcuscollins, line 271: I had totally forgotten about this! LOL.

src/sampleworks/utils/guidance_script_arguments.py
```diff
@@ -282,17 +283,21 @@ def from_cli(
                     help="Guidance method",
                 )
             pre_args, _ = pre.parse_known_args(argv)
-            model = model or pre_args.model
+            model_name = model_name or pre_args.model_name
             guidance_type = guidance_type or pre_args.guidance_type
 
+        assert model_name is not None
```
marcuscollins: What's the case for using `assert` here? Can you at least add a descriptive error message?

src/sampleworks/utils/guidance_script_arguments.py
```diff
@@ -374,13 +379,13 @@ def populate_config_for_guidance_type(self, job: JobConfig, args: argparse.Names
             self.model_checkpoint = checkpoint
         elif not getattr(self, "model_checkpoint", None):
             # Auto-resolve from baked-in /checkpoints/ or legacy fallback paths
```
marcuscollins, line 381: Would you please create an issue to cut/clean up backwards compatibility at some point soon-ish? I don't want us to get bogged down in too much of this, it will leave a lot of debt, I think, and make it hard to move forward.

src/sampleworks/utils/guidance_script_arguments.py
```diff
@@ -410,6 +415,13 @@ def as_dict(self) -> dict[str, Any]:
         output["log_path"] = _remap_container_path(str(self.log_path))
         return output
 
+    def __setstate__(self, state: dict[str, Any]) -> None:
```
marcuscollins: This is another backwards compatibility thing that I'm not sure we need, maybe we can discuss this week. I can always checkout the commit I need. (This is one reason I want to make sure we have git commit hashes written in to our job metadata)

src/sampleworks/utils/guidance_script_arguments.py
```diff
@@ -694,3 +706,10 @@ def as_dict(self) -> dict[str, Any]:
         output["output_dir"] = _remap_container_path(str(self.output_dir))
         output["log_path"] = _remap_container_path(str(self.log_path))
         return output
+
+    def __setstate__(self, state: dict[str, Any]) -> None:
```
marcuscollins: same comment as above.

tests/test_run_grid_search.py
```diff
@@ -0,0 +1,54 @@
+"""Tests for grid-search result serialization."""
+
+import json
+
+from run_grid_search import GridSearchConfig, save_results
+from sampleworks.utils.guidance_script_arguments import JobResult
+
+
+def test_save_results_normalizes_legacy_model_key(tmp_path) -> None:
```
marcuscollins: ditto issue about backwards compatibility.

marcuscollins (changes requested):
Do we also need to update the readme? I guess we can still use --model, so maybe not.

src/sampleworks/utils/guidance_script_arguments.py
manzuoni-astera: Replaced both assertions with an explicit `RuntimeError` carrying the message: `CLI parsing did not resolve a model name and guidance type`. This remains an internal invariant failure, but no longer depends on assertions being enabled and now explains the failure.

src/sampleworks/utils/guidance_script_arguments.py
manzuoni-astera: Agreed on avoiding compatibility debt. I removed the legacy pickle migration entirely, so there is no compatibility shim to track or clean up later.

src/sampleworks/utils/guidance_script_arguments.py
manzuoni-astera: Removed this `GuidanceConfig.__setstate__` compatibility path and its tests. Old queues can be read by checking out their originating commit, as suggested.

src/sampleworks/utils/guidance_script_arguments.py
manzuoni-astera: Removed the equivalent `JobResult.__setstate__` compatibility path as well.

tests/test_run_grid_search.py
manzuoni-astera: Removed legacy `results.json` normalization and deleted this compatibility-only test. New results consistently use `model_name`.

manzuoni-astera:
Addressed the review in `fa6ac7a`:

- replaced the assertions with a descriptive `RuntimeError`
- removed all legacy pickle and `results.json` migration shims and their compatibility-only tests
- kept README examples unchanged because the public `--model` flag is unchanged

Focused CLI/config/metadata tests pass (`110 passed`) and targeted type/lint checks are clean. Ready for re-review.

marcuscollins (approved):
LGTM thanks for the updates.

Issue #213, opened 2026-04-10
Evaluation scripts should get the locations of maps or input pdb files used from the results metadata, not from a config file or command line arguments


marcuscollins:
Right now, all of our evaluation scripts require a path to an input data directory, which must contain the maps and input CIF files used for ensemble generation, and a configuration file which, among other things, points to relative paths in that directory which contain the specific map and input files for each protein. Rows of this configuration file look like
`23,5MHX,chain A and resi 158-167,5MHX_single_001_density_input.cif,5MHX_uniform_1.00A.ccp4,processed/5MHX,1.0` 
(in this example there is only one atom selection string; there are usually several, semicolon-separated selections). This line defines a ProteinConfig object (https://github.com/diff-use/sampleworks/blob/main/src/sampleworks/eval/grid_search_eval_utils.py#L224). Assuming the input data directory is "/data/inputs", the evaluation scripts look for the input cif file at `/data/inputs/5MHX_single_001_density_input.cif` and the input maps used for guidance in `/data/inputs/processed/5MHX/` (an additional pattern is used to locate the exact map, see https://github.com/diff-use/sampleworks/blob/main/scripts/eval/rscc_grid_search_script.py#L98

Rather than constructing these paths after the fact, we should obtain them from the ensemble generation trial metadata directly. This data is stored in a file `job_metadata.json` in each output directory, and will soon be incorporated directly into our output CIF files https://github.com/diff-use/sampleworks/pull/209. We should extract the required paths directly from those locations and use them, rather than trying to reconstruct them ad hoc after the fact. 

Note that this depends on the paths in the metadata being actual paths on the working filesystem. Since our jobs are usually run inside Docker containers, the paths stored in the metadata today are ephemeral container paths, not the final locations of files, which depend on what external volumes are mounted to the container. See https://github.com/diff-use/sampleworks/issues/210

manzuoni-astera:
First implementation slice is up in #321: trial discovery now loads `job_metadata.json`, prefers recorded run metadata over directory reconstruction, and carries the exact recorded structure path, density path, and resolution on each `Trial`. Historical path parsing remains as fallback. Follow-up evaluator migrations are still needed before this issue can close.

xraymemory:
[...]

PR #321, opened 2026-07-14
feat(eval): discover trials from job metadata

manzuoni-astera:
## Summary
- load `job_metadata.json` while discovering completed grid-search trials
- prefer recorded model, method, scaler, ensemble, guidance settings, and explicit altloc occupancies over directory-name reconstruction
- record `altloc_occupancies` in new job metadata files
- carry recorded input structure, density-map, and resolution paths on each `Trial`
- make RSCC evaluation use those recorded map, reference-structure, and resolution values
- centralize metadata resolution and preserve path/config parsing as a historical fallback
- accept both current `model` and forthcoming `model_name` metadata schemas

Implementation work for #213. Other evaluators can now migrate to the same `Trial` fields; the legacy input/config arguments remain available for selections and historical runs.

## Validation
- `pytest tests/eval/test_grid_search_trial_metadata.py tests/eval/test_eval_dataclasses.py tests/eval/test_rscc_grid_search_script.py tests/utils/test_guidance_script_utils.py -q` (34 passed)
- targeted `ty check` passes
- `ruff check` and `ruff format --check` on touched files
- `git diff --check`


src/sampleworks/eval/grid_search_eval_utils.py
```diff
@@ -149,28 +178,43 @@ def scan_grid_search_results(
...
-        if params["guidance_weight"] is not None:
-            guidance_weight = float(params["guidance_weight"])
-        gd_steps = int(params["gd_steps"]) if params["gd_steps"] is not None else None
+        guidance_weight_value = metadata.get(
+            "guidance_weight", metadata.get("step_size", params["guidance_weight"])
+        )
+        guidance_weight = (
+            float(str(guidance_weight_value)) if guidance_weight_value is not None else None
+        )
+        gd_steps_value = metadata.get("num_gd_steps", params["gd_steps"])
```
marcuscollins: This is all fine, but very repetitive. Might make sense to encapsulate some function that resolves each attribute from metadata, so you could just call it once for each attribute you need.

src/sampleworks/eval/grid_search_eval_utils.py
```diff
@@ -149,28 +178,43 @@ def scan_grid_search_results(
...
+        # that predate job_metadata.json.
         # Expected structure: .../protein_dir/model_dir/scaler_dir/trial_dir/refined.cif
         trial_dir = current_directory
         scaler_dir = trial_dir.parent
         model_dir = scaler_dir.parent
         protein_dir = model_dir.parent
 
-        protein, altloc_occupancies = extract_protein_and_occupancy(protein_dir.name)
-        method, model = get_method_and_model_name(model_dir.name)
+        path_protein, altloc_occupancies = extract_protein_and_occupancy(protein_dir.name)
```
marcuscollins: are we still getting `altloc_occupancies` only from the directory name? Is it missing from the metadata? This will probably need to be fixed for this issue to really be completely solved. Let's track that down and and at least make a P0 issue if not fix it in this PR.

tests/eval/test_grid_search_trial_metadata.py
```diff
@@ -0,0 +1,46 @@
...
+        "density": "/inputs/1abc.ccp4",
+        "resolution": 1.8,
+    }
+    (trial_dir / "job_metadata.json").write_text(json.dumps(metadata))
+
+    trials = scan_grid_search_results(trial_dir, current_depth=4, target_depth=4)
+
+    assert len(trials) == 1
+    trial = trials[0]
+    assert trial.protein == "1ABC"
```
marcuscollins, line 32: same comment as above that this is missing the altloc occupancies.

marcuscollins (changes requested):
There's one issue about the `altloc_occupancies` that we should chase down. Otherwise LGTM.

src/sampleworks/eval/grid_search_eval_utils.py
manzuoni-astera: Refactored this into shared `_metadata_value`, `_metadata_float`, `_metadata_int`, and `_metadata_occupancies` helpers. Each field now has one resolution call, null/empty values fall back cleanly, numeric conversion is defensive, and invalid metadata logs a warning before using the path-derived fallback.

src/sampleworks/eval/grid_search_eval_utils.py
manzuoni-astera: Fixed in the latest commit. `_write_job_metadata` now records an explicit `altloc_occupancies` object derived from the run’s protein identifier. Trial discovery prefers that metadata and only falls back to the directory name for historical runs. This removes the directory-name dependency for newly generated results.

tests/eval/test_grid_search_trial_metadata.py
manzuoni-astera: Updated the test metadata to include explicit occupancies that differ from the directory (`0.25/0.75` versus `0.5/0.5`) and assert that the metadata values win. I also added a metadata-writing test and a null/empty fallback regression test.

manzuoni-astera:
Addressed the requested changes in `ce07d23`:

- new runs now write explicit `altloc_occupancies` into `job_metadata.json`
- trial discovery prefers those occupancies and uses directory parsing only for historical fallback
- metadata resolution is centralized in typed helpers, including null/empty and invalid-number fallbacks
- tests prove metadata occupancies override different directory occupancies and cover metadata writing

Focused evaluation/metadata tests pass (`34 passed`) and targeted type/lint checks are clean. Ready for re-review.

marcuscollins (approved):
Thanks for the changes. LGTM

PR #322, opened 2026-07-19
feat(eval): generate synthetic sf mtz with Fprotein and Ftotal and correct cif saving

DorisMai:
The older PR #272 is broken into 3 smaller PRs for (1) improve synthetic data generation, (2) refactor reward tests, and (3) add structure factor reward. This PR addresses (1).

## Changes relative to main branch
1. Re-implement `atomarray_to_gemmi` to fix the cif→atomarray→gemmi→cif→atomarray… round trip.
2. Write both Fprotein and Ftotal sets into one MTZ with dtype-resolved column selection.
3. Add a helper function`resolve_mtz_column` in `eval/synthetic_utils.py` to support mtz writing and parsing.
4. Bump `sfcalculator-torch` to `>=0.3.3` (0.3.2 computed Ftotal incorrectly); regenerate `pixi.lock`.

## Changes relative to the previous PR #272
Most edits were addressing reviewer comments, except for the reimplementation of `atomarray_to_gemmi` that was not exactly instructed by previous comments.

## Summary by CodeRabbit

* **New Features**
  * Synthetic MTZ generation can now emit both protein-only and total structure-factor columns into a single file when solvent simulation and scaling are enabled.
  * Added consistent per-set amplitude/uncertainty/phase column naming and improved MTZ column selection by data type.
  * Improved atomic structure conversion for preserving chain, residue, altloc, and annotation details.

* **Bug Fixes**
  * Prevented residue IDs shared across chain boundaries from being incorrectly merged.
  * Added validation for empty inputs and ambiguous or invalid MTZ column selections.

* **Tests**
  * Expanded coverage for structure round-tripping, chain/residue handling, empty input errors, and MTZ column resolution logic.

### DIFF
--- a/src/sampleworks/synthetic/generate_synthetic_sf.py
+++ b/src/sampleworks/synthetic/generate_synthetic_sf.py
@@ -196,10 +196,10 @@ def process_amplitudes_to_dataset(
         SFcalculator instance
     structure_factor_columns: dict[str, str]
         Mapping of ``label -> SFcalculator attribute``. Label must match the pattern of
-        SFcalculator attribute ``F{label}_asu`` (e.g. ``protein`` ->  ``Fprotein_asu``). 
-        One structure-factor set (``F{label}``/``SIGF{label}``/``PHIF{label}``) is 
+        SFcalculator attribute ``F{label}_asu`` (e.g. ``protein`` ->  ``Fprotein_asu``).
+        One structure-factor set (``F{label}``/``SIGF{label}``/``PHIF{label}``) is
         emitted per label, and multiple labels are merged into one MTZ sharing the same
-        HKL list (``miller_index_column``). For example, ``{"protein": "Fprotein_asu", 
+        HKL list (``miller_index_column``). For example, ``{"protein": "Fprotein_asu",
         "total": "Ftotal_asu"}`` produces ``Fprotein``/``SIGFprotein``/``PHIFprotein``/
         ``Ftotal``/``SIGFtotal``/``PHIFtotal`` and optionallyan R-free flag column.
     test_fraction: float
--- a/src/sampleworks/synthetic/generate_synthetic_sf.py
+++ b/src/sampleworks/synthetic/generate_synthetic_sf.py
@@ -1,9 +1,11 @@
 """Generate synthetic structure factor amplitudes via SFcalculator-torch.
 
-Produces an MTZ file of |Fmodel| (or |Fprotein| if not simulate solvent and
-scale) for each input PDB/mmCIF structure. The MTZ file has dummy values for
-SIGFP and optionally R-free flag column. Each structure can be optionally
-overridden with unit cell, space group, atom selection, and occupancy.
+For each input PDB/mmCIF structure, produces an MTZ file of protein structure factors
+only, or when ``--simulate-solvent-and-scale`` is set, both the protein and total sets
+(Fprotein/SIGFprotein/PHIFprotein and Ftotal/SIGFtotal/PHIFtotal) in the same
+MTZ. The MTZ file has dummy values for the SIGF column(s) and optionally an R-free flag
+column. Each structure can be optionally overridden with unit cell, space group, atom
+selection, and occupancy.
 """
 
 import argparse
@@ -25,6 +27,7 @@ from SFC_Torch.io import PDBParser
 from sampleworks.synthetic.synthetic_utils import (
     atomarray_to_gemmi,
     load_structure_for_synthetic_reward,
+    resolve_mtz_column,
     resolve_parallel_jobs,
     validate_occupancy_values,
 )
@@ -127,12 +130,60 @@ class BatchRowForMTZ:
         )
 
 
+def _build_rs_dataset_for_one_label(
+    sfc: SFcalculator,
+    label: str,
+    structure_factor_column: str,
+    miller_index_column: str,
+    sigma_f_scale: float,
+) -> rs.DataSet:
+    """Build an rs.DataSet that contains only one set of F / SIGF / PHIF columns.
+
+    ``sfc.prepare_dataset`` returns an amplitude column and a phase column (degrees)
+    for the given ``structure_factor_column`` attribute. We auto-detect those by MTZ
+    dtype (rather than assuming the unexposed ``FMODEL`` / ``PHIFMODEL`` names),
+    rename them to ``F{label}`` / ``PHIF{label}``, and synthesize a ``SIGF{label}``
+    column so several structure-factor sets (e.g. protein and total) can coexist in
+    one MTZ.
+
+    Parameters
+    ----------
+    sfc : SFcalculator
+        Structure-factor calculator holding the requested ASU amplitudes and phases.
+    label : str
+        Output column-label suffix, currenly only ``"protein"`` or ``"total"``, but
+        can also be solvent (mask) for debugging purposes.
+    structure_factor_column : str
+        SFcalculator attribute name passed to ``prepare_dataset`` (the amplitudes to emit).
+    miller_index_column : str
+        SFcalculator attribute name holding the Miller indices for ``prepare_dataset``.
+    sigma_f_scale : float
+        Multiplier used to synthesize the ``SIGF{label}`` column from the amplitudes.
+
+    Returns
+    -------
+    rs.DataSet
+        Dataset with columns ``F{label}``, ``SIGF{label}``, ``PHIF{label}`` (in that order).
+    """
+    dataset: rs.DataSet = sfc.prepare_dataset(miller_index_column, structure_factor_column)
+    amplitude_column = resolve_mtz_column(dataset, rs.StructureFactorAmplitudeDtype())
+    phase_column = resolve_mtz_column(dataset, rs.PhaseDtype())
+    logger.debug(
+        f"Auto-detected amplitude column: {amplitude_column}, "
+        f"phase column: {phase_column} for {label}"
+    )
+    f_col, phi_col, sig_col = f"F{label}", f"PHIF{label}", f"SIGF{label}"
+    dataset = dataset.rename(columns={amplitude_column: f_col, phase_column: phi_col})
+    dataset[sig_col] = (dataset[f_col] * sigma_f_scale).astype(rs.StandardDeviationDtype())
+    return dataset[[f_col, sig_col, phi_col]]
+
+
 def process_amplitudes_to_dataset(
     sfc: SFcalculator,
+    structure_factor_columns: dict[str, str],
     test_fraction: float = 0.05,
     seed: int | None = None,
     miller_index_column: str = "Hasu_array",
-    structure_factor_column: str = "Ftotal_asu",
     ccp4_convention: bool = False,
     sigma_f_scale: float = 0.2,
     output_path: Path | None = None,
@@ -143,14 +194,20 @@ def process_amplitudes_to_dataset(
     ----------
     sfc: SFcalculator
         SFcalculator instance
+    structure_factor_columns: dict[str, str]
+        Mapping of ``label -> SFcalculator attribute``. Label must match the pattern of
+        SFcalculator attribute ``F{label}_asu`` (e.g. ``protein`` ->  ``Fprotein_asu``). 
+        One structure-factor set (``F{label}``/``SIGF{label}``/``PHIF{label}``) is 
+        emitted per label, and multiple labels are merged into one MTZ sharing the same
+        HKL list (``miller_index_column``). For example, ``{"protein": "Fprotein_asu", 
+        "total": "Ftotal_asu"}`` produces ``Fprotein``/``SIGFprotein``/``PHIFprotein``/
+        ``Ftotal``/``SIGFtotal``/``PHIFtotal`` and optionallyan R-free flag column.
     test_fraction: float
         Fraction of reflections to mark as R-free test set (0 disables)
     seed: int | None
         Optional seed for reproducible R-free flag assignment
     miller_index_column: str
         Attribute name in SFcalculator for hkl indices
-    structure_factor_column: str
-        Attribute name in SFcalculator for structure factors
     ccp4_convention: bool
         If True, use CCP4 convention for R-free flag assignment. Default
         is False, which uses Phenix convention (1 = test, 0 = working).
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -270,3 +270,9 @@ unresolved-import = "ignore"
 unsupported-operator = "warn"
 unused-ignore-comment = "warn"
 unused-type-ignore-comment = "warn"
+
+[tool.ty.terminal]
+# Newer ty defaults error-on-warning=true, which would fail CI on the warn-level
+# rules above; keep it false so ty runs in CI without blocking PRs while the team
+# fixes them.
+error-on-warning = false
```
marcuscollins: Is this not fixed somewhere else?

tests/synthetic/test_generate_synthetic_sf.py
```diff
@@ -121,6 +129,83 @@ def test_fprotein_matches_direct_gemmi(
         f_direct = _compute_fprotein(stripped_gemmi, device)
         np.testing.assert_allclose(np.abs(f_atomarray), np.abs(f_direct), atol=1e-3)
 
+    def test_saved_structure_round_trips_annotations(
```
marcuscollins, line 168: for purposes of making this test really clean, I would not use `gemmi_structure_from_atomarray` here, and instead call `atomarray_to_gemmi` here directly. That way someone else can't change the `gemmi_structure_from_atomarray` fixture in some way that makes this test pass but doesn't actually do the round-trip test we want.

tests/eval/test_generate_synthetic_sf.py
```diff
@@ -121,6 +129,83 @@ def test_fprotein_matches_direct_gemmi(
         f_direct = _compute_fprotein(stripped_gemmi, device)
         np.testing.assert_allclose(np.abs(f_atomarray), np.abs(f_direct), atol=1e-3)
 
+    def test_saved_structure_round_trips_annotations(
+        self, gemmi_structure_from_atomarray, stripped_atom_array, tmp_path
+    ):
+        """Every per-atom field must survive atomarray_to_gemmi --> cif --> atomarray.
+
+        One reload guarding each column a write can silently drop/garble. Scattering-physics
```
marcuscollins: I don't understand this first sentence. DYM "Test that reloading Gemmi's CIF output returns the original atom array used to construct the Gemmi object. This ensures we don't corrupt or lose information in building the Gemmi object that would affect our structure factor calculations." ?

tests/eval/test_generate_synthetic_sf.py
```diff
@@ -121,6 +129,83 @@ def test_fprotein_matches_direct_gemmi(
...
+        out = tmp_path / "saved.cif"
+        gemmi_structure_from_atomarray.make_mmcif_document().write_file(str(out))
+
+        loaded = load_structure_with_altlocs(out)
+        ref = stripped_atom_array
+
+        assert len(loaded) == len(ref)
+
+        # topology / matching labels (exact)
+        assert np.array_equal(loaded.chain_id, ref.chain_id)
```
marcuscollins: I suggest iterating over the results of `loaded.get_annotation_categories()` (name might be wrong, but there's some method like that). That way we can be sure we test everything.

tests/synthetic/test_generate_synthetic_sf.py
```diff
@@ -121,6 +129,83 @@ def test_fprotein_matches_direct_gemmi(
...
+        assert np.array_equal(loaded.chain_id, ref.chain_id)
+        assert np.array_equal(loaded.res_id, ref.res_id)
+        assert set(np.unique(loaded.res_id)) != {-1}  # not collapsed to the degenerate -1
+        assert np.array_equal(loaded.res_name, ref.res_name)
+        assert np.array_equal(loaded.atom_name, ref.atom_name)
+
+        # scattering-physics fields (element exact; floats within mmCIF write precision)
+        assert np.array_equal(loaded.element, ref.element)
+        assert len(np.unique(loaded.element)) > 1  # not collapsed to a single/blank symbol
+        np.testing.assert_allclose(loaded.coord, ref.coord, atol=1e-3)
```
marcuscollins, line 222: `.coord` might be special and not included in the output of `get_annotation_categories()`.

tests/eval/test_generate_synthetic_sf.py
```diff
@@ -121,6 +129,83 @@ def test_fprotein_matches_direct_gemmi(
...
+        assert np.array_equal(loaded.atom_name, ref.atom_name)
+
+        # scattering-physics fields (element exact; floats within mmCIF write precision)
+        assert np.array_equal(loaded.element, ref.element)
+        assert len(np.unique(loaded.element)) > 1  # not collapsed to a single/blank symbol
+        np.testing.assert_allclose(loaded.coord, ref.coord, atol=1e-3)
+        np.testing.assert_allclose(loaded.b_factor, ref.b_factor, atol=1e-2)
+        np.testing.assert_allclose(loaded.occupancy, ref.occupancy, atol=1e-2)
+
+        # altloc: same id set, and same per-id atom counts
```
marcuscollins: Do these tests first. In fact, you might even include a test that ref has altloc id just as an extra check on that code.

tests/eval/test_generate_synthetic_sf.py
```diff
@@ -121,6 +129,83 @@ def test_fprotein_matches_direct_gemmi(
...
+        assert find_all_altloc_ids(loaded) == expected_altloc_ids
+        for altloc_id in expected_altloc_ids:
+            assert np.count_nonzero(loaded.altloc_id == altloc_id) == np.count_nonzero(
+                ref.altloc_id == altloc_id
+            )
+
+    def test_multichain_shared_res_ids_survive_round_trip(self, tmp_path):
+        """Two chains that share res_ids must not be merged across the chain boundary.
+
+        array2hier keyed residue boundaries on res_id alone, so a chain boundary where both
```
marcuscollins: Does it still do this?

tests/eval/test_generate_synthetic_sf.py
```diff
@@ -121,6 +129,83 @@ def test_fprotein_matches_direct_gemmi(
...
+        arr.chain_id = np.array(["A", "A", "B", "B"])
+        arr.res_id = np.array([1, 2, 1, 2])
+        arr.res_name = np.array(["ALA", "GLY", "ALA", "GLY"])
+        arr.atom_name = np.array(["CA", "CA", "CA", "CA"])
+        arr.element = np.array(["C", "C", "C", "C"])
+        arr.set_annotation("b_factor", np.full(n, 20.0))
+        arr.set_annotation("occupancy", np.ones(n))
+
+        out = tmp_path / "multichain.cif"
+        atomarray_to_gemmi(arr).make_mmcif_document().write_file(str(out))
```
marcuscollins: Is there anyway to test this without calling `atomarray_to_gemmi`? The issue is that if that method breaks, you won't know here whether the failure is due to that method, or one of the CIF methods from gemmi. I guess you're testing `atomarray_to_gemmi` here?

tests/eval/test_generate_synthetic_sf.py
```diff
@@ -121,6 +129,83 @@ def test_fprotein_matches_direct_gemmi(
...
+
+        One reload guarding each column a write can silently drop/garble. Scattering-physics
+        fields (coords, element, b_factor, occupancy) would corrupt structure factors without
+        raising if lost; topology fields (chain_id, res_id, res_name, atom_name) drive residue/
+        altloc matching and reconciliation. res_id and label_seq_id are the field that regressed:
+        array2hier set only the author seqid, so label_seq_id wrote as "." and atomworks (label
+        scheme) collapsed every residue to -1 on re-read.
+        """
+        out = tmp_path / "saved.cif"
+        gemmi_structure_from_atomarray.make_mmcif_document().write_file(str(out))
```
marcuscollins: I hate to ask more, but is there any easy way to check that everything is correct in the gemmi object itself?

tests/eval/test_generate_synthetic_sf.py
```diff
@@ -164,3 +249,63 @@ def test_fprotein_changes_with_occupancy(self, stripped_atom_array, stripped_gem
         )
 
         assert not np.allclose(np.abs(f_uniform), np.abs(f_custom), atol=1e-3)
+
+
+def _dataset_with_columns(columns: dict[str, MTZDtype]) -> rs.DataSet:
```
marcuscollins: You could make this a fixture. If you only need it here, then create it with scope=module

src/sampleworks/synthetic/synthetic_utils.py
```diff
@@ -226,3 +286,148 @@ def load_structure_for_synthetic_reward(
...
+    altlocs : list of str
+        Per-atom altloc labels for the whole array (gemmi convention), indexed
+        globally by atom index -- not by a residue-local offset.
+
+    Returns
+    -------
+    gemmi.Residue
+        Residue populated with the atoms ``atom_array[start_idx:stop_idx]``.
+    """
+    res_id = int(atom_array.res_id[start_idx])
```
marcuscollins, line 360: Should you check that they're all the same? It seems like good defensive programming to make sure you don't make a residue out of atoms that aren't already part of the same residue.

marcuscollins: and same name, and no atoms are repeated, etc... all of which should have tests too.

marcuscollins: It would be okay to just make an issue for these and do later.

src/sampleworks/eval/synthetic_utils.py
```diff
@@ -226,3 +286,148 @@ def load_structure_for_synthetic_reward(
...
+        One ``(start_idx, stop_idx)`` per residue, covering the atoms
+        ``atom_array[start_idx:stop_idx]`` that share the same ``(chain_id, res_id)``.
+    """
+    if len(atom_array) == 0:
+        return []
+    chain_id, res_id = atom_array.chain_id, atom_array.res_id
+    # boundary shows where a new residue begins (i.e., chain or res_id changed).
+    boundary = (chain_id[1:] != chain_id[:-1]) | (res_id[1:] != res_id[:-1])
+    start_indices = [0, *(np.flatnonzero(boundary) + 1).tolist(), len(atom_array)]
+    return list(zip(start_indices[:-1], start_indices[1:]))
```
marcuscollins: Rather than instantiate the list, you could use `yield` in a for loop.

src/sampleworks/synthetic/generate_synthetic_sf.py
```diff
@@ -215,18 +220,27 @@ def process_amplitudes_to_dataset(
...
+    # All sets share the same HKL index, so this can be just a column-wise concat.
+    dataset: rs.DataSet | None = None
+    for label, attribute in structure_factor_columns.items():
+        ds = _build_rs_dataset_for_one_label(
+            sfc, label, attribute, miller_index_column, sigma_f_scale
+        )
+        if dataset is None:
+            dataset = ds
+        else:
+            dataset[ds.columns] = ds  # graft the remaining columns on (index-aligned)
```
marcuscollins, line 252: should you check that the column name isn't already populated?

src/sampleworks/eval/generate_synthetic_sf.py
```diff
@@ -215,18 +220,27 @@ def process_amplitudes_to_dataset(
...
-    dataset: rs.DataSet = sfc.prepare_dataset(miller_index_column, structure_factor_column)
-    # assumes the first detected column of dtype F is the structure factor amplitude column
-    # avoids hardcoding unexposed column name "FMODEL" from sfc.prepare_dataset().
-    structure_factor_amplitude_column = dataset.select_mtzdtype(
-        rs.StructureFactorAmplitudeDtype()
-    ).columns[0]
-    sigma_f_column = f"SIG{structure_factor_amplitude_column}"
-    dataset[sigma_f_column] = dataset[structure_factor_amplitude_column] * sigma_f_scale
-    dataset[sigma_f_column] = dataset[sigma_f_column].astype(rs.StandardDeviationDtype())
+    # One dataset per label: for now just "protein" and "total", but it can generalizes
```
marcuscollins: Do you check that these values are the only ones used?

marcuscollins (changes requested):
There's a few things in here that should be buttoned up to make it a little more robust. But basically looks solid.

pyproject.toml
DorisMai: changes reverted

tests/eval/test_generate_synthetic_sf.py
DorisMai: moved to be within `TestResolveMtzColumn` as 3 separate named fixtures because they are only consumed by that test class and not re-used enough to be one factory fixture.

tests/synthetic/test_generate_synthetic_sf.py
DorisMai: good point, changed

src/sampleworks/eval/synthetic_utils.py
DorisMai: good point, done

src/sampleworks/synthetic/generate_synthetic_sf.py
DorisMai: added collision check

tests/eval/test_generate_synthetic_sf.py
DorisMai: no, docstring updated

tests/eval/test_generate_synthetic_sf.py
DorisMai: yes, rephrased

tests/synthetic/test_generate_synthetic_sf.py
DorisMai: indeed, asserted separately

tests/eval/test_generate_synthetic_sf.py
DorisMai: tests moved up and guarded for ref altloc id

tests/eval/test_generate_synthetic_sf.py
DorisMai: I didn't end up testing all fields, because there can be extra fields that we don't care (`is_polymer` field only exists from the parsing cif but not from pdb, and the test fixture is in pdb whereas my gemmi saves as cif). There are also fields like `ins_code` that is still under research (issue #306 ). Therefore I decide to check that the following important fields exists in both `loaded` and `ref` atomarrays and only check those fields identical in round trip: 
```
        important_annotations = {
            "chain_id", "res_id", "res_name", "atom_name", "element", "hetero",
            "b_factor", "occupancy", "altloc_id",
        }
```

src/sampleworks/eval/generate_synthetic_sf.py
DorisMai: there is no hard constraint on what the `label` has to be, the user just needs to specify the same column name derived from it when providing the mtz file for structure factor reward. As long as it's consistent, as far as functionality concerns, `label="foo"` and a provided column name `"Ffoo"` should work fine. When I output mtz file using this script, I use "protein" and "total" as the labels. Ideally I might want to declare constants somewhere to encourage the consistency between synthetic scripts and reward function / guidance scripts, but it's not immediately clear to me where best to do this (any risk of circular / bad import direction). For now I just updates the docstrings to describe this.

tests/eval/test_generate_synthetic_sf.py
DorisMai: what kind of checks do you have in mind? I feel that looping through annotations and coords should be a pretty thorough check as far as functionality concerns?

tests/eval/test_generate_synthetic_sf.py
DorisMai: It is intended to test `atomarray_to_gemmi`. I modified it to directly compare the fields without saving to cif.

src/sampleworks/synthetic/synthetic_utils.py
DorisMai: issue added: https://github.com/diff-use/sampleworks/issues/340

marcuscollins (approved):
Thanks for the updates. LGTM.

PR #323, opened 2026-07-19
test(rewards): extract reward-agnostic contract tests and scope fixtures

DorisMai:
The older PR #272 is broken into 3 smaller PRs for (1) improve synthetic data generation, (2) refactor reward tests, and (3) add structure factor reward. This PR addresses (2).

## Changes relative to main branch
1. Extract reward-agnostic contract tests into a shared `test_reward_function_contract.py` from `test_real_space_density_reward.py`.
2. Move reward fixtures from `tests/conftest.py` into a session scoped `tests/rewards/conftest.py`.
3. Add `tests/rewards/reward_input_helpers.py` for shared reward-input construction.

## Changes relative to the previous PR #272
No changes were requested or comments made for these files in the previous PR, except for the suggestion on reward fixture scope.

## Summary by CodeRabbit

* **Tests**
  * Added GPU-marked, reward-agnostic contract tests covering protocol conformance, determinism, output correctness, ranking behavior, gradient flow, batching, loss optimization improvement, and numerical stability.
  * Refocused real-space density reward tests to edge cases and input-to-reward-input validation, removing substantial prior coverage.
  * Improved reward test infrastructure with suite-scoped fixtures for parsing the CIF, building the reward function, and extracting coordinates, plus shared helpers for scattering indices and reward inputs.
  * Removed overlapping session-scoped fixtures from the broader test setup.


tests/rewards/test_reward_function_contract.py
```diff
@@ -0,0 +1,267 @@
...
+# carved 1vme (matches its .ccp4 map frame). New rewards register here alongside a matching
+# entry in `_LOSS_THRESHOLDS`.
+_REWARD_BUNDLES = {
+    "real_space": ("test_coordinates_1vme", "reward_function_1vme"),
+}
+
+# Absolute loss bar for the TRUE structure, per reward. RealSpace's loss is MSE on
+# normalized density (sigma units), so a wrong/random model scores ~O(1) while the true
+# model scores ~0.
+_LOSS_THRESHOLDS = {
```
k-chrispens, line 75: We should probably calculate what the actual value would be here, I know I was lazy before and didn't do it, but it would be helpful for the specific 1VME test case we use here.

k-chrispens: It should be very low, much less than 1!

k-chrispens (approved):
Generally looks good to me, I think we should have it somewhere on the roadmap to add some better tests here, but not quite sure where that fits in these days with all the loss function talk we had back in June!

tests/rewards/test_reward_function_contract.py
DorisMai: indeed, I tightened the bound to be 0.01.

PR #334, opened 2026-07-22
ci: pin pixi-version in setup-pixi for deterministic lock validation

xraymemory:
## Problem

CI validates the lockfile with `pixi install --locked` (strict — it fails rather than re-solving if `pixi.lock` doesn't match `pyproject.toml`). But `setup-pixi@v0.8.8` pins the *action*, not the pixi *binary* — it installs whatever pixi is "latest" at run time.

So the pixi that generates a lock locally can differ from the pixi that validates it in CI, and two versions can disagree on the manifest hash. The symptom is intermittent `lock file not up-to-date with the workspace` failures on PRs that didn't touch dependencies (seen on #300), and CI that can start failing with no code change when a new pixi ships.

## Fix

Pin `pixi-version: v0.73.0` on every `setup-pixi` step (ci.yml x3, gpu-tests.yml). v0.73.0 is the version that generated the current green lock on `main`. Lock validation is now deterministic and matches the committed lock's generator.

## Follow-up

Contributors regenerating the lock should use the same pinned pixi (`pixi self-update --version 0.73.0` or install that version), run `pixi lock`, and commit `pyproject.toml` + `pixi.lock` together. Bumping pixi later is a one-line change here plus a relock.

## Summary by CodeRabbit

* **Chores**
  * Standardized the Pixi toolchain version used by automated linting, type checks, CPU tests, and GPU tests.
  * Improved consistency and reproducibility across continuous integration workflows.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,6 +52,7 @@ jobs:
       - name: Install pixi
         uses: prefix-dev/setup-pixi@v0.8.8
         with:
+          pixi-version: v0.73.0
           environments: boltz-dev
 
       - name: Ruff lint
@@ -79,6 +80,7 @@ jobs:
       - name: Install pixi
         uses: prefix-dev/setup-pixi@v0.8.8
         with:
+          pixi-version: v0.73.0
           environments: ${{ matrix.environment }}
 
       - name: Run ty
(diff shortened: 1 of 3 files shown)


marcuscollins (approved):
I'm curious how the relock workflow is used.

PR #343, opened 2026-07-26
fix(actl): honor RUNTIME_PIXI project dir over baked image env

manzuoni-astera:
## Problem

Under `RUNTIME_PIXI=1`, jobs on a branch that adds a pixi env (e.g. `protpardelle`) fail with `unknown environment 'protpardelle'` — the banner claims the synced checkout but the job runs against `/app`, whose manifest lacks the env.

## Cause

The image bakes `SAMPLEWORKS_PIXI_PROJECT_DIR=/app`. The wrappers applied their `RUNTIME_PIXI` overrides only via `${VAR:-default}` / `[[ -z ]]` — no-ops once the var is set — so `pixi_project_dir` stayed `/app`. And the launcher is baked into `/usr/local/bin` as a frozen copy, so a wrapper-only fix needs a rebuild.

## Fix

1. `run_experiments` / `run_analysis`: assign every value unconditionally in the runtime-pixi path so baked env vars can't defeat the escape hatch.
2. `runner.py` (loaded from the synced checkout via `PYTHONPATH`, so it needs no rebuild): `_pixi_project_dir()` re-derives the checkout under `RUNTIME_PIXI` and prefers it when its manifest differs from the baked image.

## Verification

New `runner.py` tests, plus an end-to-end check on the real pixi-with-checkpoints image: the fixed runner resolves to `/home/dev/workspace` under `RUNTIME_PIXI` (was `/app`), and `/app` otherwise.

### DIFF
--- a/run_analysis
+++ b/run_analysis
@@ -232,18 +232,15 @@ runner_env="${SAMPLEWORKS_ANALYSIS_RUNNER_ENV:-analysis}"
 if truthy_env RUNTIME_PIXI; then
     export SAMPLEWORKS_ALLOW_RUNTIME_PIXI=1
 fi
-pixi_project_dir="${SAMPLEWORKS_PIXI_PROJECT_DIR:-}"
-if [[ -z "$pixi_project_dir" ]]; then
-    if ! pixi_inputs_match_image /app "$repo_root" && truthy_env SAMPLEWORKS_ALLOW_RUNTIME_PIXI; then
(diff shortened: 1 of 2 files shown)


marcuscollins (commented):
LGTM. The only thing I'd say is that the logic in these scripts is identical--should they be one script so we only have to update things in one place?

marcuscollins (approved):

PR #345, opened 2026-07-27
fix(grid-search): run protpardelle jobs in their pixi env

manzuoni-astera:
## What breaks

Every protpardelle job in `run_experiments protpardelle` dies instantly:

```
ERROR | run_grid_search:301 - Job failed with exception: [Errno 2] No such file or directory: '.../wjq_139929586767296.results.pkl'
```

`get_pixi_env()` still had a hardcoded boltz/protenix/rf3 branch, so `protpardelle` fell through to `raise ValueError("Unknown model: protpardelle")`. That call is the first statement of `run_guidance_queue_script`, so no worker ever reached `subprocess.run` — which is why there are no `wjq_*.log` files and no `Running worker N: ...` line in the output.

`StructurePredictor.PROTPARDELLE`, the `protpardelle` pixi environment in `pyproject.toml`, and `env = "protpardelle"` in `experiments/protpardelle.toml` were all already in place. Only this function was missed.

## Why the error was unreadable

The `ValueError` is raised inside a `ProcessPoolExecutor` child, so it lands on the future. The collection loop never inspected the future — it went straight to opening `wjq_*.results.pkl`, which the dead worker never wrote, and the surrounding `except Exception` reported *that* `FileNotFoundError`. The real exception was discarded.

## Changes

- `get_pixi_env()` now resolves through a `MODEL_PIXI_ENVS` mapping keyed on `StructurePredictor`, so a newly added predictor cannot silently miss an environment.
- The result collector checks `completed.exception()` first and logs it with `log.opt(exception=...)`, reports a nonzero subprocess exit code, and points at the `wjq_*.log` path when the results pickle is genuinely missing.

## Tests

Two added to `tests/test_run_grid_search.py`: every `StructurePredictor` resolves to a name in `VALID_PIXI_ENVS`, and an unknown model still raises with the valid options listed.

**These have not been run** — `sampleworks` is not installable on the machine this was written on. Please run `pixi run -e protpardelle-dev tests -- -k grid_search` before merging.

## Follow-up, not in this PR

Once this lands, the `Running worker N: [...]` line will print the real command. If it shows `pixi run -e protpardelle ...` instead of a direct `.pixi/envs/protpardelle/bin/python`, the image has no baked protpardelle env and `get_pixi_env_python()` is falling back to a runtime solve on shared storage — works, but slow. That function hardcodes `/app` and does not route through `runner._pixi_project_dir()`, which is the one taught to honor `RUNTIME_PIXI` on `michaelanzuoni/fix-runtime-pixi-project-dir`.

### DIFF
--- a/run_grid_search.py
+++ b/run_grid_search.py
@@ -224,6 +224,25 @@ def build_args_for_process_pool(
     return guidance_config
 
 
+def worker_log_path_for(job_queue_path: str) -> str:
+    """Return the path a worker redirects its subprocess output to.
+
+    Derived in one place so the collector's error messages always name the
+    file ``run_guidance_queue_script`` actually wrote.
+
+    Parameters
+    ----------
+    job_queue_path : str
+        Path to the worker's pickled job queue.
+
+    Returns
+    -------
+    str
+        Sibling ``.log`` path for that worker's subprocess output.
+    """
+    return job_queue_path.replace(".pkl", ".log")
+
+
 def run_grid_search(
     jobs: list[JobConfig],
     gpus: list[str],
@@ -302,7 +321,9 @@ def run_grid_search(
 
         for completed in concurrent.futures.as_completed(futures):  # ty: ignore
             job_queue_path = futures[completed]
-            worker_log_path = job_queue_path.replace(".pkl", ".log")
+            # Same file the worker redirects its subprocess output to; these
+            # messages exist to point the reader at it.
+            worker_log_path = worker_log_path_for(job_queue_path)
             # The worker raises before it can write results, so surface its
             # traceback here rather than reporting the missing results file.
             worker_error = completed.exception()
@@ -400,7 +421,7 @@ def run_guidance_queue_script(
         f"(selected GPU {requested_gpu} via {gpu_source})"
     )
 
-    with open(job_queue_path.replace(".pkl", ".log"), "w") as log_file:
+    with open(worker_log_path_for(job_queue_path), "w") as log_file:
         result = subprocess.run(
             cmd,
             stdout=log_file,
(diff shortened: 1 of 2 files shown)


run_grid_search.py
```diff
@@ -277,27 +301,48 @@ def run_grid_search(
             futures[future] = job_queue_path
 
         for completed in concurrent.futures.as_completed(futures):  # ty: ignore
+            job_queue_path = futures[completed]
+            worker_log_path = job_queue_path.replace(".pkl", ".log")
```
marcuscollins: This log file name is already used in `run_guidance_queue_script()` below, so we should use a different name for this one. Maybe `.replace(".pkl", ".process.log")`?

marcuscollins (changes requested):

run_grid_search.py
manzuoni-astera: It's the same file on purpose — this line isn't opening a second log, it's naming the one `run_guidance_queue_script` redirects the subprocess into, so the reader knows where to look when a worker dies. If I point it at `.process.log` the message names a file nothing ever creates, which is the failure mode this PR is trying to remove.

I think the real problem you're spotting is that both sites spelled out the `.pkl` -> `.log` substitution independently, so it reads as an accidental collision. Extracted it into `worker_log_path_for()` and used it in both places, so there's one derivation and the intent is explicit. Pushed in 17ee936.

Happy to do the literal rename instead if you'd rather have two separate logs — but then something needs to actually write the second one.

manzuoni-astera:
@marcuscollins re-requesting review — my token is read-only on this repo so I can't set the reviewer field, flagging here instead. Log-path comment addressed above.

marcuscollins (approved):

PR #350, opened 2026-07-28
feat(analysis): ship tortoize in the analysis pixi environments

manzuoni-astera:
## Problem

`analyses/external_tools.toml` runs its tortoize job with `env = "analysis"`, but nothing ever put the binary there:

- `scripts/eval/run_and_process_tortoize.py` probes `subprocess.call("tortoize")` on `PATH` and raises `RuntimeError` if it is missing.
- The public `Dockerfile` apt-installs only build tooling, then `pixi install -e {boltz,protenix,rf3,analysis}`.
- `Dockerfile.astera` adds editors and network tools, nothing scientific.
- The `analysis` feature resolves to `[dependency-groups] analysis`, which is pure PyPI with no conda dependencies, so it could not supply a compiled binary.
- `pixi.lock` had zero matches for tortoize.

So `run_analysis external_tools` (and `all`) has never been able to run its tortoize job inside the image — it aborts before doing any work. `analyze_grid_search` with `rscc`/`lddt` is unaffected, which is presumably why this went unnoticed.

## Change

bioconda publishes tortoize 2.0.16 for linux-64, osx-arm64, linux-aarch64 and osx-64, so this adds it as a conda dependency rather than a source build.

It goes in a new `external-tools` feature attached to `analysis` and `analysis-dev`, rather than onto `analysis` directly, for two reasons:

- `analysis` is also a member of `boltz-analysis` and `boltz-osx`. Touching it re-solves those environments' PyPI requirements, which cannot even be done from macOS.
- The bioconda channel stays scoped to this feature, so every other environment keeps solving against conda-forge alone.

The resulting `pixi.lock` diff is **purely additive** — 211 inserted lines, zero removals. No existing package version moves. New packages are tortoize plus its runtime deps (libcifpp, dssp, libboost, libboost-python, pcre2) on the two workspace platforms.

No Dockerfile change is needed: the image already runs `pixi install -e analysis`, so it picks this up on the next build.

## Verification

Ran on osx-arm64 against a real structure in the analysis env:

```
$ pixi run -e analysis which tortoize
.pixi/envs/analysis/bin/tortoize

$ pixi run -e analysis tortoize tests/resources/1vme/1VME_single_001_density_input.cif
ramachandran-z: -0.8303307294845581
torsion-z:      -0.8385160565376282
residues:       399 entries
```

The JSON matches the shape `flatten_residues()` and `get_protein_level_z_scores()` parse — `model.1.residues[].ramachandran["z-score"]`, `.torsion["z-score"]`, `.pdb.seqNum`.

Not verified: the linux-64 solve is from the lockfile only, since I could not install that platform locally. CI's Docker build will exercise it.

## Not in this PR

`phenix.clashscore` has the same problem — `external_tools.toml` describes itself as needing "tortoize and phenix.clashscore", and phenix is likewise absent from the image. It needs a license, so it is not a conda dependency; it should either come from the existing `phenix` actl image or from the planned SBGrid `/programs` mount.

## Summary by CodeRabbit

* **New Features**
  * Added the `tortoize` tool to the analysis environments.
  * The tool is automatically available on the command line when using the analysis environment.

* **Documentation**
  * Updated evaluation documentation with setup and usage instructions.
  * Added guidance on the error shown when `tortoize` is unavailable.

### DIFF
--- a/pixi.lock
+++ b/pixi.lock
@@ -17,10 +17,13 @@ environments:
   analysis:
     channels:
     - url: https://conda.anaconda.org/conda-forge/
+    - url: https://conda.anaconda.org/bioconda/
     indexes:
     - https://pypi.org/simple
     packages:
       osx-arm64:
+      - conda: https://conda.anaconda.org/bioconda/osx-arm64/libcifpp-8.0.1-hda5e58c_5.tar.bz2
+      - conda: https://conda.anaconda.org/bioconda/osx-arm64/tortoize-2.0.16-h639dec6_0.conda
       - conda: https://conda.anaconda.org/conda-forge/noarch/ca-certificates-2026.5.20-hbd8a1cb_0.conda
       - conda: https://conda.anaconda.org/conda-forge/noarch/python_abi-3.12-8_cp312.conda
       - conda: https://conda.anaconda.org/conda-forge/noarch/tzdata-2025c-hc9c84f9_1.conda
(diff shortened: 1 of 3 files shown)


marcuscollins (approved):
Awesome, thanks!

Issue #340, opened 2026-07-23
Validate residue-span homogeneity when converting atomarray to gemmi


DorisMai:
`_build_gemmi_residue` (src/sampleworks/eval/synthetic_utils.py:360) reads per-residue fields (res_id, res_name, seqid, subchain, het_flag, etc.) from `start_idx` alone, assuming every atom in the span is part of the same residue. 

The spans come from `_residue_group_bounds`, which only breaks on (chain_id, res_id) changes. So malformed input could slip through undetected:
- atoms in one span with differing res_name (or chain_id/hetero) — grouping wouldn't catch a res_name change alone
- duplicate atoms within a residue (same (atom_name, altloc))

For defensive programming, `_build_gemmi_residue` (or `_residue_group_bounds`) should:
- [ ] Assert all atoms in a span share res_id, res_name, chain_id, hetero
- [ ] Assert (atom_name, altloc) is unique within the span
- [ ] Tests should be added too

> Should you check that they're all the same? It seems like good defensive programming to make sure you don't make a residue out of atoms that aren't already part of the same residue. 

> and same name, and no atoms are repeated, etc... all of which should have tests too.

 _Originally posted by @marcuscollins in [#322](https://github.com/diff-use/sampleworks/pull/322/changes#r3618024747)_

xraymemory:
Fixed by #351 (merged 2026-09-01). `_prepare_residue_spans` now validates residue-span homogeneity (res_name / chain / hetero / seqid / subchain consistency, duplicate spans) before yielding spans for the gemmi conversion, with tests covering each malformed-input case. The PR said "Address #340" rather than "Fixes", so GitHub didn't auto-close this.

PR #351, opened 2026-07-28
feat(synthetic): validate atom array hierarchy before gemmi conversion

DorisMai:
Address Issue #340 , related to recently merged PR #322

`atomarray_to_gemmi` in generating synthetic SF data assumed contiguous chains/residues and read per-residue fields
(`res_name`, `hetero`, seqid, subchain) from each residue's first atom. For defensive programming, this PR added validation checks during the atomarray --> gemmi conversion againsgt malformed inputs and added tests on the validation checks. The behavior on generating synthetic data from valid files should be unchanged.

Specifcally, `_prepare_residue_spans` is a new function that first validates these assumptions before yielding the residue spans in atom array to build the gemmi Structure. Validation checks include:
- chain ids occupy one contiguous block each
- `(chain_id, res_id)` residue keys occupy one contiguous block
- atoms within a residue agree on `res_name` and `hetero`
- `(atom_name, altloc)` is unique within a residue (gemmi's definition)

Tests: new `TestGemmiHierarchyValidation` covers each rejection case.

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened structure conversion validation to catch malformed chain, residue, and span data earlier.
  * Improved diagnostics for duplicate atoms, inconsistent per-residue fields, and noncontiguous residue or chain sequences.
  * Preserved correct handling of alternate locations and shared residue IDs across chain boundaries.

* **Tests**
  * Expanded coverage for hierarchy validation, including empty inputs, mixed residue attributes, alternate locations, and duplicate atoms.
  * Added round-trip coverage to verify chain information is preserved.

### DIFF
--- a/src/sampleworks/synthetic/synthetic_utils.py
+++ b/src/sampleworks/synthetic/synthetic_utils.py
@@ -417,9 +417,7 @@ def _check_no_repeated_atoms(atom_array: AtomArray, altlocs: list[str]) -> None:
     )
 
 
-def _check_no_repeated_residues(
-    atom_array: AtomArray, residue_span_start_idx: np.ndarray
-) -> None:
+def _check_no_repeated_residues(atom_array: AtomArray, residue_span_start_idx: np.ndarray) -> None:
     """Require each ``(chain_id, res_id)`` key to occupy exactly one contiguous block.
 
     Parameters
--- a/src/sampleworks/synthetic/synthetic_utils.py
+++ b/src/sampleworks/synthetic/synthetic_utils.py
@@ -2,7 +2,9 @@
 
 import math
 import traceback
-from collections.abc import Iterator
+from collections import Counter
+from collections.abc import Hashable, Iterable, Iterator
+from itertools import pairwise
 from pathlib import Path
 
 import gemmi
@@ -24,6 +26,10 @@ from sampleworks.utils.atom_array_utils import (
     load_structure_with_altlocs,
     remove_hydrogens,
 )
+
+
+# How many explicit duplicates to show in an error message when converting atomarray to gemmi.
+MAX_REPORTED_DUPLICATES = 3
 
 
 def resolve_parallel_jobs(device: torch.device | str, n_jobs: int) -> int:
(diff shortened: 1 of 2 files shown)


src/sampleworks/synthetic/synthetic_utils.py
```diff
@@ -310,28 +316,195 @@ def _resolve_altlocs_for_gemmi(atom_array: AtomArray) -> list[str]:
...
+
+    ``res_id`` and ``chain_id`` need no check: they are the grouping key.
 
     Parameters
     ----------
     atom_array : AtomArray
-        Structure whose atoms are grouped into residues. Atoms of a residue are
-        assumed contiguous (true for arrays loaded in file order).
+        Structure to check.
+    residue_boundary_mask : np.ndarray
```
marcuscollins, line 362: Can't this come from the AtomArray itself? Is there a reason to calculate it separately outside this function? If it is computed inside this function, you guard against someone passing an incorrect boundary mask.

src/sampleworks/synthetic/synthetic_utils.py
```diff
@@ -310,28 +316,195 @@ def _resolve_altlocs_for_gemmi(atom_array: AtomArray) -> list[str]:
...
+    ------
+    ValueError
+        If any ``(atom_name, altloc)`` pair repeats within a residue.
+    """
+    _check_keys_unique(
+        zip(
+            atom_array.chain_id.tolist(),
+            atom_array.res_id.tolist(),
+            atom_array.atom_name.tolist(),
+            altlocs,
```
marcuscollins, line 413: Probably altlocs should be derived from the AtomArray directly as well?

src/sampleworks/synthetic/synthetic_utils.py
```diff
@@ -310,28 +316,195 @@ def _resolve_altlocs_for_gemmi(atom_array: AtomArray) -> list[str]:
...
+        ),
+        level="atom",
+        identity="(chain_id, res_id, atom_name, altloc)",
```
marcuscollins, line 416: It might be good to derive the iterator passed to keys from these `identity` names, something like 
```python
identities = ("chain_id", "res_id", "atom_name", "altloc")
zipped = zip(atom_array.get(k).tolist() for k in identities
_check_keys_unique(zipped, level="atom", identities = str(identities))
```
that way if anything ever changes you only have to change it one place. 

Not likely to change, so I wouldn't block this PR on this change, but something to use in the future at least.

marcuscollins: You could even make another intermediate method, check_atom_array_keys say, that accepted the AtomArray, level, and identities, that you could re-use in several places here.

src/sampleworks/synthetic/synthetic_utils.py
```diff
@@ -310,28 +316,195 @@ def _resolve_altlocs_for_gemmi(atom_array: AtomArray) -> list[str]:
...
+
+
+def _check_no_repeated_residues(atom_array: AtomArray, residue_span_start_idx: np.ndarray) -> None:
+    """Require each ``(chain_id, res_id)`` key to occupy exactly one contiguous block.
+
+    Parameters
+    ----------
+    atom_array : AtomArray
+        Structure to check.
+    residue_span_start_idx : np.ndarray
```
marcuscollins, line 427: Again I think I'd get this from the AtomArray internally to avoid mistakes.

src/sampleworks/synthetic/synthetic_utils.py
```diff
@@ -310,28 +316,195 @@ def _resolve_altlocs_for_gemmi(atom_array: AtomArray) -> list[str]:
...
-    # boundary shows where a new residue begins (i.e., chain or res_id changed).
-    boundary = (chain_id[1:] != chain_id[:-1]) | (res_id[1:] != res_id[:-1])
-    start_indices = [0, *(np.flatnonzero(boundary) + 1).tolist(), len(atom_array)]
-    yield from zip(start_indices[:-1], start_indices[1:])
+    # residue_boundary_mask[i] is True when atom i + 1 starts a new residue; atom 0 always
+    # marks a new start, which is why the flatnonzero below prepends True.
+    # (n_atoms - 1,) bool
+    residue_boundary_mask = (chain_id[1:] != chain_id[:-1]) | (res_id[1:] != res_id[:-1])
+    # (n_residues,) int
+    residue_span_start_idx = np.flatnonzero(np.concatenate([[True], residue_boundary_mask]))
```
marcuscollins, line 499: Okay, so I guess the point of creating these here, rather than inside the _check_* methods, is that residue_boundary_mask gets used in more than one place. If that's the case, it would make sense to make two separate methods `_get_residue_bndy_mask` and `_get_residue_span_starts`, so that you can individually unit test them.

tests/synthetic/test_generate_synthetic_sf.py
```diff
@@ -223,16 +234,57 @@ def test_saved_structure_round_trips_annotations(
...
+    A gemmi structure has the following hierarchy: first model is structure[0], first chain
+    is model[0], first residue is chain[0], first atom is residue[0].
+    """
+
+    def test_empty_atom_array_raises(self):
+        """An empty AtomArray fails fast rather than yielding a chain-less structure."""
+        with pytest.raises(ValueError, match="empty AtomArray"):
+            atomarray_to_gemmi(AtomArray(0))
+
+    def test_single_atom_array_converts(self):
```
marcuscollins, line 269: This honestly doesn't seem like a particularly useful test.

marcuscollins (approved):
I think there are some improvements you could make here (see comments), so I'd encourage you to make those before merging, but I will approve and leave changes up to you.

src/sampleworks/synthetic/synthetic_utils.py
DorisMai: addressed, now use biotite's own `get_residue_starts`

src/sampleworks/synthetic/synthetic_utils.py
DorisMai: yes. `_resolve_altlocs_for_gemmi` now calls `_check_no_repeated_atoms` after resolving altlocs from the atomarray.

tests/synthetic/test_generate_synthetic_sf.py
DorisMai: removed

src/sampleworks/synthetic/synthetic_utils.py
DorisMai: addressed, now just rely on biotite's get_residue_starts

src/sampleworks/synthetic/synthetic_utils.py
DorisMai: added

Issue #332, opened 2026-07-21
Add protpardelle checkpoints to test and actl images; create test fixtures and add to model registry


marcuscollins:
Protpardelle has its own set of model artifacts, which are currently stored here: `mnt/diffuse-shared/Marcus/model_params`. When running the ppdl model right now, we have to manually set the `PROTPARDELLE_MODEL_PARAMS` environment variable so that ppdl can find them. These need to be baked into our images for CI testing and actl use in order to productionize protpardelle. 

Then we need to create a protpardelle model fixture like those for the other modes (e.g. boltz-2 has a boltz2_wrapper)

You can see the resulting errors by running `RUNTIME_PIXI=1 pixi run -e boltz-dev tests` on the PR 300 branch.

marcuscollins:
Update:

In order to run tests locally and in actl, I added protpardelle to MODEL_WRAPPER_REGISTRY in `tests/conftest.py` and added this to the same file, in PR 300. 

```@pytest.fixture(scope="session")
def protpardelle_checkpoint_path() -> Path:
    if not PROTPARDELLE_AVAILABLE:
        pytest.skip("Protpardelle dependencies not installed in this environment")
    import os
    # where we will eventually install the checkpoint on our containers.
    builtin_path = Path(getsitepackages()[0]) / "release_data"/ "protpardelle" / "model_params"

    # if the user specifies a path to the model_params directory, use that instead.
    path = Path(os.environ.get("PROTPARDELLE_MODEL_PARAMS", builtin_path))
    path = path / "weights/cc89_epoch415.pth"
    return path


@pytest.fixture(scope="session")
def protpardelle_wrapper(protpardelle_checkpoint_path: Path, device: torch.device):
    if not PROTPARDELLE_AVAILABLE:
        pytest.skip("Protpardelle dependencies not installed in this environment")

    config_path = files("sampleworks.data").joinpath("cc89_epoch415.yaml")
    return ProtpardelleWrapper(
        config_path=str(Path(config_path).expanduser().resolve()),
        checkpoint_path=protpardelle_checkpoint_path,
        device=device,
    )
```

All tests now pass except a couple for the experiment presets. However we still need the checkpoint baked into our images. It should be at the above specified location: `Path(getsitepackages()[0]) / "release_data"/ "protpardelle" / "model_params"`

manzuoni-astera:
Sweep update: no active PR currently covers the requested checkpoint baking plus fixture/registry work.

PR #352, opened 2026-07-28
(feat)add protpardelle experiment preset

marcuscollins:
Adding the preset toml file and piping through some options to allow protpardelle to work correctly.

Also includes changes to pixi environment discovery so that we do not need to manually update the available environments and instead derive these from pyproject.toml automatically in several places. 

Tested with a full grid search with reduced hyper-parameter space (using cli options) but all 40 of our test proteins. https://extshell.org/experiments/3dcef558-d8f0-43cc-ab12-567ebca76bbb for those with access.

## Summary by CodeRabbit

* **New Features**
  * Added a new Protpardelle “occ-sweep” experiment job configuration.
  * Added Protpardelle as a selectable model and added the `--guidance-start` CLI option.
* **Bug Fixes**
  * Improved grid-search worker error reporting and subprocess log discovery, including safer result loading.
  * Made runtime environment discovery more resilient when the pixi manifest is missing, adjusting env validation accordingly.
* **Tests**
  * Expanded coverage for model-to-environment mappings, unknown-model rejection, and pixi environment discovery/validation behavior.
* **Chores**
  * Improved determinism in CLI/preset-related tests by using ordered collections.

### DIFF
--- a/experiments/protpardelle.toml
+++ b/experiments/protpardelle.toml
@@ -1,0 +1,29 @@
+description = "Protpardelle canonical occ-sweep job. Make sure to set PROTPARDELLE_MODEL_PARAMS"
+
+[defaults]
+DATA_DIR = "/data/inputs"
+RESULTS_DIR = "/data/results/protpardelle"
+MSA_CACHE_DIR = "/root/.sampleworks"
+PROTEINS_CSV = "${DATA_DIR}/proteins.csv"
+PPDL_CC89_CHECKPOINT = "/mnt/diffuse-shared/marcus/model_params/weights/cc89_epoch415.pth"
+
+[shared_args]
+proteins = "${PROTEINS_CSV}"
+model = "protpardelle"
+scalers = "pure_guidance"
+guidance-start = 400
+num-diffusion-steps = 500
+partial-diffusion-step = 300
+ensemble-sizes = "8"
+gradient-weights = "0.0 0.02 0.035 0.05 0.1 0.2"
+model-checkpoint = "${PPDL_CC89_CHECKPOINT}"
+gradient-normalization = true
+augmentation = false
+align-to-input = true
+
+[[jobs]]
+name = "protpardelle"
+env = "protpardelle"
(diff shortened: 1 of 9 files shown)


manzuoni-astera:
Reviewed. The lock is sorted and CI is green, so this is about the code.

Three things worth changing, all fixed in #357 against this branch so you can merge it in rather than re-doing them:

1. **`_load_valid_pixi_envs()` can crash at import, or bind to the wrong manifest.** The wheel target doesn't include `pyproject.toml`, so on a wheel install the walk-up hits the filesystem root and the `FileNotFoundError` fires at module scope — `sampleworks.runs` becomes unimportable. And when a `pyproject.toml` sits *above* the environment, which is exactly the `/app` layout in the ACTL image, the walk silently binds to that one instead of the checkout's, so the validated env list needn't match what the runner resolves. Fix routes through `SAMPLEWORKS_PIXI_PROJECT_DIR` first (matching `runner._pixi_project_dir()`), falls back to the walk, and returns empty instead of raising. Five tests added.
2. **Stray `log.warning` debug line** in `run_guidance_queue_script` — fires per worker at WARNING and duplicates the `Running worker N: {cmd}` info line right below it.
3. **`BUNDLED` is a set and it parametrizes tests**, so test IDs shuffle with `PYTHONHASHSEED`. It was a list before the move into conftest; made it a tuple.

Verified and no change needed: `--guidance-start = -1` is consistent with `GuidanceConfig` and both consumers guard on `> 0`. Worth noting that wiring it into `build_args_for_process_pool` is a real fix by itself — without it the preset's `guidance-start = 400` was being silently dropped.

One thing I left alone: `PPDL_CC89_CHECKPOINT` points at `/mnt/diffuse-shared/marcus/model_params/…`. `rf3.toml` uses `/checkpoints/…` from the baked image, and #332 tracks doing the same for protpardelle, so this preset won't resolve for anyone else. Didn't touch it since it's presumably what your current runs point at.

PR #360, opened 2026-08-05
ci: move Harbor registry refs to repository variables, add .gitleaksignore

Abdelsalam-Abbas:
Quiets the recurring internal-reference secret-scan alerts for this repo (18 findings re-posting on every PR; now 25 in grown history).

- `docker.yml`: the internal registry hostname now comes from the `ASTERA_REGISTRY` and `SAMPLEWORKS_CHECKPOINTS_SOURCE_IMAGE` repository variables, with no in-repo fallback. New fail-fast guard in the astera job: `docker/login-action` with an empty registry defaults to Docker Hub and would send the Harbor credentials there.
- `.gitleaksignore`: fingerprints for all historical findings (all `astera-internal-hostname`). README and `Dockerfile.astera` keep their references as intentional internal docs while the repo is private; going public requires republishing as a fresh single-commit repo per the runbook regardless.

**Merge requirement:** the two repository variables must be set first or the next docker workflow run fails fast:

- `ASTERA_REGISTRY` = the Harbor hostname
- `SAMPLEWORKS_CHECKPOINTS_SOURCE_IMAGE` = the digest-pinned checkpoints ref

Verified: full-history scan with the internal-refs rules is clean with this ignore file; workflow YAML parses; pre-commit hooks pass.

## Summary by CodeRabbit

* **Documentation**
  * Updated setup instructions to use the `sampleworks` ACTL image alias.
  * Documented required digest-pinned checkpoint images and registry configuration.
  * Clarified that builds stop early when the Astera registry is unavailable or invalid.

* **Bug Fixes**
  * Improved image validation and registry handling during checkpoint synchronization and Astera builds.
  * Removed hardcoded registry references from local image build instructions.
  * Added checks to reject public registries and require valid private registry destinations.


manzuoni-astera:
The workflow changes are solid. Two things about the PR itself rather than the code.

**`.gitleaksignore` isn't in the diff.** It's in the title and the body ("fingerprints for all historical findings"), but the PR touches exactly three files — `docker.yml`, `Dockerfile.astera`, `README.md` — and the file doesn't exist on `main` either. So the stated headline purpose, quieting the 25 historical findings, isn't delivered by this branch. Either it got lost before pushing, or the title/body needs to drop it.

**The merge requirement names the wrong things.** The body says to set `ASTERA_REGISTRY` and `SAMPLEWORKS_CHECKPOINTS_SOURCE_IMAGE` as repository *variables*. The diff reads `secrets.ASTERA_REGISTRY` and `vars.SAMPLEWORKS_CHECKPOINTS_SOURCE_PATH`. Following the body verbatim gets you a secret set as a variable and a variable under the old name, and the new guards then fail the next run — which is exactly the situation the "set these first" note is meant to prevent. Worth fixing since that's the bit someone will action at merge time.

Body also says README and `Dockerfile.astera` "keep their references as intentional internal docs", but the diff removes them from both. That's the better outcome — I checked and `harbor.astera.sh` no longer appears anywhere in the tree — the body just describes an earlier version.

**What's good:** checking the parts instead of the composed ref is a genuinely subtle catch, since an unset host still yields a non-empty `/library/...@sha256:...` that sails through both the emptiness and digest-pinned checks. The public-registry guard is the right call too, given `login-action` silently defaults to Docker Hub on an empty registry and the Harbor creds would follow. Both jobs that consume the secret scope it themselves, the rename is clean with no stale `_SOURCE_IMAGE` references, the YAML parses, and the push/tag/dispatch triggers sidestep the fork-PR secrets problem.

xraymemory (approved):

Issue #139, opened 2026-03-05
Use RF3's unresolved-atom placement strategy for model_atom_array initialization


k-chrispens:
The RF3 wrapper currently initializes unresolved atoms (NaN coords, occupancy=0 from `add_missing_atoms`) with `centroid + N(0,1)` noise. RF3 itself has a better strategy in its pipeline: place unresolved atoms on the nearest resolved neighbor. We should reuse that.

Two relevant atomworks functions (used in the RF3/AF3 pipeline):

1. **`PlaceUnresolvedTokenAtomsOnRepresentativeAtom`** (`atomworks.ml.transforms.featurize_unresolved_residues`) — for partially-resolved tokens, moves unresolved atoms to the representative atom position within the same token.
2. **`place_unresolved_token_on_closest_resolved_token_in_sequence()`** (`atomworks.ml.transforms.featurize_unresolved_residues`) — for fully-unresolved tokens, places all atoms on the nearest resolved token in sequence space (within the same chain). Falls back to `nan_to_num` (zeros) if no tokens in the chain are resolved.

Both operate on a `coord_to_be_noised` annotation rather than modifying `.coord` directly, which is a clean pattern we could follow.

The fix location is `src/sampleworks/models/rf3/wrapper.py`, in the block after `model_aa` is truncated to `num_atoms` (around line 330). Replace the current `centroid + noise` initialization with calls to these two functions, applied to `model_aa.coord`.

PR #362, opened 2026-08-06
feat(astera): put SBGrid on $PATH when /programs is mounted

manzuoni-astera:
Astera now installs the SBGrid collection once on a shared `sbgrid-programs` PVC, which the actl catalog auto-mounts **read-only at /programs** for every diffuse profile. The mount arrives regardless of image — but on this image a scientist still had to run three manual steps before phenix worked:

```sh
apt-get update && apt-get install -y tcsh bsdextrautils
ln -sfn /mnt/sbgrid-programs /programs
source /programs/sbgrid.shrc
```

The symlink goes away on the actl side. This PR removes the other two, so `--image sampleworks` lands you in a shell where phenix just works.

### Why tcsh is not optional

`/programs/sbgrid.shrc` aborts with *"requires the C shell"* when `/bin/csh` is missing and sets up **nothing at all** — the entire collection stays invisible. `bsdextrautils` provides `column(1)`, which SBGrid's `sbgrid-obsolete` helper shells out to on every title lookup. Both added to the existing verification loop.

### The shell init

Follows the existing ext-shell pattern (`/etc/profile.d/` + `install-*-shell-hooks.sh`). Guarded so it cannot break shells with no collection: `/programs` is created empty in the image, so a non-diffuse workspace or a `--no-mount` run is a clean no-op.

The `|| true` is **load-bearing** — sbgrid.shrc runs commands that return non-zero benignly, so sourcing it under `set -e` kills the shell. Verified on a live pod; it's why my first attempt at this failed.

### Verified

Against the real collection, on the current image, with the PVC mounted read-only exactly as the catalog will:

- interactive shell resolves `phenix.molprobity` with **no setup**, and scores 1CRN correctly (MolProbity 0.56, 0.00% Ramachandran outliers)
- `bash -lc 'phenix.version'` works for non-interactive callers
- `pixi` unaffected (0.71.2) — the analysis env and SBGrid coexist
- absent mount is a clean no-op under both plain and `set -e` shells
- `SBGRID_NO_AUTOINIT=1` opts out

### Note

This brings phenix and the ~3,200 other SBGrid commands. **tortoize is not an SBGrid title** and still comes from `pixi run -e analysis`.

Depends on Astera-org/asteractl#138 + #139 for the /programs mount; harmless to merge before them (no mount → no-op).

## Summary by CodeRabbit

* **New Features**
  * Expanded the Astera image with additional shell utilities and command-line tools.
  * Added automatic SBGrid shell initialization for Bash and Zsh environments.
  * Added support for creating and using the `/programs` directory when available.

* **Bug Fixes**
  * Prevented duplicate shell startup entries.
  * Reduced initialization errors from disrupting shell startup.
  * Added safeguards to skip unavailable configuration files.

### DIFF
--- a/Dockerfile.astera
+++ b/Dockerfile.astera
@@ -70,9 +70,9 @@ RUN install -m 0644 \
     && EXT_VERSION="${EXT_VERSION}" bash /usr/local/share/sampleworks/astera/install-ext.sh \
     && bash /usr/local/share/sampleworks/astera/install-ext-shell-hooks.sh
 
-# SBGrid lives on the shared sbgrid-programs PVC, auto-mounted read-only at
-# /programs by the actl catalog. /programs is created empty so the guarded init
-# is a clean no-op when the PVC is not mounted.
+# SBGrid lives on shared storage, mounted read-only at /programs for every
+# diffuse workspace by the actl catalog. /programs is created empty so the
+# guarded init is a clean no-op when the mount is absent.
 RUN mkdir -p /programs \
     && install -m 0644 \
         /usr/local/share/sampleworks/astera/sbgrid-shell.sh \
--- a/Dockerfile.astera
+++ b/Dockerfile.astera
@@ -34,6 +34,7 @@ ENV DEBIAN_FRONTEND=noninteractive \
 
 RUN apt-get update && apt-get install -y --no-install-recommends \
         bash \
+        bsdextrautils \
         ca-certificates \
         curl \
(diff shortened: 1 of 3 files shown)


marcuscollins (approved):
LGTM. I'm not sure what's going on with the tests and scans. I triggered them to run again, maybe it was transient. Please don't merge if they are still failing, but other wise go ahead.

PR #366, opened 2026-08-10
feat(it-opt): give each ensemble member its own trunk latent

smallfishabc:
## Why

IT-opt optimizes a single `s`/`z` latent shared by every ensemble member, so
members can only diverge through their prior noise. Under the density reward the
ensemble collapses toward one solution instead of spreading.

## What changed

Each member now gets its own independent latent. `_leaf_latents` stacks
`ensemble_size` copies of the trunk baseline onto a leading batch dim, one
optimizable leaf per member. Since stock diffusion modules take an un-batched
conditioning, `_PerMemberStepper` slices that dim back apart and runs the
model's normal `step` once per member — same FLOPs as one batched call.

Stacked on `fy/it-optimization` (#313), which is merged in, so the diff here is
only the incremental work.

## Summary by CodeRabbit

- **New Features**
  - Added IT-opt experiment presets for 1-, 4-, and 8-GPU runs, including configurable outer-loop variants.
  - Added support for running multiple jobs per GPU to improve experiment throughput.
  - Added command-line configuration for latent-optimization options.
  - IT-opt now supports independent ensemble-member optimization and partial-diffusion starts.
  - Added denoised predictions for each sampling step to run metadata.

- **Tests**
  - Expanded coverage for latent optimization, geometry rewards, and bundled experiment configurations.


src/sampleworks/core/scalers/latent_optimization.py
```diff
@@ -126,6 +126,70 @@ def __call__(self, latents: Sequence[Tensor], baselines: Sequence[Tensor]) -> Te
...
+    """Denoise one ensemble member at a time, each with its OWN trunk latent.
+
+    The it_opt "multiple-leaf" scheme gives every ensemble member its own latent, so ``s``/``z``
+    carry a leading ``ensemble_size`` batch dim. Stock model diffusion modules instead take an
+    un-batched conditioning and broadcast it internally, so a batched latent either crashes
+    (Protenix) or mis-broadcasts (Boltz's ``multiplicity``). This adapter sidesteps that in a
+    model-agnostic way: for each member it slices that member's latent, runs the wrapped model's
+    normal un-batched ``step`` on that member alone, and stacks the results. Members stay
+    independent, so gradients stay per-member; the summed density reward couples them only through
+    the ensemble average. Cost is N forwards instead of one batched call, with an equivalent result.
```
marcuscollins, line 139: We should keep this in mind in the future, especially if the performance hit is large. I imagine this could slow down inference a lot.

src/sampleworks/core/scalers/latent_optimization.py
```diff
@@ -126,6 +126,70 @@ def __call__(self, latents: Sequence[Tensor], baselines: Sequence[Tensor]) -> Te
...
+
+    def __init__(
+        self,
+        model,
+        io: AttrLatentIO,
+        *,
+        optimize_single: bool,
+        optimize_pair: bool,
+        ensemble_size: int,
+    ):
```
marcuscollins, line 151: Needs a doc string to explain what all the arguments are.

src/sampleworks/core/scalers/latent_optimization.py
```diff
@@ -126,6 +126,70 @@ def __call__(self, latents: Sequence[Tensor], baselines: Sequence[Tensor]) -> Te
...
+        self._model = model
+        self._io = io
```
marcuscollins: Calling this "io" will probably confuse casual readers--it makes it seem like it's file I/O, when it is really an object for holding the latents, right? Maybe just call it `self._latents` or `self._optimizable_latents`

src/sampleworks/core/scalers/latent_optimization.py
```diff
@@ -126,6 +126,70 @@ def __call__(self, latents: Sequence[Tensor], baselines: Sequence[Tensor]) -> Te
...
+        self._optimize_single = optimize_single
+        self._optimize_pair = optimize_pair
+        self._ensemble_size = ensemble_size
```
marcuscollins: You might see if you can set this lazily, on the first call to `step`, so that you can initialize the class without needing to know the batch size in advance. In fact, ideally you never need to know the batch size beyond initializing data loaders, and your `step` function just reacts to whatever data it is passed.

src/sampleworks/core/scalers/latent_optimization.py
```diff
@@ -126,6 +126,70 @@ def __call__(self, latents: Sequence[Tensor], baselines: Sequence[Tensor]) -> Te
...
+
+    def step(self, x_t: Tensor, t, *, features: GenerativeModelInput) -> Tensor:
+        """Loop the wrapped model's ``step`` over ensemble members; stack the per-member results."""
+        cond = features.conditioning
+        per_member: list[Tensor] = []
+        for i in range(self._ensemble_size):
```
marcuscollins: As in the previous comment, you can probably extract the batch/ensemble size just by inspecting `cond`.

src/sampleworks/core/scalers/latent_optimization.py
```diff
@@ -126,6 +126,62 @@ def __call__(self, latents: Sequence[Tensor], baselines: Sequence[Tensor]) -> Te
...
+            if self._optimize_single:
+                cond_i = self._io.write_single(cond_i, self._io.read_single(cond)[i])
+            if self._optimize_pair:
+                cond_i = self._io.write_pair(cond_i, self._io.read_pair(cond)[i])
+            t_i = t
+            if isinstance(t, Tensor) and t.ndim >= 1 and t.shape[0] == x_t.shape[0]:
+                t_i = t[i : i + 1]
+            features_i = GenerativeModelInput(conditioning=cond_i)
+            per_member.append(self._model.step(x_t[i : i + 1], t_i, features=features_i))
+        return torch.cat(per_member, dim=0)
```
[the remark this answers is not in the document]
marcuscollins: A better solution is just to get the ensemble size from your conditioning variable `cond` and then you won't have to worry about getting it wrong.

src/sampleworks/core/scalers/latent_optimization.py
```diff
@@ -259,6 +327,16 @@ def sample(
         schedule = sampler.compute_schedule(num_steps=self.num_steps)
         grad_enabler = _GradEnablingScaler()
 
+        # Denoise per member so each uses its own latent: stock model diffusion modules take an
+        # un-batched conditioning, so the batched per-member latents can't go through in one call.
+        stepper = _PerMemberStepper(
```
marcuscollins, line 345: Do you want to make this optional?

src/sampleworks/core/scalers/latent_optimization.py
```diff
@@ -367,7 +450,14 @@ def _leaf_latents(self, features: GenerativeModelInput, io: AttrLatentIO):
...
                 )
             baseline = baseline.detach()
-            leaf = baseline.clone().requires_grad_(True)
+            # it_opt "multiple-leaf" scheme: give each ensemble member its OWN latent. We stack
+            # ensemble_size independent copies of the trunk baseline into a leading batch dim, so
+            # each member gets its own gradient and can diverge, instead of collapsing onto one
+            # shared latent. (The reference does the same via batch-expand-then-clone.) Stock
+            # diffusion modules take an un-batched conditioning, so _PerMemberStepper slices this
+            # batch dim back apart and runs one forward per member at denoise time.
+            member_copies = [baseline for _ in range(self.ensemble_size)]
```
marcuscollins, line 471: do these need to be cloned? I think they might still be linked to the same object if they aren't. But I'm not exactly sure of the behavior of `torch.stack`, maybe it makes fresh copies anyway?

src/sampleworks/utils/cif_utils.py
```diff
@@ -247,6 +248,79 @@ def resolve_mixed_hetatm_atom_altlocs(cif_path: Path | str) -> Path:
     return tmp_path
 
 
+def remap_altlocs_to_ab(cif_path: Path | str) -> Path:
+    """Relabel a residue's non-A/B alternate altloc(s) to fill the A/B slots.
```
marcuscollins: Meta-comment: each PR should ideally do only one thing. This _probably_ could be PR'd separately.

Question: when does/did this happen, and do you think it might have affected any earlier sampleworks results? I thought all of our code was agnostic to the actual altloc IDs... And if there _is_ code that assumes the labels are A/B, it would be better to fix that where the assumption is... because:

Possible serious problem: in fact there are cases where there are more than two altlocs (A,B,C). So it's possible that this will do something we actually don't want, e.g., say for residue 10 there are altlocs A, B, and C, and for residue 23 there are altlocs A and C only, but altloc C of residue 23 clashes with altloc B of residue B. Renaming residue 23 altloc C to B would then force a clash into the structure that we don't want.

src/sampleworks/utils/guidance_script_arguments.py
```diff
@@ -404,6 +411,24 @@ def populate_config_for_guidance_type(self, job: JobConfig, args: argparse.Names
             self.fk_lambda = args.fk_lambda
             self.fk_resampling_interval = args.fk_resampling_interval
             self.ensemble_size = job.ensemble_size
+        elif job.scaler == GuidanceType.LATENT_OPT:
+            # IT-opt tunables are scalar flags rather than grid axes, so they come straight off
+            # the driver's namespace. This copy is what makes them work at all: the grid builds
```
marcuscollins: I have no idea what this means. Can you explain better what's going on here?

marcuscollins: Oh, I see what's going on. This comment (I guess it came from Claude?) is more confusing than helpful. I would just remove it. What you're doing here fits with the pattern of the rest of the `populate_config...` method, so to call it out specifically here is just confusing. 

In general, Claude over-explains itself and does so at a very local level in the code, which I think can lead to more confusion, since it makes it seem that there's something different about the some particular part of the code, when really it is just following a pattern. I suggest reading these comments and removing them if they aren't calling out special, unusual, or different behavior.

src/sampleworks/utils/guidance_script_arguments.py
```diff
@@ -404,6 +411,24 @@ def populate_config_for_guidance_type(self, job: JobConfig, args: argparse.Names
...
+            # it every --which-latent / --learning-rate / --outer-steps flag is a silent no-op,
+            # left at the add_latent_opt_args default that __post_init__ already seeded.
+            for attr in _LATENT_OPT_ATTRS:
+                value = getattr(args, attr, None)
+                if value is not None:
+                    setattr(self, attr, value)
+            # _run_guidance reads ensemble_size directly rather than via getattr, so it must
+            # be set here or a latent_opt job raises AttributeError.
+            self.ensemble_size = job.ensemble_size
+            # Deliberately no step_size / step_scaler_type here. add_latent_opt_args does not
```
marcuscollins: As above, remove this comment.

src/sampleworks/utils/guidance_script_arguments.py
```diff
@@ -404,6 +411,24 @@ def populate_config_for_guidance_type(self, job: JobConfig, args: argparse.Names
...
+            # the driver's namespace. This copy is what makes them work at all: the grid builds
+            # its GuidanceConfig directly and never calls from_cli(), so the _DYNAMIC_ATTRS loop
+            # that normally moves parsed values onto a config is unreachable from here. Without
+            # it every --which-latent / --learning-rate / --outer-steps flag is a silent no-op,
+            # left at the add_latent_opt_args default that __post_init__ already seeded.
+            for attr in _LATENT_OPT_ATTRS:
+                value = getattr(args, attr, None)
+                if value is not None:
+                    setattr(self, attr, value)
+            # _run_guidance reads ensemble_size directly rather than via getattr, so it must
```
marcuscollins: I suspect this is probably wrong. `setattr(self, "ensemble_size", job.ensemble_size)` should work. OTOH, it doesn't really matter, and this approach is fine too.

Remove the comment, it's just noise.

src/sampleworks/utils/guidance_script_utils.py
```diff
@@ -507,6 +507,10 @@ def _run_guidance(args: GuidanceConfig, guidance_type: str, model_wrapper, devic
 
         structure = annotate_structure_for_protenix(
             structure,
+            # Root Protenix's per-sample input dump (protenix_input*.json) under the job's output
+            # dir; otherwise out_dir falls back to the input id, resolves against the CWD, and
+            # leaves a stray <input_id>/ folder there on every run.
+            out_dir=str(Path(args.output_dir) / "protenix_input"),
```
marcuscollins: have you tested that this change doesn't break the existing diffusion guidance? 

This is the kind of thing that _should_ probably be a separate PR into main, so that you can test it independently by making sure a diffusion guidance grid search with protenix works as expected. Then you would rebase your PR branch onto main to pick up this change.

marcuscollins (changes requested):

src/sampleworks/core/scalers/latent_optimization.py
smallfishabc: Yes. And I think we should consider this a bit more and maybe try to optimize this in the future.

src/sampleworks/core/scalers/latent_optimization.py
smallfishabc: I think may be I can all it _latent_io or something like that since it is a wrapper for the conditioning.

src/sampleworks/core/scalers/latent_optimization.py
smallfishabc: torch.stack allocates a new output tensor and copies each input into its slice. So I think this should be fine.

src/sampleworks/core/scalers/latent_optimization.py
smallfishabc: Yes. Do you have any recommendations on which kind of optional? Like a flag or other options?

src/sampleworks/utils/guidance_script_utils.py
smallfishabc: I think this one will not impact the default behavior. I can comment it out so that it should not impact this PR. Or I can specify a default for other model

src/sampleworks/utils/cif_utils.py
smallfishabc: I moved this out of current PR

src/sampleworks/core/scalers/latent_optimization.py
smallfishabc: I will do that in a new PR

docs/IT_OPT_DESIGN.md
```diff
@@ -74,8 +74,8 @@ LatentOptimization(
 )
 ```
 
-- `ensemble_size` structures are sampled in parallel and **share** the latents (per-member latents
-  are a documented follow-up).
+- `ensemble_size` structures are sampled in parallel, each with its **own** `s`/`z` leaf, so members
+  can diverge under the reward instead of collapsing onto one solution.
```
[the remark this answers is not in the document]
marcuscollins: @smallfishabc do you understand this comment? Is there a setting where you optimize the single and pair latents together? Does it also use per-member latents? If this is just saying that we don't use per-member latents if we aren't optimizing latents at all, then we can ignore it. Otherwise we should clarify.

src/sampleworks/core/scalers/latent_optimization.py
```diff
@@ -139,6 +216,7 @@ def __init__(
         num_steps: int = 200,
         guidance_t_start: float = 0.0,
         *,
+        t_start: float = 0.0,
```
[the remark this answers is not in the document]
marcuscollins: @coderabbit make an issue to address this validation.

src/sampleworks/utils/guidance_script_utils.py
```diff
@@ -673,6 +675,8 @@ def _run_guidance(args: GuidanceConfig, guidance_type: str, model_wrapper, devic
         # carries it as an integer step count, so we convert it here and default to optimizing from
         # the first step.
         guidance_t_start = args.guidance_start / num_steps if args.guidance_start > 0 else 0.0
+        # Same conversion for partial diffusion, matching the pure-guidance branch above.
+        t_start = args.partial_diffusion_step / num_steps if args.partial_diffusion_step else 0.0
```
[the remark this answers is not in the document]
marcuscollins: @coderabbit make an issue to address this validation.

marcuscollins (approved):
I left some comments here and communicated over Slack as well. You might want to change a couple things in a .md file.

PR #255, opened 2026-06-12
Changes to pixi files to facilitate development on Mac OSX

marcuscollins:
These changes make it possible to install most environments on Mac OS X. To do this it changes the `[dependency-groups]` to be actual `[tool.pixi.feature.*]`, and created a `boltz-osx` feature. For protenix in particular this prevents pixi from trying to install environments on platforms where they are not compatible. 

## Summary by CodeRabbit

* **Chores**
  * Updated project configuration to support macOS ARM64 architecture
  * Refined environment setup with explicit feature definitions for specialized workflows
  * Updated core dependencies with specific version constraints for improved stability and compatibility

PR #371, opened 2026-08-13
chore(docs): Update docs to reflect support for macOS

ekiefl:
I was expecting installation for macOS development to fail based on this note in the README:

> **Note**: `pixi install -a` resolves all environments. This (currently) requires CUDA 12 and will fail on machines without it.

Instead `pixi install -a` succeeded, skipping the CUDA-only environments:

```
WARN Skipping environment(s) that do not support the current platform: boltz, boltz-analysis, boltz-dev, protenix, protenix-dev
✔ The following environments have been installed: default, analysis, analysis-dev, boltz-osx, protpardelle, protpardelle-dev, rf3, rf3-dev.
```

I see that macOS (arm64) landed in #255, so this PR updates the README to reflect that. I'm reading between the lines, but from what I gather macOS isn't supported as a first-class platform (no CUDA), just as a convenience for local development. I've tailored the language according to that assumption.

Relatedly, I noticed `AGENTS.md` is stale because it lists all the pixi environments but doesn't mention `boltz-osx`. I considered adding it but I think the proper solution is to avoid this "agentic doc debt" altogether (`protpardelle` is also missing) by pointing agents to the cmd `pixi workspace environment list`. Can file an issue / submit PR for this if requested.

## Summary by CodeRabbit

- **Documentation**
  - Updated installation instructions to include macOS arm64 support for local development.
  - Clarified that CUDA environments are available only on Linux.
  - Added guidance for installing specific environments and viewing the complete environment list.


k-chrispens (approved):
Thanks for this! The AGENTS.md update would also be much appreciated - it is probably due for a more comprehensive update, there are a few parts of the codebase that aren't referenced in it currently that ought to be. Pointing agents towards programmatic interactions with the code base is a perfect way to make sure it stays updated for the pixi envs!

marcuscollins (approved):
Thanks for updating this documentation!

PR #373, opened 2026-08-13
feat(rewards): add a prepare hook to the reward protocol and scalers

manzuoni-astera:
> Merged with `prepare(atom_array, *, device)`. The 4 Sep decision to pass `RewardInputs` instead is implemented in #402.

## Summary

The structure-factor reward merged in #324 cannot run. It is built in two phases, and nothing
in `src/` has ever called the second one, so the reward has tests but no way to reach a guidance
job. This adds a prepare hook to the reward protocol and calls it from both trajectory scalers
once the model atom array exists. It makes #324's reward reachable and gives #319 and #372 one
hook to share instead of three private variants. Independent of the registry work in #374 and
#375.

## Changes

`PreparableRewardFunctionProtocol` joins `core/rewards/protocol.py`, alongside a
`prepare_reward_if_needed(reward, atom_array, *, device)` helper that no-ops on rewards which
don't implement it. Callers apply it unconditionally, so the density path is untouched.

The signature is the one #324 already shipped: `prepare(atom_array, *, device) -> None`, mutating
the reward rather than returning a new object. Worth settling now, because the two rewards queued
behind this disagree with it. #319's prepare returns a value, and #372's takes a structure rather
than an atom array.

Both call sites sit in `sample()`, right after the processed structure is built and before the
denoising loop: `core/scalers/pure_guidance.py` and `core/scalers/fk_steering.py`. They pass
`processed_structure.reward_atom_array`, a new property that returns the model atom array when
the model exposes one and the input structure's otherwise. That is deliberately the same choice
`to_reward_inputs` makes. If the two ever diverge, a reward's atom ordering stops matching the
coordinate tensor it is scoring, and the failure is silent: wrong numbers, no error.

The property also replaces `model_atom_array or atom_array`. Whether an empty `AtomArray` is
falsy is biotite's decision, not something we should build on.

Also in here: the protocol docstring now states that rewards are minimized, since guidance
backprops the value and FK steering selects with `argmin`. Nothing said so before, and weighted
combinations in #375 only mean something if every term agrees on the sign.

## Testing

`tests/rewards/test_prepare_hook.py` covers the helper directly: it prepares a two-phase reward
with the atom array and device it was given, leaves a one-phase reward alone, and can be re-run
to rebind a reward to a different topology.

`tests/integration/test_pipeline_integration.py` drives both trajectory scalers end to end with a
recording reward, asserting that `prepare` ran exactly once, before the first evaluation, with
the model's atom array. The model there has four atoms and the structure five, so preparing
against the input structure fails the test rather than passing by coincidence. A real
`DataSpaceDPSScaler` is used so the reward is genuinely evaluated. Both of those came out of
CodeRabbit review on this PR, and the first version of the test passed vacuously without them.

CI is green: lint, four typecheck environments, four test environments. The GPU workflow has not
run.

## Rollout

Nothing. No new dependencies, no CLI or configuration changes, and no behavior change for any
reward that doesn't implement `prepare`. Merge this before #374 and #375, which build on it.

## Summary by CodeRabbit

* **New Features**
  * Added support for reward functions that prepare themselves against the model’s atom topology before evaluation.
  * Reward inputs now consistently follow the model atom order, including structures with added, removed, or reordered atoms.
  * Added validation for finite coordinates and positive occupancies.
  * Clarified that reward values and weighted combinations follow a minimization convention.

* **Tests**
  * Expanded coverage for reward preparation, topology mismatches, and minimization behavior.


tests/integration/test_pipeline_integration.py
```diff
@@ -585,6 +585,57 @@ def test_trajectory_scaler_handles_multiple_particles(
...
+            structure=mock_structure,
+            model=mock_wrapper,
+            sampler=sampler,
+            step_scaler=mock_step_scaler,
+            reward=reward,
+            num_particles=1,
+        )
+
+        assert reward.prepared_atom_counts == [mock_wrapper.num_atoms]
+        assert reward.calls_before_prepare == 0
```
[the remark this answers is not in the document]
manzuoni-astera: Fixed in the latest push. You were right that the assertion was vacuous: the reward was never called at all, because `MockStepScaler` doesn't evaluate it. The test now uses a real `DataSpaceDPSScaler`, counts calls, and asserts the count is nonzero.

tests/integration/test_pipeline_integration.py
```diff
@@ -585,6 +585,57 @@ def test_trajectory_scaler_handles_multiple_particles(
...
+        trajectory_scaler.sample(
+            structure=mock_structure,
+            model=mock_wrapper,
+            sampler=sampler,
+            step_scaler=mock_step_scaler,
+            reward=reward,
+            num_particles=1,
+        )
+
+        assert reward.prepared_atom_counts == [mock_wrapper.num_atoms]
```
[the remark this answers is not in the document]
manzuoni-astera: Also right, and worse than you said: `MockFlowModelWrapper` exposes no model atom array at all, so `reward_atom_array` was falling back to the structure's array and the test name was overclaiming. It now drives a `MismatchCaseWrapper` where the model has four atoms and the structure five, so preparing against the wrong array fails.

tests/rewards/test_prepare_hook.py
```diff
@@ -0,0 +1,85 @@
...
+    assert reward(torch.ones(1, 4, 3)) == 12.0
+
+
+def test_prepare_is_rerunnable_for_a_new_topology():
+    reward = PreparableReward()
+
+    prepare_reward_if_needed(reward, make_atom_array(3), device="cpu")
+    prepare_reward_if_needed(reward, make_atom_array(5), device="cpu")
+
+    assert reward.prepared_with == [(3, "cpu"), (5, "cpu")]
```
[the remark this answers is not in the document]
manzuoni-astera: Added one-line docstrings to the new test functions and to `RecordingPreparableReward`'s methods. I've left out Parameters/Returns sections on the test bodies, since the guideline is aimed at the library surface and full sections on assertions would be noise.

DorisMai:
@manzuoni-astera 
Oh gees I just saw this draft PR and I realized I had repeated quite a bit of the work on my branch.  I think this PR is a better implementation of the prepare hook, so I would stop my branch and make some suggestion here. 

The main issue is that `reward_atom_array` isn't the reconciled reference structure. For Boltz the template also carries a flat b_factor=20.0 placeholder and for RF3 unresolved atoms carry unseeded Gaussian noise. The SF reward's prepare needs the model atom array topology (total number of atoms for example, rather than input file that could having missing atoms), and needs the reconciled coordinates to estimate the bulk solvent percentage. 

I suggest bringing in something like [`RewardInputs.to_atom_array`](https://github.com/diff-use/sampleworks/blob/a8a98499fd189708369bbb325fa6eaf931b07cc8/src/sampleworks/core/rewards/protocol.py#L138-L190)) from my branch (dm/reward-prepare-hook), and check out the tests ([integration](https://github.com/diff-use/sampleworks/blob/a8a98499fd189708369bbb325fa6eaf931b07cc8/tests/integration/test_mismatch_integration.py#L704-L745) and [rewardinputs](https://github.com/diff-use/sampleworks/blob/a8a98499fd189708369bbb325fa6eaf931b07cc8/tests/rewards/test_reward_inputs.py#L74)) too.

src/sampleworks/core/rewards/protocol.py
```diff
@@ -180,6 +188,59 @@ def __call__(
         ...
 
 
+@runtime_checkable
+class PreparableRewardFunctionProtocol(RewardFunctionProtocol, Protocol):
+    """Protocol for reward functions that must see the model topology first.
```
marcuscollins: Nit: by model do we mean "processed atomic structure topology" here?

tests/rewards/test_prepare_hook.py
```diff
@@ -0,0 +1,91 @@
...
+        """Record the topology and device this reward was bound to."""
+        self.prepared_with.append((atom_array.array_length(), str(device)))
+
+    def __call__(self, coordinates: torch.Tensor, *args, **kwargs) -> torch.Tensor:
+        if not self.prepared_with:
+            self.calls_before_prepare += 1
+        return super().__call__(coordinates)
+
+
+def make_atom_array(n_atoms: int = 4) -> AtomArray:
```
marcuscollins: Is there a reason not to use `build_test_atom_array`? Why make another new function?

src/sampleworks/core/scalers/pure_guidance.py
```diff
@@ -89,6 +89,9 @@ def sample(
 
         reconciler = processed_structure.reconciler.to(coords.device)
         reward_inputs = processed_structure.to_reward_inputs(device=coords.device)
+        prepare_reward_if_needed(
```
marcuscollins, line 99: Can you double check for me: does .to_reward_inputs call the reconciler to prepare the reward_inputs? And does it update `processed_structure` in place with the updated atom array? (I think it does) Otherwise the coordinates passed here could be incorrect.

marcuscollins: Assuming that it does do those things, please add a comment so that people don't get confused or worried when they read this.

tests/integration/test_pipeline_integration.py
```diff
@@ -585,6 +591,83 @@ def test_trajectory_scaler_handles_multiple_particles(
...
+        device: torch.device,
+    ):
+        """Two-phase rewards see the model atom array before any reward evaluation.
+
+        The model and the structure deliberately have different atom counts. A reward
+        prepared against the input structure would bind the wrong atom ordering, which
+        is how the structure-factor reward silently scores the wrong thing.
+        """
+
+        class RecordingPreparableReward(MockGradientRewardFunction):
```
marcuscollins: Why is this class inline?

marcuscollins: Also, could you merge this with the PreparableReward class in test_prepare_hook.py and put it where both tests can access it?

marcuscollins (approved):
Please have a look at the comments and make changes accordingly, but it all looks basically good to me. Mostly just trying to reduce redundancy wherever possible.

src/sampleworks/core/rewards/protocol.py
xraymemory: Other way round, actually — "model" here is the generative model's atom array (`reward_atom_array`), not the processed input structure. The two can differ, which is the whole reason the hook exists. Reworded the docstring in 7115792 so it says that outright.

tests/rewards/test_prepare_hook.py
xraymemory: No good reason, I just didn't know it was there. Swapped to `build_test_atom_array(n_atoms=…)` and deleted the one-off helper (7115792).

src/sampleworks/core/scalers/pure_guidance.py
xraymemory: Checked. Neither, as it turns out: `to_reward_inputs` doesn't touch `processed_structure` (frozen dataclass) and doesn't reconcile coordinates. It builds the reward tensors in model atom order and only uses the reconciler's index maps to copy the structure's B-factors onto the common atoms. `reward_atom_array` is that same model-order array, so the topology `prepare` binds to is the ordering the sampled coords follow — that part is right. 


Added the comment at both to clarify(7115792).

tests/integration/test_pipeline_integration.py
xraymemory: In there now: `MockPreparableRewardFunction` in `tests/mocks/rewards.py`, subclassing `MockGradientRewardFunction`, records `prepared_with` / `calls` / `calls_before_prepare`. Both the hook unit tests and the integration test use it; the inline class and the private `PreparableReward`/`PlainReward` in `test_prepare_hook.py` are gone (7115792).

marcuscollins (approved):

xraymemory:
Follow-up on this thread now that #373 has merged as-is: the change we settled on today with Marcus is in #402. `prepare` takes the `RewardInputs` rather than an atom array, so a reward is prepared against exactly what its `__call__` is fed, and nothing is written back into the model atom array.

`RewardInputs` carries the topology it was built from (`atom_array`, set by `from_atom_array`) and gains `to_atom_array()`, your converter with the template supplied by the inputs themselves. `StructureFactorRewardFunction.prepare` builds its gemmi structure from it, so SFcalculator's solvent estimate sees the reconciled coordinates and B-factors, not the template's. Your two tests are carried over (`test_reward_inputs.py::TestRewardInputsToAtomArray`, `test_mismatch_integration.py::test_to_atom_array_has_model_topology_and_reconciled_values`), the second one also asserting the model atom array is left as it was.

#374 and #375 are updated to the new signature (`CompositeReward.prepare`, the build-integration test).

Issue #358, opened 2026-07-30
Build a reward configuration system


marcuscollins:
We are starting to have lots of different kinds of rewards (see PRs 324, 313, 319). I would like us to develop a configuration system so that we can choose rewards and in the future combine them. Minimally, a configuration should look like:

```
real_space_density:
    weight: 0.4
    reward_options:
        some_option: "good"
some_other_reward:
    weight: 0.6
    reward_options:
        option_1: 10
        option_2: "yay"
```

(YAML not required, could be JSON, something else)

manzuoni-astera:
Had a go at this over the last day: #373, #374, #375. All three are up as drafts, since a few
things below are worth agreeing on before anyone spends review time on them.

The shape: each reward declares its options once as a frozen dataclass and registers a builder.
The CLI flags, the config-file schema, the validation messages and the run metadata all come off
that one declaration, so adding a reward type is a schema plus one registry entry. No argparse
plumbing to edit, no `if/elif` on reward type anywhere.

Two ways to configure a run, one structure underneath:

```bash
sampleworks-guidance ... --reward-type structure_factor --mtzfile 1vme.mtz --bulk-solvent combined
sampleworks-guidance ... --reward-config rewards.yaml   # the mapping from this issue
```

Both produce the same `RewardConfig`, and the file is the only way to combine rewards. Weights
are 1/N when none are given, verbatim when all are, and an error when only some are. A single
reward at full weight is built and returned directly, so runs going on right now keep the
gradients they have today. The default is still `real_space_density` with the same flag
spellings, so nothing anyone has typed before stops working.

Marcus, this picks up most of your #319 review: reward names are a StrEnum, weights come from a
configuration rather than a flag per reward, dispatch is uniform, missing-input errors live in
the reward functions, and `prepare()` returns None. I didn't touch #319 itself and I'm not
asking anyone to redo work there. It rebases down to the reward class, an options dataclass and
one registry entry.

One thing worth knowing regardless of any of this: the structure-factor reward from #324 was
unreachable. Nothing in `src/` ever called its `prepare()`, so it could not run from a guidance
job at all. #373 fixes that on its own and is worth landing even if the rest gets argued about.

Three things I'd rather decide together than guess at:

- **Sign convention (#363).** Rewards are minimized, since guidance backprops the value and FK
  steering takes argmin. That's in the protocol docstring now and the contract tests enforce it,
  which means the geometry term in #313 has to flip. Renaming `RewardFunctionProtocol` to
[...]

PR #374, opened 2026-08-13
feat(rewards): add reward option schemas, a registry, and configuration parsing (#358)

manzuoni-astera:
## Summary

The data model for #358, with no CLI changes; those are in #375. Reward arguments currently live
in `add_generic_args` as if density were the only reward we will ever have, and reward
construction is hardwired to `RealSpaceRewardFunction`. Each reward now declares its options once
and registers a builder, so adding a reward type is a schema plus one registry entry rather than
an edit to shared argparse plumbing. Also adds `RewardConfig`, which reads the mapping from #358
as JSON, YAML or TOML, with relative weights.

## Changes

Three modules under `core/rewards/`:

`options.py` holds one frozen dataclass per reward. It is the single declaration that argparse,
the config-file schema, container path remapping and the error messages all read, so a flag, its
config key and its schema cannot drift apart. Each option carries metadata for help text,
choices, whether it is a path, and whether it arrives as JSON.

`registry.py` holds a `RewardSpec` per reward: its option schema, a builder addressed as a
`"module:function"` string, and the options the reward cannot be built without. The string
address matters, because this module has to import in every pixi environment to render `--help`
or validate a config, while `structure_factor.py` imports `SFC_Torch` and `reciprocalspaceship`
at module level. A reward's real dependencies are only touched when that reward is actually
built. Builders take `(options, *, device)` and nothing else: neither reward needs the input
structure at build time, and anything that depends on the model topology binds to it in
`prepare()`.

`config.py` holds `RewardConfig`. Weights are relative: `resolved_weights()` normalizes them to
sum to 1, omitting them all gives `1/N`, and giving only some raises, because a default silently
disagreeing with a number someone typed is worse than a complaint. A single reward is built and
returned unwrapped whatever weight it was given. Entry parsing lives in
`RewardEntry.from_mapping`, next to the schema check in `validated()`. YAML goes through
OmegaConf, already a dependency, so `${oc.env:VAR}` interpolation behaves the way it does in the
run presets. `to_mapping(remap_path=...)` writes the config for run metadata with path options
passed through the same container-to-host remapping as every other recorded path.

Builders live next to their rewards and raise their own errors for inputs they cannot do
without, naming both the flag and the config key, which is what Marcus asked for in #319 review.
`get_reward_function_and_structure` is deleted: loading the structure is reward-agnostic and
becomes `load_guidance_structure`, and its one caller builds the density reward through the new
builder. `RealSpaceRewardFunction`'s unused `selection` argument is now optional so the builder
does not have to fabricate one. `Rewards.STRUCTURE_FACTOR` joins the enum, which #324 never
added.

## Testing

`tests/rewards/test_reward_registry.py`, `test_reward_config.py` and `test_composite.py`.
Parsing is checked against the literal YAML from this issue's description, and the same
configuration written as JSON, YAML and TOML is asserted to parse identically.

The registry has a contract test that runs over every registered reward: its builder resolves,
its required options name real options, and anything whose name looks like a file declares
`path=True`. That last one is a hand-maintained bit of metadata and the easiest thing to forget
when adding a reward.

Weight resolution is covered including normalization, the partially-weighted error and the
all-zero error. Copilot caught on an earlier revision that a `reward_options` value that isn't a
mapping reached `dict()` and raised `TypeError` instead of a usage error; that is fixed with a
test.

## Rollout

Nothing user-facing. No new dependencies, no flags, no configuration anyone has to write, and no
change to how a run is invoked, since the CLI still goes through the old path until #375.

## Summary by CodeRabbit

* **New Features**
  * Added configurable reward selection from CLI options or JSON, YAML, and TOML files.
  * Added support for combining multiple rewards with normalized weighting.
  * Added structure-factor rewards and expanded real-space density reward configuration.
  * Added validation for reward types, options, required inputs, and weights.
  * Added improved structure loading for guidance workflows.

* **Bug Fixes**
  * Improved handling of optional selections and reward configuration defaults.

* **Tests**
  * Expanded coverage for reward composition, configuration parsing, validation, and structure loading.


src/sampleworks/core/rewards/options.py
```diff
@@ -0,0 +1,151 @@
...
+    Returns
+    -------
+    Any
+        The option's value type, e.g. ``float``, ``bool``, ``list[str]``.
+    """
+    hint = typing.get_type_hints(options_cls)[name]
+    args = [arg for arg in typing.get_args(hint) if arg is not type(None)]
+    if not args:
+        return hint
+    return args[0] if len(args) == 1 else hint
```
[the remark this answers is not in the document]
manzuoni-astera: Good catch, fixed. `option_type` now only unwraps unions, so a bare `list[str]` stays `list[str]`. No current option declares one, but the next person to write `expcolumns: list[str]` without `| None` would have silently lost `nargs` on its flag. Added a test with a bare generic to hold it.

src/sampleworks/core/rewards/config.py
```diff
@@ -0,0 +1,461 @@
...
+        """
+        entries = []
+        for name, raw_entry in data.items():
+            reward = get_reward_spec(name).name
+            entry = {} if raw_entry is None else raw_entry
+            if not isinstance(entry, Mapping):
+                raise ValueError(
+                    f"Configuration for reward '{reward.value}' must be a mapping with "
+                    f"'{WEIGHT_KEY}' and/or '{REWARD_OPTIONS_KEY}' keys, got "
+                    f"{type(entry).__name__}."
```
marcuscollins: Nit: the `__name__` reference is unnecessary.

src/sampleworks/core/rewards/config.py
```diff
@@ -0,0 +1,461 @@
...
+                )
+
+            unexpected = sorted(set(entry) - {WEIGHT_KEY, REWARD_OPTIONS_KEY})
+            if unexpected:
+                raise ValueError(
+                    f"Unexpected key(s) {unexpected} in the configuration for reward "
+                    f"'{reward.value}'. Reward options belong under '{REWARD_OPTIONS_KEY}'."
+                )
```
marcuscollins: I don't think you need to check this here, do you? Couldn't you check it after you create the RewardEntry, and then call `validated`? It would reduce the amount of different validation code we have, without making it less robust.

src/sampleworks/core/rewards/config.py
```diff
@@ -0,0 +1,461 @@
...
+                (spec.data_path_option, None if path is None else str(path)),
+                (spec.resolution_option, resolution),
+            ):
+                if option_name is not None and value is not None:
+                    options.setdefault(option_name, value)
+            entries.append(replace(entry, options=options))
+
+        return RewardConfig(tuple(entries))
+
+    def remapped_paths(self, remap: Any) -> dict[str, Any]:
```
marcuscollins: Isn't a similar method defined somewhere else? Can we re-use it?

src/sampleworks/core/rewards/config.py
```diff
@@ -0,0 +1,461 @@
...
+    -------
+    RewardFunctionProtocol
+        A single reward or a weighted combination of several.
+    """
+    weights = config.resolved_weights()
+    rewards = [
+        build_single_reward(entry.reward, entry.options, context) for entry in config.entries
+    ]
+
+    if len(rewards) == 1 and weights[0] == 1.0:
```
marcuscollins: If `len(rewards) == 1`, does it matter what the weight is? Shouldn't it default to 1 then?

src/sampleworks/core/rewards/options.py
```diff
@@ -0,0 +1,155 @@
...
+    exclude_free_reflections: bool = opt(False, help="Score the working set only, dropping R-free")
+    batch_partition: int = opt(10, help="Ensemble chunk size for the SFcalculator batch")
+    sfcalculator_kwargs: dict | None = opt(
+        None,
+        help="Extra SFcalculator keyword arguments as JSON, e.g. '{\"n_bins\": 15}'",
+        json_arg=True,
+    )
+
+
+def option_type(options_cls: type, name: str) -> Any:
```
marcuscollins: Does this actually get used anywhere? I thought I saw it, but I can't find it now. Maybe it will be used in #375? If it isn't used, remove.

src/sampleworks/core/rewards/registry.py
```diff
@@ -0,0 +1,203 @@
...
+from sampleworks.utils.guidance_constants import Rewards
+
+
+if TYPE_CHECKING:
+    import torch
+    from sampleworks.core.rewards.protocol import RewardFunctionProtocol
+
+
+@dataclass(frozen=True)
+class RewardBuildContext:
```
marcuscollins: I wonder if we actually need this if we have the .prepare method? Are we providing two different paths to the same outcome--i.e. do all of our Reward functions actually need to see a structure in order to set up the reward object, and if so, should we just always defer that to .prepare()?

src/sampleworks/utils/guidance_script_utils.py
```diff
@@ -288,26 +296,32 @@ def get_reward_function_and_structure(
...
+    density: str | Path,
+    device: torch.device,
+    em,
+    loss_order,
+    resolution,
+    structure_path: str | Path,
+) -> tuple[RealSpaceRewardFunction, dict[str, Any]]:
+    """Load structure and density inputs and build the real-space reward function.
 
+    .. deprecated::
```
marcuscollins: Add a deprecation warning.

marcuscollins: Also, if we aren't removing all calls to this method in this PR, can we create an issue to remove them and then remove this method?

tests/rewards/test_reward_config.py
```diff
@@ -0,0 +1,218 @@
...
+            }
+        }
+        config_file = tmp_path / f"rewards{suffix}"
+        if suffix == ".toml":
+            config_file.write_text(
+                "[structure_factor]\nweight = 1.0\n"
+                '[structure_factor.reward_options]\nmtzfile = "/data/x.mtz"\nbatch_partition = 4\n'
+            )
+        else:
+            config_file.write_text(json.dumps(mapping))  # valid YAML too
```
marcuscollins: Is this valid YAML? I wouldn't have thought so?

tests/rewards/test_reward_config.py
```diff
@@ -0,0 +1,218 @@
...
+
+        assert RewardConfig.from_file(config_file) == RewardConfig.from_mapping(mapping)
+
+    def test_yaml_resolves_environment_interpolation(self, tmp_path, monkeypatch):
```
marcuscollins, line 78: this seems like it is testing a behavior of OmegaConf, is that correct? If so, it doesn't seem like something we need to test?

tests/rewards/test_reward_config.py
```diff
@@ -0,0 +1,218 @@
...
+
+    def test_mapping_round_trips(self):
+        mapping = {
+            "real_space_density": {"weight": 0.4, "reward_options": {"loss_order": 1}},
+            "structure_factor": {"weight": 0.6, "reward_options": {"mtzfile": "/data/x.mtz"}},
+        }
+
+        assert RewardConfig.from_mapping(mapping).to_mapping() == mapping
+
+    def test_mapping_survives_json_and_pickle(self):
```
marcuscollins: These are behaviors of json and pickle, remove

tests/rewards/test_reward_config.py
```diff
@@ -0,0 +1,218 @@
...
+        """A flat entry is the most likely mistake; say where the options go."""
+        with pytest.raises(ValueError, match="Reward options belong under 'reward_options'"):
+            RewardConfig.from_mapping({"real_space_density": {"loss_order": 1}})
+
+    def test_empty_configuration_is_rejected(self):
+        with pytest.raises(ValueError, match="needs at least one reward"):
+            RewardConfig(())
+
+    def test_negative_weight_is_rejected(self):
+        with pytest.raises(ValueError, match="must be non-negative"):
```
marcuscollins, line 141: I think this has already been tested? L95 in test_composite--does that not hit the same code?

tests/rewards/test_reward_config.py
```diff
@@ -0,0 +1,218 @@
...
+
+    def test_a_single_reward_is_unweighted(self):
+        assert RewardConfig.single(Rewards.REAL_SPACE_DENSITY).resolved_weights() == (1.0,)
+
+    def test_given_weights_are_used_as_given(self):
+        config = RewardConfig.from_mapping(
+            {"real_space_density": {"weight": 2.0}, "structure_factor": {"weight": 3.0}}
+        )
+
+        assert config.resolved_weights() == (2.0, 3.0)
```
marcuscollins: I was wondering about this above--can you take a poll of Karson and Doris whether they think we should normalize the weights? I think we should, since I want to set the ratios here and then set a single weight, rather than try to remember that I have to multiply these weights by the global guidance weight.

tests/rewards/test_reward_config.py
```diff
@@ -0,0 +1,218 @@
...
+class TestExperimentalDataInjection:
+    """Grid search resolves the data file per protein, after the reward is chosen."""
+
+    def test_data_lands_in_each_reward_s_own_option(self):
+        config = RewardConfig.from_mapping(
+            {"real_space_density": {}, "structure_factor": {}}
+        ).with_experimental_data(path="/data/1vme.mtz", resolution=1.8)
+
+        by_reward = {entry.reward: entry.options for entry in config.entries}
+        assert by_reward[Rewards.REAL_SPACE_DENSITY] == {
```
marcuscollins: Oh, this is kind of weird, I didn't realize the behavior would be like this. I don't think there's ever a time we should assign the same data file to different kinds of rewards. So I think you may need to get rid of the .with_experimental_data method (and there's some method that figures out what the data argument should be for different rewards that probably is then unused.) But tbh I was never sure when this scenario should come up. 

I guess it solves the problem that you want to set up the generic arguments for a reward, and then later specify what data will be used? I think probably then .with_experimental_data needs to take the specific arguments that are passed to specific rewards (e.g., you pass "density" and that is interpreted as a density for real_space_density)

Anyway, whatever the case, we can't assign the same file to all the rewards in the config.

marcuscollins (changes requested):
Mostly nits and optional things, but the behavior of .with_experimental_data doesn't make sense and we'll have to fix that. LMK if you have questions.

manzuoni-astera:
Thanks Marcus. Everything is pushed; here is what changed and where I landed on each thread.

**`with_experimental_data`** is gone, along with the `data_path_option` / `resolution_option` fields on `RewardSpec` that existed only to serve it. You're right that one file should never fan out to every reward in the config. Nothing called it either: #375 reconciles the per-job `density` / `resolution` fields with the density entry directly, so there was no need for a reward-agnostic injection point. `RewardSpec` now declares `required_options` instead, which `missing_required_options()` was already reading but no spec had populated.

**Weights are normalized.** `resolved_weights()` scales them to sum to 1, so `{a: 2, b: 3}` and `{a: 0.4, b: 0.6}` are the same run, a single reward always has weight 1 whatever was written, and `build_reward` returns a lone reward unwrapped regardless (your `len(rewards) == 1` point). All-zero weights raise. @DorisMai @k-chrispens, Marcus asked for your take on this: the argument for normalizing is that you set ratios here and one global guidance weight on the scaler, instead of remembering to multiply. Shout if you'd rather have them used verbatim.

**`RewardBuildContext` is gone too.** You asked whether rewards actually need the structure at build time, and they don't. The structure-factor builder already ignored it, and the density builder only used it to compute an occupancy mask that `RealSpaceRewardFunction.__init__` stores and never reads (its own `# TODO: selection doesn't do anything right now`). Builders now take `(options, *, device)`, `build_reward` takes `device=`, and anything topology-dependent goes through `prepare()`. The density reward's `selection` argument became optional so the builder doesn't have to invent one; callers that pass it are unchanged. Making the density reward genuinely two-phase, with a real selection taken from `RewardInputs` in `prepare`, is the natural follow-up, but it changes behavior for every direct constructor call so I kept it out of this PR.

**Validation.** Entry parsing moved into `RewardEntry.from_mapping`, so the shape checks (entry is a mapping, only `weight` / `reward_options` keys, `reward_options` is a mapping) sit next to `validated()` in one class, and the three "must be a mapping" checks share one helper. The unexpected-key check can't live inside `validated()` because the entry never keeps the raw keys, but `RewardConfig.from_mapping` is a one-liner now and there is one place to read for what a file entry may contain. `__name__` dropped from the messages.

**`remapped_paths`** folded into `to_mapping(remap_path=...)`. The remap function itself is the existing `_remap_container_path` in `guidance_script_arguments`, which #375 passes in; `config.py` can't import it without a cycle, hence the callable. The only new logic is knowing which reward options are paths, and nothing existing did that.

**`option_type`** removed here. Its one caller is `add_reward_args` in #375, so it moves there.

**`get_reward_function_and_structure`** deleted outright rather than deprecated. Its one caller, `_run_guidance`, loads the structure and builds the density reward directly, so there is nothing left to track in an issue.

**Tests.** The format-agreement test writes real YAML now (JSON is technically valid YAML 1.2, but it wasn't a fair test of the YAML path). The json/pickle test is gone. The negative-weight test in `test_reward_config` stays alongside the one in `test_composite` because they hit different code: with normalization, `{a: -1, b: -1}` would become `(0.5, 0.5)` and never reach `CompositeReward`'s check, so the parse-time one is the one that matters. I kept the OmegaConf interpolation test but gave it a docstring saying why: it pins the loader choice, since a plain `yaml.safe_load` would hand `${oc.env:...}` to the reward as a literal path.

Also merged main (#400, #402); the only conflict was AGENTS.md. #375 gets rebased onto this once it's in.

marcuscollins (approved):
Maybe worth addressing a few things code rabbit found, just validation things.

PR #375, opened 2026-08-13
feat(cli): select and configure rewards via --reward-type and --reward-config

manzuoni-astera:
Stacked on #402 and #374 (#373 has merged), and opened against `main` for the same reason they
are: a fork branch can't be a base. The diff includes both. Review from `feat(cli): select and
configure rewards via --reward-type and --reward-config` onwards, and merge the other two first.

## Summary

Puts the registry from #374 on the command line. `--reward-type` picks a reward and brings that
reward's own flags with it, generated from its schema; `--reward-config` takes the same
configuration from a file and is the only way to combine rewards. Both produce one
`RewardConfig`, so going from one reward to two is a change to the run, not to the plumbing. The
default is unchanged, so existing command lines, presets and tests keep working. This is also
what finally makes the structure-factor reward from #324 runnable.

## Changes

```bash
sampleworks-guidance ... --reward-type structure_factor --mtzfile 1vme.mtz --bulk-solvent combined
sampleworks-guidance ... --reward-config rewards.yaml
```

The reward is resolved in the parser's existing first pass, beside `--model`, because it decides
which flags the second pass registers. Registering only the selected reward's options gets
cross-reward rejection for free: `--density` under `--reward-type structure_factor` is an
argparse error, the same way a Boltz flag already is on a Protenix run. Passing a config file and
a `--reward-type` together is rejected outright rather than one silently winning.

`--reward-type` defaults to `real_space_density`, and its flags keep their spellings, so nothing
anyone has typed before changes meaning. `tests/cli/test_guidance_cli.py` passes untouched, which
is the compatibility check that matters.

Two constraints shaped the serialization, and both are easy to get wrong:

`reward_config` serializes as a JSON string rather than a nested mapping. `as_dict()` is written
into the output CIF as the `sampleworks` category, and `add_category_to_cif` treats any
non-string iterable as a column of rows, so a nested dict there gets iterated into its keys and
produces a ragged category at the very end of an expensive run.

The flat `density`, `resolution`, `loss_order` and `em` fields stay on `GuidanceConfig`.
`run_grid_search.py` builds configs from them, `grid_search_eval_utils.py` reads density and
resolution back out of `job_metadata.json`, and job queues are pickled by one build and unpickled
by another. So the two representations are reconciled in one method, in whichever direction has
the information, and `__setstate__` fills in `reward_config` for pickles written before it
existed.

A reward missing a required input is reported as a usage error before a model is loaded. The
builders check the same thing for callers that arrive another way.

`CompositeReward` lands here too. `build_reward` returns a single reward at full weight as
itself, so current runs keep the gradients they have, and anything else becomes a weighted sum
with 1/N defaults. README and AGENTS.md gain a section on picking rewards and on what adding a
reward type involves.

## Testing

Full fast suite green, 972 tests. New CLI coverage for reward selection, config files in three
formats, cross-reward flag rejection, the config-plus-`--reward-type` conflict, the legacy pickle
path, and a metadata round trip that writes the output CIF and reads it back.

`tests/rewards/test_reward_build_integration.py` runs a command line through to a scored
structure for both rewards and for a two-reward config. That seam is what let #324 merge
unreachable, so it now has a test of its own.

CI is green: lint, four typecheck environments, four test environments.

Not covered: the GPU workflow has not run on any of the stack, so the structure-factor numerics
have only been exercised on CPU. The gpu-marked reward tests are 122 passed and 21 failed on my
laptop, with every failure being "Torch not compiled with CUDA enabled", and the same tests fail
identically on `main`. Someone with a GPU box should confirm before merge.

## Rollout

No migration. Existing command lines, preset TOMLs, pickled job queues and the evaluation scripts
all keep working, and the new flags are optional. Docs ship in this PR.

Two things to know after it lands. Grid search is still density-only: the proteins CSV is
`structure,density,resolution`, and structure factors would need it to carry an MTZ, with
`RewardConfig.with_experimental_data()` as the hook. And #319 conflicts with this by design; once
this is in, it rebases down to a reward class, an options dataclass and one registry entry.

Merge after #373 and #374.


src/sampleworks/core/rewards/config.py
```diff
@@ -0,0 +1,454 @@
...
+                )
+
+            weight = entry.get(WEIGHT_KEY)
+            entries.append(
+                RewardEntry(
+                    reward=reward,
+                    weight=None if weight is None else float(weight),
+                    options=dict(entry.get(REWARD_OPTIONS_KEY) or {}),
+                )
+            )
```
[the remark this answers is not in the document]
manzuoni-astera: Fixed. `reward_options` is checked for mapping-ness and raises a `ValueError` naming the reward and the expected shape, so a typo in a config file reads as a usage error instead of a `TypeError` traceback out of `dict()`.

src/sampleworks/utils/guidance_script_arguments.py
```diff
@@ -372,6 +455,50 @@ def __post_init__(self):
...
+                DEFAULT_REWARD_TYPE,
+                **{name: value for name, value in options.items() if value is not None},
+            ).to_mapping()
+            return
+
+        density_options = self.reward_config.get(DEFAULT_REWARD_TYPE.value, {}).get(
+            REWARD_OPTIONS_KEY, {}
+        )
+        self.density = density_options.get("density")
+        self.resolution = density_options.get("resolution")
```
smallfishabc, line 488: I think the structure factor and real space density process resolution differently. Will this compatible with both of them?

src/sampleworks/core/rewards/config.py
```diff
@@ -0,0 +1,461 @@
...
+        weights = tuple(float(entry.weight) for entry in self.entries)  # ty:ignore[invalid-argument-type]
+        total = sum(weights)
+        if abs(total - 1.0) > 1e-6:
+            logger.warning(
+                f"Reward weights sum to {total:g}, not 1. Using them as given; scale them "
+                "yourself if you meant them to be relative."
+            )
+        return weights
+
+    def with_experimental_data(
```
smallfishabc, line 360: I think this function is only called in test but not src or run grid search

src/sampleworks/core/rewards/composite.py
```diff
@@ -0,0 +1,138 @@
...
+import torch
+from jaxtyping import Float, Int
+from sampleworks.core.rewards.protocol import prepare_reward_if_needed, RewardFunctionProtocol
+
+
+if TYPE_CHECKING:
+    from sampleworks.core.rewards.protocol import RewardInputs
+
+
+class CompositeReward:
```
smallfishabc, line 17: RealSpaceRewardFunction implements precompute_unique_combinations, but CompositeReward doesn't. I am not sure about the impact

src/sampleworks/utils/guidance_script_utils.py
```diff
@@ -476,13 +459,12 @@ def _three_state_resolver(value: str | bool | None, default: bool) -> bool:
...
 def _run_guidance(args: GuidanceConfig, guidance_type: str, model_wrapper, device):
     """Run one configured guidance trajectory and save its outputs."""
-    reward_function, structure = get_reward_function_and_structure(
-        args.density,  # str/path to a map file.
-        device,  # this needs to come from the global context, not the args object.
-        args.em,
-        args.loss_order,
-        args.resolution,
-        args.structure,  # path/string to a structure file.
+    structure = load_guidance_structure(args.structure)
```
smallfishabc, line 462: There are some potential collision with my ITO code and I need to dig into it or we can discuss about how to make it compatible. 

#375 deletes get_reward_function_and_structure() entirely and replaces the reward-building step in _run_guidance() with load_guidance_structure() + build_reward(RewardConfig, RewardBuildContext). It also removes the flat --density/--resolution/--loss-order/--em CLI args in favor of --reward-type/--reward-config.
#313's new LATENT_OPT branch in _run_guidance() is written assuming the old function still exists and that reward_function/structure come from it before the if/elif dispatch on guidance_type

smallfishabc (commented):
I have several some suggestion and on of them may related to my ito code. I will keep reviewing it after I better understand the potential interaction of my code and this PR

src/sampleworks/core/rewards/composite.py
k-chrispens: This is fine, the precompute thing is a hold over from a while ago, and probably should be removed for now.

src/sampleworks/utils/guidance_script_arguments.py
k-chrispens: They process it right now roughly the same

src/sampleworks/utils/guidance_script_utils.py
k-chrispens: @smallfishabc depending on which merges first, that PR will need to address this, but not something this PR can address for now

k-chrispens (approved):
This reward configuration looks good to me, seems extensible enough. I suppose in the future it will need to be more integrated with the mmobs repo, but that is for a later time.

PR #376, opened 2026-08-14
fix(docker): force pixi environment materialization and pin pixi

### LINKED ISSUE #377: Image build red since 12 Aug: pixi reports a successful install and leaves an empty prefix
The public image build has failed on every run since 12 August. Three consecutive failures,
most recently today at 14:16, all at the same step and for the same reason: `pixi install`
reports success and leaves an empty prefix, and the guard added in #368 refuses to ship it.

```
#34 0.373 The boltz environment has been installed.
#34 0.487 The protenix environment has been installed.
...
#34 0.826 FATAL: pixi environment 'boltz' has no interpreter at /app/.pixi/envs/boltz/bin/python.
```

Five multi-gigabyte environments all "installed" in under a second, on a runner whose rattler
cache mount starts empty. The guard is working as designed. The underlying bug it was written to
catch, which had been silently shipping empty environments before that, is still there.

### Why this matters right now

The published `sampleworks` image is from 7 August. Anything merged since then is not in it,
including the SBGrid shell hooks from #362, and the actl catalog still pins the older digest. So
`--image sampleworks` will not pick up SBGrid even now that the `/programs` mount is live.

### What I ruled out

Not reproducible off the builder. On linux/amd64 with this exact manifest and lock:

- `boltz` installs correctly under pixi 0.76.2. Interpreter present, 100 seconds of real work.
- `default` installs correctly under 0.76.2 and 0.73.0.
- Warm rattler cache with the prefix deleted reinstalls properly rather than no-opping, so
  cached download state is not making pixi skip the work.
- pixi 0.73.0 emits the same `[system-requirements]` deprecation warning that appears in the
  failing log, so that warning is not evidence the builder is on a newer pixi.
- No earlier layer creates `/app/.pixi`: the `source` stage copies explicit paths only, so the
  `rm -rf` runs against a workspace that has no prefix to begin with.

### Next step

#376 pins the build's pixi version, which was the one unpinned input to an otherwise fully
digest-pinned image, and makes the guard print the prefix, its `conda-meta` bookkeeping and
pixi's own directory config before exiting. If the pin fixes it, we know the cause was a moving
pixi or a stale cached layer. If not, the next build log should say what pixi thinks it is doing.

manzuoni-astera:
Fixes #377.

## Summary

The image build has failed on every run since 12 August because `pixi install --frozen` reports success without running a package transaction. All five environments finish in under a second and `/app/.pixi/envs/<env>/bin/python` does not exist. The guard from #368 correctly stopped those empty environments from being published.

This stops relying on `install` deciding the prefix is stale. It uses Pixi's supported `reinstall` operation to force every locked package to be materialized, then starts each environment's Python before the image can be published.

## Changes

- Pin Pixi to v0.73.0, matching CI and the lock producer.
- Delete all five prefixes and run `pixi reinstall -e <env> --frozen --no-config`.
- Use project-specific, locked rattler and uv cache mounts.
- Fail immediately if reinstall fails, Python is absent, or Python cannot execute.
- Retain detailed prefix diagnostics for any unexpected empty result.

`--no-config` removes hidden system/user Pixi configuration as an input. `reinstall` bypasses the up-to-date path that produced the false successful installs, so this fix does not depend on cache rotation happening to repair builder state.

## Full validation

The complete candidate `public` image built successfully on astera-sh-builder without a push: https://github.com/Astera-org/docker-images/actions/runs/31830168103

That run used this PR branch and the real Dockerfile end to end: the digest-pinned CUDA base, authenticated digest-pinned checkpoint image, all checkpoint COPY and inventory checks, all five forced reinstalls, interpreter checks, source cleanup, runtime scripts, and final public stage. Every environment executed Python 3.12.13. The full build completed in 19 minutes.

A separate stress run deleted and rebuilt every prefix twice in one BuildKit step, first with an empty package cache and then with it warm: https://github.com/Astera-org/docker-images/actions/runs/31827709692

- Cold package cache: boltz 129s, protenix 111s, rf3 124s, protpardelle 113s, analysis 123s.
- Prefixes deleted, warm package cache: boltz 32s, protenix 102s, rf3 102s, protpardelle 106s, analysis 120s.
- All ten reinstalls executed Python 3.12.13.
- Final prefixes were 13-14 GB each, rather than the previous 20 KB metadata shells.

Also validated locally:

- `docker build --platform linux/amd64 --target source-check ...`
- The resulting image carries Pixi 0.73.0.
- Pixi 0.73.0 `reinstall` recreates a deleted Linux prefix and starts its interpreter.
- `git diff --check` passes.

All twelve repository checks and CodeRabbit are green on the final commit. The first post-merge run on `diffuse-sh-builder` is now a publish confirmation rather than the first full build test.

## Rollout

No coordination is required. The changed RUN instruction invalidates the bad environment layer. The build will be slower than the sub-second broken path because it now performs the required package transactions. Once the image publishes, the actl sampleworks digest should be reconciled so the SBGrid shell hooks from #362 reach users.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -135,20 +135,13 @@ RUN --mount=type=bind,from=checkpoints,target=/ck \
 # Docker layers duplicates shared conda packages (numpy, CUDA libs, etc.) and can
 # add tens of GB to the image.
 #
-# The `rm -rf` and the assertion below are both load-bearing. Published images
-# have shipped with all five environments as empty shells: the directories and
-# pixi's own bookkeeping in conda-meta present, but no package records, no bin/,
-# no lib/. `pixi install --frozen` then reports "The <env> environment has been
-# installed" against them and exits 0, so the breakage is invisible at build
-# time and only surfaces when a scientist finds /app/.pixi/envs/protenix has no
-# python. `SAMPLEWORKS_REQUIRE_PREBUILT_PIXI=1` makes the runner refuse to fall
-# back, so the image promises environments it does not carry.
-#
-# Clearing the prefix first forces a genuine install even when a stale or stub
-# prefix arrives from a cached layer, the base image, or the registry
-# buildcache. The per-environment check then makes an empty env fail the build
-# instead of shipping: a directory that exists but has no interpreter is exactly
-# the state that got published, and `pixi install` alone does not catch it.
+# Published images have shipped all five environments as empty shells:
+# conda-meta/pixi exists, but there are no package records, bin/, or lib/.
+# `pixi install --frozen` reports success against that state without running a
+# package transaction. `pixi reinstall` is the supported command that bypasses
+# that up-to-date decision and materializes every package again. It also creates
+# a missing prefix, so clearing prefixes first covers both stale layers and the
+# empty-prefix failure mode seen on diffuse-sh-builder.
 #
 # `/root/.cache/pixi` is deliberately not cached across builds — it carries
 # pixi's own "is this environment current" state, which is the thing that can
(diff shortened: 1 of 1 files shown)


manzuoni-astera:
Tracking the failure itself in #377.

xraymemory (approved):

PR #378, opened 2026-08-15
fix(synthetic): drop setup_entities() from atomarray_to_gemmi so written cifs round-trip chain_info

DorisMai:
## What

`atomarray_to_gemmi()` called `structure.setup_entities()` before writing. That fabricates entities with an empty `full_sequence`, so the written cif carries `_entity` / `_entity_poly` but no `_entity_poly_seq`. Atomworks reads that partial block and then raises `KeyError` in model wrappers accessing fields like `processed_entity_canonical_sequence`.

Dropping the call means gemmi writes no entity block at all, and atomworks infers the sequence from `_atom_site` instead — which is what the model wrappers need.

Also refactored `parse_structure` function in `atom_array_utils.py` to more consistently reuse production kwagrs for `parse`:
```
hydrogen_policy="remove",
add_missing_atoms=False,
ccd_mirror_path=None,
```
A few call sites are affected: `guidance_script_utils.py` and `rscc_grid_search_script.py`.

## Cost

`chain_info` loses `rcsb_entity`. Only Protenix reads it, and it falls back to the chain id.

`residue.subchain` now carries the sole assignment of `label_asym_id` (nothing else sets it once `setup_entities()` is gone), so it must stay single-char — SFcalculator's PDB-header step rejects the multi-char subchain ids `setup_entities()` invents. Comment updated at the assignment site to say so.

## Tests

Adds round-trip coverage in `tests/synthetic/test_generate_synthetic_sf.py` asserting the written cif reloads with intact `chain_info`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved mmCIF output to preserve single-character subchain IDs.
  * Prevented incomplete or fabricated entity metadata from being written.
  * Preserved chain metadata, atom identity, residue names, sequences, counts, and IDs when structures are saved and reloaded.

* **Tests**
  * Added coverage verifying metadata and atom identity survive mmCIF round trips.

### DIFF
--- a/tests/synthetic/test_generate_synthetic_sf.py
+++ b/tests/synthetic/test_generate_synthetic_sf.py
@@ -41,7 +41,19 @@ CROSS_MODEL_CHAIN_INFO_KEYS = ("chain_type", "processed_entity_canonical_sequenc
 
 
 def _parse_at_production_kwargs(path: Path) -> dict:
-    """Parse a structure with the kwargs ``run_guidance`` uses."""
+    """Parse a structure with the kwargs ``run_guidance`` uses.
+
+    Parameters
+    ----------
+    path : Path
+        Path to the structure file to parse.
+
+    Returns
+    -------
+    dict
+        The parsed Atomworks structure, keyed by ``"asym_unit"``, ``"chain_info"``, and the
+        rest of the parse metadata.
+    """
     return parse(path, hydrogen_policy="remove", add_missing_atoms=False, ccd_mirror_path=None)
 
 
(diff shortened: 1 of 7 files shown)


tests/synthetic/test_generate_synthetic_sf.py
```diff
@@ -33,6 +35,27 @@
 
 DMIN = 2.0
 
+# chain_info fields the model wrappers read: chain_type (all), the canonical sequence
+# (Boltz/Protpardelle polymer YAML), and res_name (Boltz ligand CCD code).
+CROSS_MODEL_CHAIN_INFO_KEYS = ("chain_type", "processed_entity_canonical_sequence", "res_name")
+
+
+def _parse_at_production_kwargs(path: Path) -> dict:
```
marcuscollins: Do we not have a method already that does this parsing?

marcuscollins (approved):
Please search around thoroughly to see if we don't already have a method that makes the call to parse with default options like you have here. If it doesn't exist... maybe make it? Otherwise approved.

PR #380, opened 2026-08-18
feat(synthetic): forward calculation of Bragg and diffuse via lunus.sf

mewall:
Adds a second structure-factor engine alongside the SFcalculator path, built on
the differentiable splat/FFT code in `lunus.sf`, plus the shared adapter the rest
of the diffuse work sits on.

`core/forward_models/xray/lunus_sf.py` translates between sampleworks structures
and the lunus engine: gemmi-derived space-group operations (lunus ships only a
cctbx path, and cctbx is not in these environments), symmetry-commensurate grid
sizing, per-atom scattering kernels built from the elements actually present, and
Cartesian to fractional conversion. Setup is coordinate-independent and built
once, matching the two-phase pattern `StructureFactorRewardFunction` uses.

`synthetic/generate_synthetic_sf_lunus.py` writes the same MTZ layout as
`generate_synthetic_sf.py`t is interchangeable. It
takes ensembles rather than altlocs: a multi-model input gives one configuration
per model and the emitted altloc collapse the
SFcalculator script performs cannot express the second moment, which is what
diffuse targets need. `--wecond moment
(`<|F|²> - |<F>|²`) with an `ID` intensity column, matching lunus xtraj's
convention; `--altlocs-as-d multi-conformer structure
into the ensemble it already is, which is how a nonzero diffuse target comes out
of the single-model files

lunus **is** declared in ` tag `v0.2.0` with the `sf`
extra, and resolved in `pixi.lock`. Every pixi environment gets it; no manual
install step is part of no_lunus.sh` remains only for
the case the lock cannot cover — testing a lunus commit newer than the pin.

The lock is regenerated by the Relock workflow, never locally: `pixi lock` is
all-or-nothing, with no `-rm` selector, and solving
linux-64 needs a linux-64 prefix. The cause is sdist-only transitive
dependencies — `ihm` publion, and every `modelcif` requires it — so this is permanent rather than transitional, and it is not
caused by the `sampleworks` editable path dependency (which declares no dynamic
metadata, so pixi reads it statically and never builds it). See
`DIFFUSE_SCATTERING_PLAN.md`.

Verified on 1VME at 1.8 Å (P 1 21 1, 86,499 reflections):

- vs SFcalculator, identical atoms → correlation 1.000000, R 0.0002
- P1 vs P2₁ reflection counts 170,595 / 86,499, the group order
- identical configurations → diffuse at 4e-8 of Bragg intensity
- `<F>` over N copies vs one copy → 9e-8 relative
- translation under symmetry → 0.72, vs 9e-4 in P1
- bulk solvent mask occupancy 0.494, taper over 13.4%

The last of those corrected the plan: diffuse-only scoring CAN see the mean
structure in a real crystal. The invariance argument holds in P1 alone —
translating an asymmetric unit moves a mate at `Rx+t` to `Rx+t+Rd`, so packing
changes and most of the diffuse signal with it.

Also included since the original description: the lunus declaration and two
relocks, the pin moved from a rev to the `v0.2.0` release tag,
`scripts/install_lunus.sh` fixed to scope its `--force-reinstall` with
`--no-deps` and to read the pin from `pyproject.toml` rather than duplicate it,
and `SF_REWARD_LUNUS_BACKEND.md`, a suggestion to the `dm/sf-reward` owners for
using this engine behind the structure-factor reward.

Claude-Session: https://claude.ai/code/session_017gzg6C5T8hgaHCmp6h3PZi

### DIFF
--- a/DIFFUSE_SCATTERING_PLAN.md
+++ b/DIFFUSE_SCATTERING_PLAN.md
@@ -1,0 +1,455 @@
+# Plan: Diffuse Scattering as a Guidance Target (lunus.sf)
+
+Status: **built, wired, and running end to end on the pod.** A full guided run
+completes; what it does not yet do is converge — see Phase 4a. The recovery test
+(Phase 5) is the remaining milestone. Written 2026-08-12; last updated
+2026-08-18.
+
+Adds diffuse scattering as an experimental target for guidance, using the
+differentiable structure-factor engine in `lunus/lunus/sf/`.
+
+## Why
+
+Every reward in sampleworks today scores a **first moment** of the ensemble.
+`RealSpaceRewardFunction.__call__` ends in `.sum(0)` over the batch with
+occupancy `1/N` (`core/rewards/real_space_density.py:311`), which is a mean
+density; `StructureFactorRewardFunction` does the analogous complex sum. The
+ensemble is generated and then averaged away.
+
+Diffuse scattering is the **second moment**: `<|F|²> − |<F>|²`. It is the part
+of the experiment that actually reports on conformational spread, and
+`lunus.sf.structure_factor_torch.mean_and_diffuse()` computes it
+differentiably. lunus's "configurations" axis maps directly onto sampleworks'
+ensemble batch axis.
+
+The engine choice is also about scale: the `SFC_Torch` path is memory-bound on
+its ASU-grid batch and does not reach large systems. lunus is the scale-oriented
+path.
+
+## Decisions (locked)
+
+| Decision | Choice |
+|---|---|
+| Scope of the lunus reward | **Diffuse only** — score `<\|F\|²> − \|<F>\|²`; all amplitude targets stay with SFC_Torch |
+| Existing `StructureFactorRewardFunction` | **Untouched** — separate effort, not in scope here |
+| Batched configurations | **Landed in lunus** — `structure_factors_batch()`, see below |
+| Packaging (`lunus.sf` importable) | **Done in the lunus repo** — PEP 621 `pyproject.toml`, installable as `lunus[sf]` |
+| First target data | **Synthetic**, from a known multi-state ensemble |
+
+## Prior art in this repo — read before building
+
+- `core/rewards/structure_factor.py` — the closest existing analogue. Establishes
+  the **two-phase construction** pattern (`__init__` stores config; `prepare()`
+  builds everything needing the model atom array and device) that the diffuse
+  reward should reuse. Also the pluggable-loss pattern (`AmplitudeLoss`, line 24).
+- It is **complete and tested but not wired into the pipeline**: `.prepare()` is
+  called only from tests, and `get_reward_function_and_structure()`
+  (`utils/guidance_script_utils.py:295-311`) hardcodes `RealSpaceRewardFunction`.
+  Phase 4 below builds the missing hook generically, so that effort can use it too.
+- `core/rewards/protocol.py` — `RewardInputs`, and the `occupancies = 1/N`
+  convention at line 121. See "Correctness traps".
+- `utils/atom_reconciler.py` — alignment into the crystal frame. Diffuse is no
+  more SE(3)-invariant than density; assume coordinates arrive pre-aligned.
+
+## Upstream state (lunus, as of 2026-08-16)
+
+The adapter is written against this API:
+
+```python
+from lunus.sf import structure_factors_batch, mean_and_diffuse, SolventModel
+
+structure_factors_batch(
+    frac_coords_batch,    # (n_configs, n_atoms, 3), differentiable
+    element_idx, occ,     # (n_atoms,) shared, OR (n_configs, n_atoms) per-config
+    atom_A, atom_lam, elem_offsets, atom_radius_ang,
+    grid_shape, orth_matrix, cell_volume, hkl, taper_width,
+    blur=0.0, max_atoms_per_batch=50_000, grid_ops=None, compile_core=True,
+    use_checkpoint=False, solvent=None, supercell=None,
+) -> (n_configs, n_refl) complex
+```
+
+`lunus/lunus/sf/__init__.py` re-exports lazily (PEP 562), so the import above
+costs nothing until first use. `mean_and_diffuse()` takes the result directly.
+
+### Per-configuration occupancies
+
+`occ` may be `(n_atoms,)` (shared) or `(n_configs, n_atoms)` (per member), and
+**gradients flow to `occ` as well as to coordinates**. Two consequences for this
+work:
+
+- Unequal ensemble populations are now expressible at the engine level. The
+  `1/N` in `RewardInputs` is a sampleworks *choice*, no longer a constraint
+  imposed from below. Refining populations is out of scope here, but this is
+  where it would start.
+- Varying hydration between members is expressed as one array slot at two
+  occupancies, never as atoms entering and leaving. The atom array must
+  correspond across configurations — same slots, same count — which the
+  reconciler already guarantees.
+
+### Bulk solvent
+
+`lunus/lunus/sf/solvent_torch.py` implements the flat/mask model:
+`solvent_mask`, `f_solvent`, `f_total`, `calibrate_cutoff`, `mask_occupancy`,
+`shell_voxels`, and a `SolventModel` config object. Passing `solvent=None`
+leaves today's behaviour bit-identical.
+
+**Masks are applied per configuration, and that is the only choice that reaches
+diffuse.** One mask shared across the ensemble has zero variance: it changes
+`<F>` but contributes exactly nothing to `<|F|²> − |<F>|²`. Per-configuration
+masks fluctuate anti-correlated with the protein — where an atom moves out,
+solvent moves in — which is the excluded-volume contrast correction. Note this
+decides the `mask(⟨ρ⟩)` vs `⟨mask(ρ)⟩` question the opposite way from SFC's
+`bulk_solvent="combined"` default.
+
+Two knobs to be aware of before using it:
+
+- `mask_blur` is this model's probe radius (default σ ≈ 1.13 Å). Setting it to
+  `0.0` restores an unsmoothed threshold and does **not** reproduce conventional
+  solvent scales on real data.
+- `detach_mask` defaults to `False`, which is the consistent choice — measured
+  4.8× lower refinement loss on 7FPV with the mask live. See the trap below
+  before changing it.
+
+**Memory is solved; throughput is not.** `structure_factors_batch` is still a
+Python loop over configurations, with `use_checkpoint=True` recomputing each
+splat during backward instead of retaining it. Measured (3000 atoms, 90³ grid,
+peak RSS above baseline):
+
+| N | retained | checkpointed |
+|---|---|---|
+| 8 | 839 MB | 361 MB |
+| 16 | 1604 MB | 371 MB |
+
+Flat in N rather than linear, for ~2.4× the time, with bit-identical gradients
+(`lunus/lunus/sf/tests/test_batch.py`). Fusing the splat into one kernel remains
+open upstream, as does checkpointing only the members that do not fit rather
+than all of them; both are listed under "Not yet built" in
+`lunus/lunus/sf/README.md`. So guidance should assume **N sequential splat+FFT
+passes per guided step**, with `use_checkpoint=True` once N or the grid is large.
+
+### Dependency wiring (done)
+
+lunus is **not declared in `pyproject.toml`**. Install it with
+`scripts/install_lunus.sh`, which pins the same rev.
+
+That is a workaround for a workspace-wide constraint, not a preference:
+
+- **This workspace cannot be re-locked on any single machine.** Every
+  environment carries `sampleworks` as an editable path dependency, and a source
+  dependency must be *built* to yield metadata — per platform. macOS fails
+  solving linux-64; the Linux runner fails solving osx-arm64 for `boltz-osx`.
+  `pixi lock` is all-or-nothing, with no `--environment` or `--platform`
+  selector, so neither can produce a complete lock. Declaring any new dependency
+  therefore requires a lock nobody can generate. This predates lunus and is not
+  specific to it.
+- **It worked once.** The relock that first added lunus succeeded and resolved
+  it into all 13 environments including osx-arm64 ones. Every attempt since has
+  failed at `boltz-osx`. The likeliest explanation is that adding a dependency
+  is an incremental update while moving one between features forces a full
+  re-resolve — unverified.
+- **Two ways out, both outside this work.** Drop `osx-arm64` from the workspace
+  (a repo-owner decision — CI never builds macOS, so the platform exists purely
+  for local development), or have lunus publish a wheel: `lunus.sf` is pure
+  Python, so one `py3-none-any` artifact needs no build anywhere and the
+  constraint disappears for this dependency at least.
+- **Only ever re-lock with the pinned pixi version.** The Relock workflow pins
+  v0.73.0. Others rewrite the lock wholesale — 0.76.2 stripped the `osx-arm64`
+  entries, 0.71.2 produced a ~9,900-line diff. A lock change not confined to the
+  dependency you touched is the signal to stop.
+
+`scripts/install_lunus.sh` covers environments where the pixi declaration is not
+in play; it becomes redundant once the declared route is verified end to end.
+
+## Phases (sampleworks)
+
+### Phase 1 — adapter
+
+New: `core/forward_models/xray/lunus_sf.py`. Resolves the impedance mismatch
+between sampleworks' reward inputs and lunus's calling convention:
+
+- Cartesian Å → fractional (`inv(M) @ x`; differentiable, trivial).
+- sampleworks scattering indices (`utils/elements.py`) → element symbols for
+  `lunus.sf.elements.IT92_COEFFS`.
+- Cell + space group → `orth_matrix()`, `grid_shape_for_resolution()`,
+  `adjust_grid_for_symmetry()`, `build_grid_ops()`.
+- Per-atom B via `build_atom_kernels_torch` (B is fixed at setup; guidance only
+  needs ∂/∂coords).
+
+**The one piece with no existing implementation**: `build_grid_ops_from_cctbx`
+requires cctbx, which is not in the sampleworks pixi envs. A gemmi-based
+equivalent has to be written — gemmi (pinned at `pyproject.toml:214`) exposes the
+same rotations and translations.
+
+All of this is coordinate-independent, so it lives behind
+`prepare(atom_array, device)`.
+
+### Phase 2 — the reward
+
+New: `core/rewards/diffuse_scattering.py`, satisfying `RewardFunctionProtocol`.
+Scores the variance term only. Loss pluggable:
+
+- **MSE** for the synthetic milestone — target and calculation come off the same
+  engine, so scales match by construction.
+- **Resolution-shell-binned Pearson CC** for real data, which arrives on an
+  arbitrary scale. Precedent: `lunus/lunus/sf/xtraj.py:1562` scores experimental
+  diffuse by correlation, not residual.
+
+A CC loss has different gradient character than the density reward, so DPS step
+sizes will not transfer — expect to re-tune `--step-size`.
+
+### Phase 3 — target ingestion (deferred)
+
+Not needed for the synthetic milestone: targets are generated at integer hkl by
+the same path that evaluates them.
+
+For real data later: lunus lattices sample reciprocal space at sub-integer
+intervals (`points_per_hkl`), while `compute_fcalc` extracts integer Miller
+indices off the FFT grid, so the target needs reduction to integer hkl — see
+`xtraj.py:1559` (`common_sets`). Read targets via `reciprocalspaceship`/gemmi
+(both already dependencies) rather than cctbx's `any_reflection_file`.
+
+### Phase 4 — pipeline wiring
+
+- `--target-type {density,diffuse}` and dispatch in
+  `get_reward_function_and_structure()`.
+- A `prepare()` hook for two-phase rewards, in the trajectory scalers right after
+  `process_structure_to_trajectory_input()` (`core/scalers/pure_guidance.py:83`,
+  `core/scalers/fk_steering.py:108`) — the point where the model atom array and
+  device are both known. Build it generically so the SFC reward can use it later.
+
+### Phase 4a — what the first guided run showed
+
+A full `sampleworks-guidance --target-type diffuse` run completed end to end
+(Boltz-2, 1VME, N=4, 200 steps, `bragg_weight=0.5`). The wiring works. The
+science did not: **loss rose from 0.7669 to 0.7999**. Two causes, both
+configuration rather than code, and both must be fixed before Phase 5 means
+anything.
+
+**Alignment was off.** The run used `align_to_input=False`. The reward scores in
+the crystal frame and the model generates in an arbitrary one, so the gradient
+was computed on a structure in the wrong frame. `--augmentation
+--align-to-input` are not optional for any reciprocal-space target — the same
+requirement the density reward has, per AGENTS.md's alignment section.
+
+**The model and the target were chemically different objects.** The log reported
+`model=6613, structure=6462, common=4515`, and the model's elements came back as
+`['C', 'Fe', 'N', 'O', 'S']` — no **Se**. 1VME is selenomethionine; Boltz models
+MSE as ordinary methionine, so the model has sulfur where the target has
+selenium, and the target additionally contains waters and ligands the model never
+produces. Roughly a third of the atoms on either side are unmatched.
+
+That puts a floor under the loss that no amount of guidance can remove, and it is
+not specific to diffuse — it applies to any target generated from a deposited
+structure and scored against a model's topology.
+
+**So the recovery test must generate its target from the model's own atom array**,
+not from the deposited file. Concretely: run the model once, take
+`model_atom_array`, generate Bragg and diffuse targets from a perturbed ensemble
+of *that*, then guide. Anything else measures the MSE/MET discrepancy rather than
+the reward.
+
+### Phase 5 — tests
+
+The **recovery test** is the milestone that says the whole thing works: build a
+synthetic two-state ensemble via `synthetic/`, compute its diffuse, initialize a
+mismatched ensemble, confirm guidance drives the diffuse loss down and the
+ensemble spread toward truth.
+
+Cross-engine validation against SFC_Torch is out of scope under diffuse-only —
+SFC has no diffuse term. Parity of the underlying `|F|` is already covered by
+lunus's own gemmi comparison (`lunus/lunus/sf/tools/compare_icalc_mtz.py`).
+
+### Phase 6 — cost
+
+Measured 2026-08-17/18 on 1VME: 6,819–7,092 atoms, grid 96×160×160, 86,499
+reflections, 2 configurations. A100 for the GPU figures, pod CPU otherwise.
+
+| phase | cost | |
+|---|---|---|
+| `build_setup` | 0.09–1.7 s | once per structure; BLAS-bound, see threading below |
+| anisotropic transform build | 46–139 ms | 24 shells, 8 MB basis |
+| target load, intersect, resolution cut | ~0.05 s | after vectorizing; minutes before |
+| forward pass, GPU, **first call** | ~3.8 s | dominated by `torch.compile` warmup |
+| forward pass, GPU, **steady state** | **~0.08 s per configuration** | measured in a guided loop |
+| forward pass, CPU | ~4–5 s | what makes the reward tests ~3 min |
+
+**The steady-state figure is the one that matters, and it is ~50× better than the
+first-call figure.** A guided Boltz-2 run — 200 steps, N=4, reward and backward on
+every step — took **62 s total, 3.19 it/s**, including the model's own forward
+pass. An earlier revision of this section extrapolated ~30 s/step and "hours per
+trajectory" from a single generator invocation; that measurement paid the entire
+compile warmup and never amortized it, which a 200-step loop does immediately.
+
+Consequences: cost is **not** currently a reason to reach for the fused batched
+splat or for `use_checkpoint`, and the time-dependent conditioning lever
+(AGENTS.md §5) is available if wanted but not needed at this system size. Both
+become relevant again on a much larger cell, where the grid rather than the
+warmup dominates.
+
+#### Threading
+
+`OMP_NUM_THREADS` matters only where the compute is on the CPU, and then it
+mostly speeds up *setup* rather than the hot loop:
+
+- **Generator on GPU**: 4.017 / 3.893 s unset vs 3.744 s at `OMP=2` —
+  indistinguishable, since only host-side work is on the CPU.
+- **Reward tests on CPU**: 232.5 s → 179.7 s, **−23%**. Concentrated in
+  `build_setup` (1.67 s → 0.086 s) and the spline basis (139 ms → 46 ms), both
+  BLAS-bound. A forward-pass-dominated test went 22.7 s → 24.2 s, i.e. slightly
+  worse: the compiled splat manages its own threading.
+
+Single samples with a warm cache on the second, so treat 23% as indicative.
+lunus reports a much larger effect on a more heavily oversubscribed pod — splat
+287.9 ms/frame at 104 threads on 3 CPUs vs 66.3 ms at `OMP=2` — because threads
+beyond the cgroup quota spin-wait and get the whole process throttled. BLAS reads
+the environment before the process starts, so exporting the variable is not the
+same as calling `torch.set_num_threads`.
+
+#### Memory
+
+The anisotropic basis is `n_refl × n_bins`: 8 MB here, but **~330 MB** for the
+1.57M-reflection xtraj map, and it is built in float64 before being cast, so peak
+setup is roughly 3×. `prepare()` logs the size.
+
+Upstream, checkpointing makes the ensemble splat flat in N rather than linear:
+839 MB → 361 MB at N=8, 1604 MB → 371 MB at N=16.
+
+#### The noise floor moves between runs
+
+Three independent sightings: the identical-ensemble diffuse RMS ratio moved
+3.1e-8 → 4.4e-8; a mixed-weight recovery loss came out 4.53e-10 against 3.08e-10
+predicted from its pure-diffuse counterpart; and a regenerated diffuse target's
+minimum went from exactly 0 to −0.00391. Reduction order varies and diffuse is a
+difference of nearly-equal large numbers. Individual near-zero diffuse values are
+unresolved — which is also why the generator does not clip negatives, since
+clipping would hide exactly this and bias the target upward where the signal is
+weakest.
+
+## Correctness traps
+
+### Occupancy convention
+
+`RewardInputs` sets `occupancies = 1/N` (`core/rewards/protocol.py:121`), a
+convention the sum-based rewards depend on. The diffuse reward must **divide it
+out** and splat each configuration at full occupancy before averaging. Otherwise
+both moments scale by `1/N²` and the variance term collapses toward zero — which
+presents as "guidance does nothing" rather than as an error. Assert the
+convention explicitly rather than silently correcting it.
+
+Note this is now a sampleworks-side convention only: lunus accepts
+per-configuration occupancies and differentiates them (see "Upstream state"). The
+`1/N` flattening is a choice this pipeline makes, and the place to revisit if
+unequal ensemble populations ever become a target.
+
+### The occupancy gradient is incomplete when the mask is detached
+
+With `SolventModel(detach_mask=True)`, `d(F_total)/d(occ)` carries the protein
+term but **not** the bulk term that replaces it as an occupancy falls. Forward
+the two stay continuous; in the gradient they do not. The default is `False`, so
+this only bites if something turns detaching on — most plausibly as a memory
+measure inside a guidance run. Assert rather than document: a run that both
+refines occupancies and detaches the mask is computing a gradient of a different
+model than the one it evaluates.
+
+### Diffuse-only guidance CAN see the mean structure — in a real crystal
+
+An earlier draft of this plan claimed the opposite, at ~85% confidence, and
+flagged it for numerical checking. **The check was run on 2026-08-17 and the
+claim was wrong.**
+
+The invariance argument holds only in P1. A common rigid translation multiplies
+every `F_b` by one phase factor, leaving `<|F|²>` and `|<F>|²` unchanged —
+measured relative deviation 9.2e-4 on 1VME chain A recomputed in P1, and that
+residual is grid discretization (atoms moving relative to voxel centres), not
+round-off: float32 noise in the same pipeline measures 3.1e-8.
+
+In a real space group it fails, because translating the asymmetric unit is not
+translating the crystal. A symmetry mate at `Rx + t` moves to `Rx + t + Rd`, so
+the packing changes. Equivalently, in
+
+```
+F_total(h) = Σ_g exp(2πi h·t_g) · F_ASU(h R_g)
+```
+
+a translation by `d` multiplies each term by `exp(2πi (h R_g)·d)` — a phase
+depending on `g`, which therefore cannot factor out of the sum. Measured on 1VME
+(P 1 21 1) at 1.8 Å: relative deviation 0.72, roughly 800× the P1 residual.
+
+**This is good news for the method.** Packing contacts against symmetry mates make
+absolute position observable, so diffuse-only scoring is not the blind-to-the-mean
+reward this plan assumed. The argument for needing a composite reward purely to
+pin down the mean is correspondingly weaker — a composite may still be wanted for
+signal strength or for combining data types, but not for that reason.
+
+Both regimes are pinned by
+`tests/synthetic/test_generate_synthetic_sf_lunus.py::TestSelfConsistency`.
+
+## Related work: the lunus-backed synthetic generator
+
+`src/sampleworks/synthetic/generate_synthetic_sf.py` (SFcalculator-backed) is
+being ported to lunus as a **separate script first, merged behind an `--engine`
+flag later**. It is worth doing before Phase 1 because it exercises every adapter
+primitive — cell → orthogonalization matrix, space group → grid ops via gemmi,
+elements → IT92, per-atom B → kernels, resolution → grid sizing — forward-only,
+non-differentiably, against an oracle (SFC) that already works. Build those
+shared pieces in `core/forward_models/xray/lunus_sf.py` and have the script
+import them.
+
+Two things settled since that discussion:
+
+- **Bulk solvent is no longer a blocker.** `--simulate-solvent-and-scale` has a
+  real lunus path now, so the lunus generator can emit both protein and total
+  sets rather than rejecting the flag.
+- **The generator should take a multi-model structure, not altlocs.** The
+  existing script collapses altloc conformers into one `Fprotein` (deliberately —
+  it matches the reward's `bulk_solvent="combined"` mode). That collapse destroys
+  the second moment, so it cannot produce a diffuse target at all. An
+  `AtomArrayStack` input — the same form guidance *outputs* — maps one model per
+  configuration and feeds `structure_factors_batch` directly. Leave the altloc
+  path on the SFC generator; the reward fixtures in `tests/rewards/conftest.py`
+  depend on it.
+
+Validation: same structure through both engines, compared with lunus's own
+`tools/compare_icalc_mtz.py` (correlation and R-factor, overall and by shell).
+Expect agreement at the level of lunus-vs-gemmi, ~0.999989 correlation and
+R ≈ 0.0077 — the taper is the main source of the difference. That ~0.8% is a
+floor under any fit that pairs a lunus-generated target with the SFC-based
+reward, so prefer to keep engine pairs consistent.
+
+## Sequencing
+
+**Done and verified on the pod:**
+
+1. lunus: packaging, batched entry point, bulk solvent, per-configuration
+   occupancies, and the anisotropic-component extraction (`sf/aniso.py`,
+   `sf/aniso_torch.py`, with `to_aniso` refactored onto it and pinned bit-exact).
+2. Phase 1, the adapter — cross-checked against SFcalculator at correlation
+   1.000000, R 0.0002 over 86,499 reflections.
+3. The lunus-backed synthetic generator, including `--write-diffuse` and
+   `--altlocs-as-models`, which is what produces a nonzero diffuse target from
+   the single-model files in `tests/resources`.
+4. Phase 2, the reward — `DiffuseBraggRewardFunction`, 15 tests covering
+   recovery (~1e-14 Bragg, ~1e-9 diffuse), discrimination (monotonic from a
+   0.02 Å displacement), per-configuration gradients, and gradient sign.
+
+**Remaining:**
+
+5. Phase 4, wiring — **done**. `--target-type` dispatch, the target arguments,
+   and a generic `prepare()` hook in both trajectory scalers, which also
+   unblocks the stranded SFC reward. A full guided run completes; see Phase 4a
+   for what it showed.
+6. Phase 5, the recovery test — guidance actually driving a mismatched ensemble
+   toward the target. The milestone, and still ahead. Needs the two fixes in
+   Phase 4a first: alignment on, and a target generated from the model's own
+   topology.
+7. Phase 3 and real experimental data after that.
+
+### What is still unproven
+
+Everything above establishes that the reward is correct and differentiable at and
+near the answer. None of it shows that **guidance converges** — that a DPS step
+size exists for which the diffuse term improves an ensemble without wrecking the
+geometry the generative prior supplies. That is Phase 5's job, and the two open
+questions feeding it are the step size (a normalized loss has different gradient
+scale than the density reward, so nothing transfers) and whether a mixture beats
+either pure target.
--- a/src/sampleworks/core/forward_models/xray/lunus_sf.py
+++ b/src/sampleworks/core/forward_models/xray/lunus_sf.py
@@ -1,0 +1,430 @@
+"""Adapter between sampleworks structures and the ``lunus.sf`` engine.
+
+``lunus.sf`` computes structure factors by splatting atomic Gaussian density onto
+a unit-cell grid and FFT-ing it: differentiable with respect to atomic fractional
+coordinates and (since 2026-08-16) occupancies. It speaks
+fractional coordinates, IT92 element symbols and integer grid symmetry
+operations; sampleworks speaks Cartesian coordinates, biotite ``AtomArray``\\ s
+and gemmi crystal metadata. This module is the translation.
+
+Everything here is coordinate-independent setup, deliberately: building the
+kernels, grid and symmetry operations is expensive and depends only on the
+topology and crystal, so it is done once in :func:`build_setup` and reused for
+every configuration and every step. That is the same two-phase split
+``StructureFactorRewardFunction`` uses for ``SFcalculator``.
+
+The one piece with no upstream equivalent is :func:`space_group_operations`:
+lunus ships ``build_grid_ops_from_cctbx``, but cctbx is not in the sampleworks
+environments, so the rotations and translations are read from gemmi instead.
+
+Symmetry expansion
+------------------
+``F(hkl)`` is the transform of the whole unit cell, so an input that is only the
+asymmetric unit must be symmetry-expanded before the FFT. That is this module's
+regime, and :func:`build_setup` builds the grid operations accordingly (for P1
+they come back empty, which lunus reads as "nothing to do").
+
+It is *not* universally correct. Coordinates that already fill the cell — an MD
+box, or pre-expanded models — must not be expanded, or every atom is counted once
+per symmetry operation. Nothing here detects that case; if this module ever needs
+to serve it, the expansion has to become an explicit choice by the caller rather
+than an assumption baked into the setup.
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import TYPE_CHECKING
+
+import gemmi
+import numpy as np
+import torch
+from jaxtyping import Complex, Float, Int
+from loguru import logger
+
+from sampleworks.utils.elements import normalize_element
+
+
+if TYPE_CHECKING:
+    from biotite.structure import AtomArray
+
+
+# gemmi's default sampling rate for density grids; lunus's grid_shape_for_resolution
+# takes the same convention (spacing = d_min / (2 * rate)).
+DEFAULT_GRID_RATE = 1.5
+
+# Isotropic B applied on top of each atom's own, widening the Gaussians so the
+# grid samples them adequately, then divided back out analytically in
+# compute_fcalc. 0.0 disables it.
+DEFAULT_BLUR = 0.0
+
+
+def space_group_operations(
+    space_group: str | gemmi.SpaceGroup,
+) -> tuple[list[np.ndarray], list[np.ndarray]]:
+    """Read a space group's symmetry operations from gemmi, for ``build_grid_ops``.
+
+    The gemmi counterpart of lunus's ``build_grid_ops_from_cctbx``. Returns the
+    operations in the fractional basis, in the form
+    ``lunus.sf.symmetry_torch.build_grid_ops`` and ``adjust_grid_for_symmetry``
+    expect: integer rotation matrices and fractional translations, *including*
+    the identity (``build_grid_ops`` drops it itself, matching gemmi's
+    ``get_scaled_ops_except_id``).
+
+    Parameters
+    ----------
+    space_group
+        Hermann-Mauguin symbol (e.g. ``"P 21 21 21"``) or a ``gemmi.SpaceGroup``.
+
+    Returns
+    -------
+    rotations : list of numpy.ndarray
+        ``(3, 3)`` integer-valued rotation matrices, one per operation.
+    translations : list of numpy.ndarray
+        ``(3,)`` fractional translations, one per operation.
+
+    Raises
+    ------
+    ValueError
+        If the symbol is not one gemmi recognizes, or if a rotation matrix is not
+        integer-valued to within a small tolerance (which would mean the
+        operations are not in the fractional basis and grid symmetrization would
+        be silently wrong).
+    """
+    if isinstance(space_group, str):
+        resolved = gemmi.SpaceGroup(space_group)
+        if resolved is None:  # gemmi raises for most bad input, but be explicit
+            raise ValueError(f"gemmi does not recognize space group {space_group!r}")
+    else:
+        resolved = space_group
+
+    rotations: list[np.ndarray] = []
+    translations: list[np.ndarray] = []
+    for op in resolved.operations():
+        # float_seitz() is the 4x4 augmented matrix with the rotation in the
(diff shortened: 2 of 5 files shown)


pyproject.toml
```diff
@@ -214,6 +214,11 @@ pandas = "==2.3.1"
 gemmi = "==0.6.7"
 
 [tool.pixi.pypi-dependencies]
+# Differentiable structure factors and diffuse scattering for the diffuse
+# guidance target; the `sf` extra pulls torch and gemmi, already in every env.
+# Pinned to a release tag, not a branch -- a branch pin resolves once at lock
+# time and never moves, so new work would never arrive. Bump the tag instead.
+lunus = {tag = "v0.2.0", extras = ["sf"], git = "https://github.com/lanl/lunus.git"}
```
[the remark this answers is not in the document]
marcuscollins: @mewall This is probably worth doing. It does get a little cumbersome to add to every environment, but you can create a pixi "feature" and then add that feature to only the environments that need it (for instance, "analysis" presumably doesn't need to install lunus.

SF_REWARD_LUNUS_BACKEND.md
```diff
@@ -0,0 +1,100 @@
+# Suggestion: try `lunus.sf` behind `StructureFactorRewardFunction`
```
marcuscollins: This is useful information but should be removed from the PR. I suggest making an issue and we'll go from there.

DIFFUSE_SCATTERING_PLAN.md
```diff
@@ -0,0 +1,461 @@
+# Plan: Diffuse Scattering as a Guidance Target (lunus.sf)
```
marcuscollins: Maybe move this to a design documents folder?

.github/workflows/relock.yml
```diff
@@ -5,6 +5,12 @@ name: Relock
 # from-scratch solve must build them for linux-64 — which cannot be done on
 # macOS. Run this workflow (Actions tab -> Relock -> Run workflow, pick the
 # branch) whenever pyproject.toml changes and the lock needs refreshing.
+#
+# Bumping a git dependency's rev: start from a lock that does not already carry
```
marcuscollins: This seems to be saying that the re-lock action isn't working as expected? If so, can you please file an issue and assign to Michael.

src/sampleworks/core/forward_models/xray/lunus_sf.py
```diff
@@ -0,0 +1,430 @@
...
+environments, so the rotations and translations are read from gemmi instead.
+
+Symmetry expansion
+------------------
+``F(hkl)`` is the transform of the whole unit cell, so an input that is only the
+asymmetric unit must be symmetry-expanded before the FFT. That is this module's
+regime, and :func:`build_setup` builds the grid operations accordingly (for P1
+they come back empty, which lunus reads as "nothing to do").
+
+It is *not* universally correct. Coordinates that already fill the cell — an MD
```
marcuscollins: Do I read this correctly to mean that we should pass in the asymmetric unit, or make sure that we use P1?

src/sampleworks/core/forward_models/xray/lunus_sf.py
```diff
@@ -0,0 +1,430 @@
...
+            raise ValueError(f"gemmi does not recognize space group {space_group!r}")
+    else:
+        resolved = space_group
+
+    rotations: list[np.ndarray] = []
+    translations: list[np.ndarray] = []
+    for op in resolved.operations():
+        # float_seitz() is the 4x4 augmented matrix with the rotation in the
+        # upper-left block and the fractional translation in the last column,
+        # already divided by gemmi's DEN. Using it avoids hand-scaling op.rot.
```
marcuscollins: what is DEN? The grid density?

src/sampleworks/core/forward_models/xray/lunus_sf.py
```diff
@@ -0,0 +1,430 @@
...
+    blur
+        Extra isotropic B applied during the splat and divided back out in the FFT.
+    n_atoms
+        Atom count the kernels were built for. Coordinates passed to
+        :func:`structure_factors` must match.
+    """
+
+    grid_shape: tuple[int, int, int]
+    orth_matrix: Float[torch.Tensor, "3 3"]
+    orth_matrix_np: np.ndarray
```
marcuscollins: Can't you just pass in the tensor and internally store a copy in numpy form? That seems cleaner and safer than forcing the user to create two copies themselves.

src/sampleworks/core/forward_models/xray/lunus_sf.py
```diff
@@ -0,0 +1,430 @@
...
+    cell_volume: float
+    grid_ops: list
+    element_idx: Int[torch.Tensor, " n_atoms"]
+    atom_A: Float[torch.Tensor, "n_atoms 5"]
```
marcuscollins, line 287: I know Karson has code for constructing these tables (A, lam) which we should re-use. In fact, it feels like maybe we ought to just pass in a list of elements and then construct these tensors here, rather than forcing the user to do it. Then we can re-use the existing code that creates the tables and make sure everything is handled the same way everywhere.

src/sampleworks/core/forward_models/xray/lunus_sf.py
```diff
@@ -0,0 +1,430 @@
...
+        Unit-cell volume in Å³.
+    grid_ops
+        Integer grid operations from ``build_grid_ops``, excluding the identity.
+        Empty for P1, which lunus reads as "no symmetry expansion".
+    element_idx
+        ``(n_atoms,)`` index into the distinct-element ordering.
+    atom_A, atom_lam
+        ``(n_atoms, 5)`` per-atom Gaussian kernel coefficients.
+    elem_offsets
+        Per-element candidate voxel offsets for the splat.
```
marcuscollins: I'm not at all sure what this means, except that above you've referred to the "splat" as being a Gaussian blur. Please provide more detail here.

src/sampleworks/core/forward_models/xray/lunus_sf.py
```diff
@@ -0,0 +1,430 @@
...
+    ValueError
+        If the atom array carries non-finite B-factors, which would produce
+        meaningless kernels.
+    """
+    from lunus.sf import (
+        adjust_grid_for_symmetry,
+        build_atom_kernels_torch,
+        build_grid_ops,
+        grid_shape_for_resolution,
+        it92_coefficients,
```
marcuscollins: I guess these are the coefficients for the Gaussians? Karson has these somewhere else in the package, as well as code for computing the element->scattering factor tables. I wonder if we should standardize?

src/sampleworks/core/forward_models/xray/lunus_sf.py
```diff
@@ -0,0 +1,430 @@
...
+    # Fetch coefficients for exactly the elements present, rather than using
+    # lunus's IT92_COEFFS constant: that is a convenience table covering a
+    # DEFAULT set, and anything outside it (Se in a selenomethionine structure,
+    # metals, halides) raises KeyError deep inside the kernel builder.
+    # it92_coefficients() reads any element gemmi knows, at run time.
+    try:
+        coefficients = it92_coefficients(distinct_elements)
+    except KeyError as e:
+        raise ValueError(
+            f"No IT92 scattering coefficients for element {e} in "
```
marcuscollins: Here `e` will not be the element, it will be the error object, so this error message needs to be fixed.

src/sampleworks/core/forward_models/xray/lunus_sf.py
```diff
@@ -0,0 +1,430 @@
...
+            f"{distinct_elements}. Check the structure's element annotations; "
+            "strip or rename the offending atoms if the element is spurious."
+        ) from e
+
+    a, b, c = unit_cell.a, unit_cell.b, unit_cell.c
+    orth_np = build_orth_matrix(a, b, c, unit_cell.alpha, unit_cell.beta, unit_cell.gamma)
+
+    rotations, translations = space_group_operations(space_group)
+    raw_shape = grid_shape_for_resolution(a, b, c, resolution, rate)
+    grid_shape = adjust_grid_for_symmetry(raw_shape, rotations, translations)
```
marcuscollins, line 428: Can you add a comment explaining what these two lines do? I probably once knew, but I don't remember now why you'd adjust the grid for symmetry.

marcuscollins: I guess the grid itself needs to be invariant under the group?

src/sampleworks/core/forward_models/xray/lunus_sf.py
```diff
@@ -0,0 +1,430 @@
...
+    orth_t = torch.as_tensor(orth_np, dtype=dtype, device=device)
+    logger.info(
+        f"lunus setup: {len(elements)} atoms, grid {tuple(grid_shape)}, "
+        f"{len(rotations)} symmetry operations ({len(grid_ops)} beyond identity), "
+        f"elements {distinct_elements}"
+    )
+
+    return LunusSetup(
+        grid_shape=tuple(grid_shape),
+        orth_matrix=orth_t,
```
marcuscollins: The one thing I'll say here is that you can add a `__post_init__` method to the LunusSetup class that could handle things like storing the tensor version of orth_np, and compute the cell volume. Then you could just pass orth_np which would be a little cleaner.

src/sampleworks/core/forward_models/xray/lunus_sf.py
```diff
@@ -0,0 +1,430 @@
...
+        atom_lam=atom_lam,
+        elem_offsets=elem_offsets,
+        atom_radius_ang=atom_radius_ang,
+        taper_width=taper_width,
+        blur=blur,
+        n_atoms=len(elements),
+    )
+
+
+def cartesian_to_fractional(
```
marcuscollins: We have other code that does this conversion (from qFit originally, I think). Perhaps check to see whether that will do the job, or could be modified to do the job? I'm trying to avoid too much replication, especially where divergences might lead to confusion down the road when comparing methods.

src/sampleworks/core/forward_models/xray/lunus_sf.py
```diff
@@ -0,0 +1,430 @@
...
+        ``[n_atoms]`` shared across configurations, or ``[n_configs, n_atoms]``
+        per configuration.
+    hkl
+        Miller indices ``[n_refl, 3]`` to extract.
+    solvent
+        A ``lunus.sf.SolventModel``, or None for no bulk solvent. Applied per
+        configuration — each member gets a mask from its own density, which is
+        the only choice that contributes to a diffuse observable.
+    use_checkpoint
+        Recompute each configuration's splat during backward instead of retaining
```
marcuscollins, line 513: What is the splatting for, exactly? You're adding a little blur, is that right?

src/sampleworks/core/forward_models/xray/lunus_sf.py
```diff
@@ -0,0 +1,430 @@
...
+        Complex ``F(hkl)``, shape ``[n_configs, n_refl]``. Pass to
+        ``lunus.sf.mean_and_diffuse`` for the ensemble observables.
+
+    Raises
+    ------
+    ValueError
+        If ``coords`` is not 3-dimensional, or its atom count disagrees with the
+        setup's.
+    """
+    from lunus.sf import structure_factors_batch
```
marcuscollins: Why import inline instead of at the top of the file?

src/sampleworks/synthetic/generate_synthetic_sf_lunus.py
```diff
@@ -0,0 +1,847 @@
...
+form cannot express the second (``<|F|²> − |<F>|²``), which is why diffuse work
+needs this shape. ``--write-diffuse`` emits that second moment as a separate MTZ
+from the same forward pass, and ``--altlocs-as-models`` turns a deposited
+multi-conformer structure into the ensemble it already is — which is how to get
+a nonzero diffuse target out of a file like ``1vme_final.cif``.
+
+**Grid method, not direct summation.** lunus splats density onto a unit-cell
+grid, symmetry-expands, and FFTs, where SFcalculator sums over atoms in
+reciprocal space. Agreement is close but not exact — lunus measures correlation
+0.999989 and R ≈ 0.0077 against gemmi, with its smooth density taper the main
```
marcuscollins: I don't follow--what is the smooth density taper? I guess something to avoid hard edges in the density and resulting HF components in the FFT? But I thought the density was already smooth, by way of being a sum of gaussian balls?

src/sampleworks/synthetic/generate_synthetic_sf_lunus.py
```diff
@@ -0,0 +1,847 @@
...
+"""
+
+import argparse
+import sys
+import traceback
+from pathlib import Path
+
+import gemmi
+import numpy as np
+import reciprocalspaceship as rs
```
marcuscollins, line 58: Are there things in rs that aren't in gemmi, and vice versa? I'd like to try to avoid additional dependencies.

marcuscollins: Also, if we need this as an explicit dependency, a couple things:
1. make sure we have tests that run over this code, so we'll catch if we lose the dependency in pyproject.toml
2. add an explicit dependency on rss in pyproject.toml, since right now we're getting it only through sf calculator.

src/sampleworks/synthetic/generate_synthetic_sf_lunus.py
```diff
@@ -0,0 +1,847 @@
...
+    altlocs_as_models: bool = False,
+) -> tuple[AtomArray, np.ndarray]:
+    """Load a structure as a topology plus a stack of configuration coordinates.
+
+    Single-model files go through the shared
+    :func:`load_structure_for_synthetic_reward`, so selection, stripping and the
+    occupancy modes behave exactly as in the SFcalculator script.
+
+    Multi-model files are loaded directly and **support none of those options**:
+    applying a selection consistently across models needs index plumbing that
```
marcuscollins: This doesn't make sense. AtomArrayStacks absolutely support selections, stripping of hydrogens, etc... we do this all the time. Support for occupancies isn't great, that's true.

src/sampleworks/synthetic/generate_synthetic_sf_lunus.py
```diff
@@ -0,0 +1,847 @@
...
+    loaded = load_any(structure_path, altloc="all", extra_fields=["occupancy", "b_factor"])
+
+    if isinstance(loaded, AtomArrayStack) and loaded.stack_depth() > 1:
+        unsupported = [
+            name
+            for name, active in (
+                ("selection", row.selection),
+                ("--remove-hydrogens", strip_hydrogens),
+                ("--remove-waters", strip_waters),
+                ("--remove-ligands", strip_ligands),
```
marcuscollins: This needs to be fixed. We can handle these things with AtomArrayStack.

src/sampleworks/synthetic/generate_synthetic_sf_lunus.py
```diff
@@ -0,0 +1,847 @@
...
+            raise ValueError(
+                f"{structure_path.name} holds {loaded.stack_depth()} models, and "
+                f"{', '.join(unsupported)} is not supported for multi-model input. "
+                "Preprocess the ensemble, or use a single-model file."
+            )
+        logger.info(f"Loaded {loaded.stack_depth()} models from {structure_path.name}")
+        return loaded[0], np.asarray(loaded.coord, dtype=np.float64)
+
+    if altlocs_as_models:
+        return _expand_altlocs(loaded, structure_path, row.selection)
```
marcuscollins: See `sampleworks.utils.atom_array_utils.map_altlocs_to_stack` and use it here. We need to avoid creating too many versions of the same functions.

src/sampleworks/synthetic/generate_synthetic_sf_lunus.py
```diff
@@ -0,0 +1,847 @@
...
+        strip_waters=strip_waters,
+        strip_ligands=strip_ligands,
+        selection=row.selection,
+    )
+    if atom_array is None:
+        raise ValueError(f"Failed to load {structure_path}")
+    return atom_array, np.asarray(atom_array.coord, dtype=np.float64)[None, ...]
+
+
+def _expand_altlocs(
```
marcuscollins: Won't review, as I think you should be using the existing `map_altlocs_to_stack` method.

src/sampleworks/synthetic/generate_synthetic_sf_lunus.py
```diff
@@ -0,0 +1,847 @@
...
+    Returns
+    -------
+    hkl : numpy.ndarray
+        ``(n_refl, 3)`` Miller indices.
+    mean_f : numpy.ndarray
+        ``(n_refl,)`` complex ensemble mean ``<F>``.
+    diffuse : numpy.ndarray
+        ``(n_refl,)`` real ``<|F|²> − |<F>|²``. Zero for a single configuration.
+    """
+    from lunus.sf import mean_and_diffuse
```
marcuscollins: Avoid inline imports. Move them to the top unless you're "hiding" them in situations where the function may require packages that aren't usually installed, or if it is a big import that you will only use some of the time.

src/sampleworks/synthetic/generate_synthetic_sf_lunus.py
```diff
@@ -0,0 +1,847 @@
...
+    setup = lunus_sf.build_setup(atom_array, unit_cell, space_group, resolution, device=device)
+
+    hkl_np = generate_asu_hkl(unit_cell, space_group, resolution)
+    hkl = torch.as_tensor(hkl_np, dtype=torch.long, device=device)
+    coords_t = torch.as_tensor(coords, dtype=torch.float32, device=device)
+    occupancies = torch.as_tensor(np.asarray(atom_array.occupancy, dtype=np.float32), device=device)
+
+    solvent = None
+    if solvent_cutoff is not None:
+        from lunus.sf import SolventModel
```
marcuscollins, line 378: This is an example where an inline import might make sense. Presumably you don't always need it, and it is controlled by user input.

src/sampleworks/synthetic/generate_synthetic_sf_lunus.py
```diff
@@ -0,0 +1,847 @@
...
+        # Populated only when check_occupancy is on, SolventModel's default.
+        if solvent.last_occupancy is None:
+            logger.info("Solvent mask applied; occupancy not measured (check_occupancy off)")
+        else:
+            n_voxels = int(np.prod(setup.grid_shape))
+            shell_fraction = solvent.last_shell_voxels / n_voxels
+            logger.info(
+                f"Solvent mask: occupancy {solvent.last_occupancy:.3f}, "
+                f"taper shell {solvent.last_shell_voxels} voxels "
+                f"({shell_fraction:.1%} of the {n_voxels} in the grid) "
```
marcuscollins: As a comment in the code, or part of this message, can you provide some nominal ranges for each of the reported parameters to help people identify whether they need to investigate further?

src/sampleworks/synthetic/generate_synthetic_sf_lunus.py
```diff
@@ -0,0 +1,847 @@
...
+                f"(cutoff {solvent.cutoff}, taper {solvent.taper_width} e/A^3)"
+            )
+
+    return hkl_np, mean_f.cpu().numpy(), diffuse.cpu().numpy()
+
+
+def dataset_from_intensities(
```
marcuscollins: The organization of this code isn't intuitive to me, and this is an example. It seems like the inputs here are mostly the same at the method above (space group, Miller indices, space group, etc...) plus the intensities that method generated. Why not just return rs.DataSet objects from `compute_ensemble_amplitudes`, and then write the mtz files from those?

src/sampleworks/synthetic/generate_synthetic_sf_lunus.py
```diff
@@ -0,0 +1,847 @@
...
+
+    if output_path is not None:
+        output_path.parent.mkdir(parents=True, exist_ok=True)
+        dataset.write_mtz(str(output_path))
+        logger.info(f"Saved diffuse intensities to {output_path}")
+
+    return dataset
+
+
+def dataset_from_amplitudes(
```
marcuscollins: If you do decide to keep `dataset_from_intensities`, I might do a couple things for organization:
1. separate the creation of the dataset from writing it. 
2. combine it with this method. You can make both `structure_factors` and `intensities` optional arguments, raise an error if both are None or both are not None, and then switch behavior internally based on which one is present.

The principle I have in mind is "don't repeat yourself". Sometimes it actually helps to repeat yourself because it makes the code easier to read and digest, but here I think there's so much overlap that it would be easier to have a "Swiss Army knife" method to create the rs.DataSet

src/sampleworks/synthetic/generate_synthetic_sf_lunus.py
```diff
@@ -0,0 +1,847 @@
...
+            solvent_taper_width=solvent_taper_width,
+        )
+    except Exception as e:
+        logger.error(
+            f"Failed to compute structure factors for {row.filename} "
+            f"({type(e).__name__}): {e}\n{''.join(traceback.format_tb(e.__traceback__))}"
+        )
+        return
+
+    if write_diffuse:
```
marcuscollins, line 713: Save yourself some compute and check this before calling `compute_ensemble_amplitudes`, then you can skip that call and move on to the next row.

src/sampleworks/synthetic/generate_synthetic_sf_lunus.py
```diff
@@ -0,0 +1,840 @@
...
+            output_dir=output_dir,
+            resolution=resolution,
+            occupancy_mode=occupancy_mode,
+            test_fraction=test_fraction,
+            seed=seed,
+            device=device,
+            **row_kwargs,
+        )
+        for row in rows
+    )
```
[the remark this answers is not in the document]
marcuscollins: This is a good thing to do: write out some report file that indicates which runs failed and which succeeded. It's easier to look through afterwards than trying to parse the logs. Once you know which runs to look for you can go into the logs to see what actually happened.

tests/synthetic/test_generate_synthetic_sf_lunus.py
```diff
@@ -0,0 +1,333 @@
...
+
+
+@pytest.fixture(scope="module")
+def device() -> torch.device:
+    """CPU: these tests check numerics, not throughput, and must run in CI."""
+    return torch.device("cpu")
+
+
+@pytest.fixture(scope="module")
+def structure_1vme(resources_dir: Path):
```
marcuscollins: Surely we have this fixture somewhere else? I would try to re-use it from elsewhere rather than recreating it here. Check `conftest.py` files (there are a couple)

tests/synthetic/test_generate_synthetic_sf_lunus.py
```diff
@@ -0,0 +1,333 @@
...
+    """Chain A of 1VME with hydrogens and waters stripped, as a topology + coordinates.
+
+    Deliberately the same selection the SFcalculator reward fixtures use
+    (``tests/rewards/conftest.py``), so the two engines are compared on identical
+    inputs.
+    """
+    from sampleworks.synthetic.generate_synthetic_sf import BatchRowForMTZ
```
marcuscollins: put these imports at the top of the file.

tests/synthetic/test_generate_synthetic_sf_lunus.py
```diff
@@ -0,0 +1,333 @@
...
+    from sampleworks.synthetic.generate_synthetic_sf_lunus import load_configurations
+
+    source_dir = resources_dir / "1vme"
+    if not (source_dir / SOURCE_CIF).exists():
+        pytest.skip(f"Source structure not found at {source_dir / SOURCE_CIF}")
+
+    row = BatchRowForMTZ(filename=SOURCE_CIF, selection="chain A")
```
marcuscollins: this seems very roundabout, you just need the AtomArray, right? You can get the coordinates from it later easily enough, and then do any manipulations you need on that.

tests/synthetic/test_generate_synthetic_sf_lunus.py
```diff
@@ -0,0 +1,333 @@
...
+        strip_hydrogens=True,
+        strip_waters=True,
+    )
+    return atom_array, coords
+
+
+@pytest.fixture(scope="module")
+def crystal_1vme(resources_dir: Path):
+    """Unit cell and space group read from the deposited file."""
+    import gemmi
```
marcuscollins: Again, since these fixtures will be created no matter what as the file is read, there's no reason to have inline imports.

tests/synthetic/test_generate_synthetic_sf_lunus.py
```diff
@@ -0,0 +1,333 @@
...
+        # absolute residual on the largest reflection against a mean intensity.
+        # Measured on 1VME chain A at 1.8 A: 3.1e-8 to 4.4e-8 across runs, i.e.
+        # float32 epsilon. It varies run to run because the splat's reduction
+        # order does; the bound allows for that rather than pinning one value.
+        intensity = np.abs(mean_f).astype(np.float64) ** 2
+        rms_intensity = float(np.sqrt(np.mean(intensity**2)))
+        rms_diffuse = float(np.sqrt(np.mean(diffuse.astype(np.float64) ** 2)))
+        assert rms_intensity > 0, "degenerate <F>; the calculation produced nothing"
+
+        ratio = rms_diffuse / rms_intensity
```
marcuscollins, line 170: why a ratio and not a difference?

marcuscollins (changes requested):

src/sampleworks/core/forward_models/xray/lunus_sf.py
mewall: Yes, those are the Gaussian coefficients. 

On standardizing, this can be done by generalizing lunus.sf to handle more Gaussians. What would you think about making that an issue we come back to? The current tables follow what SFCalculator/gemmi does. 

Alternatively, maybe Karson could add support for IT92 params to make contact with these other methods?

tests/synthetic/test_generate_synthetic_sf_lunus.py
mewall: Here's what Claude Code says. I'm happy to go ahead with whatever you think is best.

---

Checked all seven conftest files. None returns what this module needs: a
topology plus a coordinate array carrying a configuration axis.

The closest is `tests/rewards/conftest.py::structure_1vme_sf`, and it isn't
close — a single `AtomArray` with no configuration axis, loaded from a generated
cif that costs a full `_process_single_row` run (SFcalculator plus an MTZ write)
to produce, with occupancies deliberately set to `custom` 0.5/0.5 where these
tests want `default`. It also lives under `tests/rewards/`, so it isn't visible
from `tests/synthetic/`. `tests/conftest.py::structure_1vme` is a parsed
atomworks dict, and `load_configurations` takes a path rather than a parsed
dict, so even the file loading doesn't carry over.

The one that looks most like a duplicate is `ensemble` in
`tests/rewards/test_diffuse_bragg_reward.py` (#381). Same loader, different
arguments, and so a different structure:

    this module : 3357 atoms, coords (1, 3357, 3)   chain A, H + waters stripped
    ensemble    : 6819 atoms, coords (2, 6819, 3)   whole structure, altlocs as models

Different selection, stripping and altloc handling, hence a different atom count
and a different number of configurations. Sharing them would mean a
parameterised factory whose arguments are the entire body, not a shared fixture.

That said, I think what prompted this is real: the fixture was named
`structure_1vme`, which shadows the session fixture of that name in
`tests/conftest.py` returning a different type — so it reads as a recreation of
something that already exists. Renamed to `configurations_1vme`, with a
docstring saying what it is distinct from. Same treatment for the local
`device`, which was shadowing the GPU-preferring session fixture while
deliberately forcing CPU.

Happy to promote a shared factory into `tests/conftest.py` once #381 lands and
there are two real callers. One caveat if we do: it would need a lazy
`load_configurations` import, otherwise every environment that imports
`tests/conftest.py` requires lunus — including the analysis ones we are
scoping it out of.

src/sampleworks/synthetic/generate_synthetic_sf_lunus.py
mewall: See https://github.com/prism-science/sampleworks/pull/380#discussion_r4012095704

src/sampleworks/synthetic/generate_synthetic_sf_lunus.py
mewall: Again, Claude Code's response. I don't really like the way these responses read but haven't figured out how to make them sound better and want to see if I can get through these tonight. Will go with your recommendation:

---

Not an additional dependency: it was already in the tree before this PR, imported
by `core/rewards/structure_factor.py`, `synthetic/synthetic_utils.py` and
`synthetic/generate_synthetic_sf.py`, and it arrived transitively through
sfcalculator-torch. Your follow-up was the right catch, though — it was
undeclared, and I have now added it to `project.dependencies` explicitly.

On what each is for, they are not really substitutes here. gemmi supplies the
crystallography: `UnitCell`, `SpaceGroup`, and `make_miller_array` for the ASU
reflection list. reciprocalspaceship supplies MTZ I/O with typed columns —
`rs.DataSet`, `rs.HKLIndexDtype`, `rs.IntensityDtype` — so the amplitudes and
intensities land in the file with the MTZ column types a downstream reader
expects. gemmi can write MTZ as well, so this is not something only rs can do;
it is that rs is what every other MTZ writer in the repo already uses, and
writing this one differently would be the inconsistency worth avoiding.

src/sampleworks/core/forward_models/xray/lunus_sf.py
mewall: It's designed to handle the case where one passes the asymmetric unit  (crystal structure or structure prediction model) or a P1 model system (MD simulation). It will be good to discuss use cases and the results it's capable of producing, it's pretty flexible.

src/sampleworks/core/forward_models/xray/lunus_sf.py
mewall: The docstring has been clarified with the following.

Splatting refers to adding the density of an atom to the total grid density.

Blurring is an anti-aliasing trick (as I recall due to Lynn Ten Eyck in an early
DFT paper). If B factors are low, the atom densities are sharp, and a finer grid
is needed to avoid errors in structure factor calculations, even when calculating
to lower resolutions. If the B factors are increased uniformly one can use a
coarser grid; then, after calculating, the structure factors can be corrected by
removing the effect of the uniform B factor.

In commonly used codes the footprint of the atom densities is decreased by
truncating at long distances. Gemmi does it using a density threshold, for
example. The truncation leads to a sharp, discontinuous edge in each atomic
density contribution. The taper smooths the edge to bring it to zero continuously
over some distance (taper_width).

src/sampleworks/core/forward_models/xray/lunus_sf.py
mewall: Yes, the grid has to be invariant under the group. 

Comment added at the call site.

tests/synthetic/test_generate_synthetic_sf_lunus.py
mewall: Claude wrote the test and came up with the following explanation. I'm open to changing if you have suggestions...

Because the quantity being tested is a cancellation error, and its size scales
with the numbers being cancelled. The diffuse term is `<|F|^2> - |<F>|^2`,
identically zero for identical configurations, so what the test measures is how
close to zero float32 gets. That residue is proportional to the intensity scale,
which depends on atom count, occupancies and resolution — so an absolute
threshold would have to be re-derived for every structure the test is ever
pointed at, and would silently become meaningless if the fixture changed.

RMS-to-RMS divides that scale out and leaves a dimensionless number that is
about float32 rather than about 1VME: measured 3.1e-8 to 4.4e-8 across runs,
bounded at 1e-6. On aarch64 it comes out at 4.3e-9.

mewall:
@marcuscollins 

Regrets for the lazy use of AI in these responses. I'm trying to stay ahead of the movement of main, so we can work towards a merge. Whereas other responses are accompanied by code changes, these are described as needing some direction before moving forward. Let me know if anything is unclear (I'm not sure I completely understand it all after a cursory read) or if you want to follow up on anything. 


---
Four left from your review that I did not want to decide
unilaterally. Taking the easy one first.

**Inline import** — https://github.com/prism-science/sampleworks/pull/380#discussion_r3929565294
Already fixed: `structure_factors_batch` is imported at the top of the module
now, along with the other lunus.sf names, under the rule you gave elsewhere in
the review. No inline lunus imports remain in the adapter; the only one left in
the generator is `SolventModel`, which is gated on `--solvent-cutoff` and which
you endorsed keeping inline.

**Fractional conversion** — https://github.com/prism-science/sampleworks/pull/380#discussion_r3929542120
The conversion itself cannot be shared: `qfit/unitcell.py` is NumPy throughout,
and `cartesian_to_fractional` has to stay differentiable torch, since the
gradient runs from the reward back through it to the coordinates. So
`calc_orth_to_frac` is not a drop-in.

What *is* shareable is the matrix construction, which is coordinate-independent
and built once at setup. But the divergence you are worried about does not seem
to exist yet — I compared the two constructions on 1VME
(55.242, 95.831, 90.13, 90, 95.43, 90):

        lunus orth        qFit orth
        [55.242  0  -8.52896]   [55.242  0  -8.52896]
        [ 0  95.831  0      ]   [ 0  95.831  0      ]
        [ 0   0  89.72555   ]   [ 0   0  89.72555   ]

Bit-identical, max abs difference 0.0. Same convention, so a comparison across
methods is not currently at risk. Happy to route the reciprocal-space path
through qFit's matrix anyway if you would rather have one source than two that
agree — say the word and I will, with a test asserting they stay equal.

**dataset_from_intensities / dataset_from_amplitudes** —
https://github.com/prism-science/sampleworks/pull/380#discussion_r3935640081
https://github.com/prism-science/sampleworks/pull/380#discussion_r3935684262
I would like to take your second suggestion and not the first, if that is
acceptable. Splitting dataset creation from writing, and merging the two into
one constructor with `structure_factors` and `intensities` as mutually exclusive
optional arguments, is clearly right — the overlap is most of both bodies.

Returning `rs.DataSet` objects from `compute_ensemble_amplitudes` I would rather
not do. That function is also the reward's forward path, where the result is a
torch tensor that has to stay on the device and stay differentiable; wrapping it
in a pandas-backed DataSet would mean converting to numpy and back, and the
guidance loop calls it every step. Keeping it returning arrays, with the
DataSet construction as a separate step the generator calls and the reward does
not, keeps that boundary where it is.

**Test fixture** — https://github.com/prism-science/sampleworks/pull/380#discussion_r3935865514
You are right for these tests. `load_configurations` returns
`(topology, coords)` with a leading configuration axis, but this fixture selects
chain A of a single-model file, so the shape is `(1, 3357, 3)` and
`atom_array.coord` is exactly `coords[0]` — I checked. Every test in the module
then builds what it needs from `coords[0]` anyway.

The reason it is a tuple is that it calls the generator's own loader rather than
a parallel path, so the fixture exercises the code under test. If you would
rather it took just the AtomArray I will change it, but then the configuration
axis stops being pinned anywhere in this module, and that axis is the whole
point of the diffuse work — so I would want to add a small test asserting the
loader's return shape to replace what the fixture currently demonstrates by
using it.

.github/workflows/relock.yml
mewall: Having looked into it I think Claude Code was incorrect about this — Relock works. The comment was overstating a narrow failure, and was also wrong about its cause, header rewritten.

SF_REWARD_LUNUS_BACKEND.md
mewall: diffuse pixi feature has been added. I've removed the doc, let's come back to it when we have time to discuss the path forward.

tests/synthetic/test_generate_synthetic_sf_lunus.py
mewall: I think this has been addressed.

tests/synthetic/test_generate_synthetic_sf_lunus.py
mewall: I think there was one inline import that was kept with justification by Claude Code, the rest should have been moved to the top.

DIFFUSE_SCATTERING_PLAN.md
mewall: This was removed. I can put it in a design docs folder when we work out what we want that to look like.

src/sampleworks/core/forward_models/xray/lunus_sf.py
mewall: Revised as follows:  

        # float_seitz() is the 4x4 augmented matrix with the rotation in the
        # upper-left block and the fractional translation in the last column.
        # gemmi holds both as integers scaled by Op.DEN, a fixed denominator
        # chosen so that every crystallographic rational translation (1/2, 1/3,
        # 1/4, 1/6, ...) is represented exactly; float_seitz() has already
        # divided it out, so op.rot and op.tran need no hand-scaling here.

src/sampleworks/core/forward_models/xray/lunus_sf.py
mewall: Karson's code uses a 6-Gaussian calculation and doesn't make use of (A, lam) like lunus.sf does. On the other hand lunus.sf, gemmi, and SFCalculator use a 5-Gaussian. 

The code re-use/consistency was addressed by adding a new it92_coefficients() in utils/elements.py (reads IT92 from gemmi, returns (a1..a4, b1..b4, c)) instead of using the lunus.sf reader. Also added IT92 as a source in Karson's function. setup_scattering_params gains it92_mode: bool = False, packing the same coefficients into the existing (n_indices, 6, 2) layout as a = [a1..a4, 0, c], b = [b1..b4, 0, 0]. It's not instrumented elsewhere to keep the PR focused but later we can use it to add an it92 option in Karson's code to increase consistency among the density, structure factor, diffuse methods. 

3 new tests added in TestSetupScatteringParams.

src/sampleworks/core/forward_models/xray/lunus_sf.py
mewall: I went ahead and added a helper to get the IT92 params in the same way as Karson's code gets the Gaussian coefficients he's using, as mentioned above. That should help with adding support for IT92 in Karson's code path, in a future revision. In this way we can standardize around IT92 first, and then maybe later I can add the qfit method to lunus.sf (I don't think gemmi supports it yet, not sure about cctbx).

> Yes, those are the Gaussian coefficients.
> 
> On standardizing, this can be done by generalizing lunus.sf to handle more Gaussians. What would you think about making that an issue we come back to? The current tables follow what SFCalculator/gemmi does.
> 
> Alternatively, maybe Karson could add support for IT92 params to make contact with these other methods?

src/sampleworks/core/forward_models/xray/lunus_sf.py
mewall: fixed in a315e25 —  the f-string now inserts e.args[0] instead of e, the exception object itself.

src/sampleworks/core/forward_models/xray/lunus_sf.py
mewall: Done — `orth_np` is an `InitVar` now, and `__post_init__` derives both the tensor
and the volume from it.

This also improved accuracy.  The volume was being taken from the stored
tensor, so at `dtype=float32` it inherited float32 truncation (e.g. 72569.3167 for a
cell whose volume is 72569.3159). Deriving it from the float64 numpy input gives
the right answer at either dtype.

src/sampleworks/core/forward_models/xray/lunus_sf.py
mewall: Yes, qFit's matrix is now used — it's exactly inv(orth), difference 0.0 — the flagged function is gone (the conversion is a local matmul), and a test pins all three matrix sources against each other.

src/sampleworks/core/forward_models/xray/lunus_sf.py
mewall: See response https://github.com/prism-science/sampleworks/pull/380#discussion_r3975524185 for info about splatting and blur.

src/sampleworks/core/forward_models/xray/lunus_sf.py
mewall: Fixed in 4ee2dfd — it's at module scope now, paired with scoping lunus to the diffuse pixi feature

src/sampleworks/synthetic/generate_synthetic_sf_lunus.py
mewall: See https://github.com/prism-science/sampleworks/pull/380#discussion_r3975524185 for info about the taper

src/sampleworks/synthetic/generate_synthetic_sf_lunus.py
mewall: > Are there things in rs that aren't in gemmi, and vice versa? I'd like to try to avoid additional dependencies.

The forward model now uses neither rs nor SFcalculator, and rs enters only at the MTZ-writing. The blocker is add_rfree, which Claude Code can't find a gemmi equivalent for. We could maybe roll our own add_rfree and lose the dependency here?

src/sampleworks/synthetic/generate_synthetic_sf_lunus.py
mewall: you're right — fixed in 7450564; selection and stripping now apply to stacks

src/sampleworks/synthetic/generate_synthetic_sf_lunus.py
mewall: Yes, see fix in https://github.com/prism-science/sampleworks/pull/380#discussion_r4011992251

src/sampleworks/synthetic/generate_synthetic_sf_lunus.py
mewall: I inlined _expand_altlocs as it's only used here. It calls map_altlocs_to_stack and does some other things as well. Let me know if you'd like it tidied away again, or want to handle this in some other way.

src/sampleworks/synthetic/generate_synthetic_sf_lunus.py
mewall: mean_and_diffuse moved to module scope in 4ee2dfd, the atom_array_utils inline import went away with the altloc inlining in 9a19650

src/sampleworks/synthetic/generate_synthetic_sf_lunus.py
mewall: Yes, SolventModel is not always needed. Added a comment stating why the inline import is used here.

src/sampleworks/synthetic/generate_synthetic_sf_lunus.py
mewall: Some comments added in the code about things to look out for, and some examples of values for a test case mentioned in the logger.

src/sampleworks/synthetic/generate_synthetic_sf_lunus.py
mewall: I had Claude analyze the code to try to align better with conventions around naming and documenting flow. Following the `runs/loader.py` example the module docstring now names its three stages
(`load_configurations` → `compute_ensemble_amplitudes` → `dataset_from_*`) and why they're
separate. : `_process_single_row` wraps each in its own `try`/`except` so `batch_report.json` can
say which stage a row died in (a failed diffuse write still lets the amplitudes through), and
keeping the forward pass free of `reciprocalspaceship` means a caller wanting only arrays never
touches the MTZ layer. We can revisit this if you have a strong preference for refactoring.

Two things the analysis turned up, now changed: `process_batch` takes its parameters explicitly
with a full numpydoc block instead of `**row_kwargs`, and `batch_report.json` uses
`run_grid_search.py`'s `{"runs": [...], "summary": {...}}` shape rather than its own flat keys.

mewall: Update: after the two commits below the pipeline is now four stages, not three —
`load_configurations` → `compute_ensemble_amplitudes` → `dataset_from_bragg_amplitudes` /
`dataset_from_diffuse_intensities` → `save_mtz`.

The original writers were given confusing names by Claude, they have been renamed using Bragg vs diffuse (`03ebcf1`)

A deeper search found the `save_density` in `generate_synthetic_density.py` precedent you've mentioned for returning the rs.DataSet objects, so that's been done as well (`a50d2c8`)

src/sampleworks/synthetic/generate_synthetic_sf_lunus.py
mewall: See https://github.com/prism-science/sampleworks/pull/380#discussion_r4022457375

src/sampleworks/synthetic/generate_synthetic_sf_lunus.py
mewall: Done in a315e25 — the check runs before the structure-factor call now. Note the consequence: a single-model row with --write-diffuse is skipped entirely with failure_stage: "validate", rather than still writing its Bragg MTZ.

tests/synthetic/test_generate_synthetic_sf_lunus.py
mewall: I looked at this again, and found that this can be addressed by switching the tests over to 6B8X. This is done.

One change to that fixture: it needed `"b_factor"`
alongside `"occupancy"`, since the scattering kernels are built per atom from B. This was added.

The smallest B fac in 6B8X is 2.0, smaller than 12.1 in 1VME, so this is also a stricter test, helping to find issues with the blur.  `build_setup` now picks the blur itself; details in `5a94e33`. I filed an issue to change this in lunus.sf as well.

Worth considering the same switch to 6B8X for the SFcalculator fixtures in `tests/rewards/conftest.py`,
which are all 1VME — 6B8X is a harder test there too, and its P 31 2 1 ASU is less than half
the size, so SFcalculator peaks at 4.94 GB rather than 9.92 GiB at 2.2 Å. 

> Here's what Claude Code says. I'm happy to go ahead with whatever you think is best.
> 
> Checked all seven conftest files. None returns what this module needs: a topology plus a coordinate array carrying a configuration axis.
> 
> The closest is `tests/rewards/conftest.py::structure_1vme_sf`, and it isn't close — a single `AtomArray` with no configuration axis, loaded from a generated cif that costs a full `_process_single_row` run (SFcalculator plus an MTZ write) to produce, with occupancies deliberately set to `custom` 0.5/0.5 where these tests want `default`. It also lives under `tests/rewards/`, so it isn't visible from `tests/synthetic/`. `tests/conftest.py::structure_1vme` is a parsed atomworks dict, and `load_configurations` takes a path rather than a parsed dict, so even the file loading doesn't carry over.
> 
> The one that looks most like a duplicate is `ensemble` in `tests/rewards/test_diffuse_bragg_reward.py` (#381). Same loader, different arguments, and so a different structure:
> 
> ```
> this module : 3357 atoms, coords (1, 3357, 3)   chain A, H + waters stripped
> ensemble    : 6819 atoms, coords (2, 6819, 3)   whole structure, altlocs as models
> ```
> 
> Different selection, stripping and altloc handling, hence a different atom count and a different number of configurations. Sharing them would mean a parameterised factory whose arguments are the entire body, not a shared fixture.
> 
> That said, I think what prompted this is real: the fixture was named `structure_1vme`, which shadows the session fixture of that name in `tests/conftest.py` returning a different type — so it reads as a recreation of something that already exists. Renamed to `configurations_1vme`, with a docstring saying what it is distinct from. Same treatment for the local `device`, which was shadowing the GPU-preferring session fixture while deliberately forcing CPU.
> 
> Happy to promote a shared factory into `tests/conftest.py` once #381 lands and there are two real callers. One caveat if we do: it would need a lazy `load_configurations` import, otherwise every environment that imports `tests/conftest.py` requires lunus — including the analysis ones we are scoping it out of.

tests/synthetic/test_generate_synthetic_sf_lunus.py
mewall: Right, especially after 58c451c. 2857510 uses the AtomArray and each test derives the coordinates it needs.

tests/synthetic/test_generate_synthetic_sf_lunus.py
mewall: Looked at this again. The gemmi import moved to module scope in 58c451c along with the two loader imports, atomarray_to_gemmi followed in 8f427bc. The two SFcalculator imports stay because the dependency is optional — now with a comment saying so.

tests/synthetic/test_generate_synthetic_sf_lunus.py
mewall: I had another look. The name "ratio" was misleading. I changed it to "normalized_residual". Dividing by the RMS intensity puts the RMS diffuse on a scale that doesn't vary as much depending on space group, resolution, etc. So the test should work for a wider variety of systems.

SF_REWARD_LUNUS_BACKEND.md
mewall: Permalink: https://github.com/prism-science/sampleworks/blob/2ac07e20c23dcd42eab6fa43dfcd86c9370dbed8/pyproject.toml#L55

src/sampleworks/core/forward_models/xray/lunus_sf.py
mewall: Permalink: https://github.com/prism-science/sampleworks/blob/2ac07e20c23dcd42eab6fa43dfcd86c9370dbed8/src/sampleworks/core/forward_models/xray/lunus_sf.py#L20

src/sampleworks/core/forward_models/xray/lunus_sf.py
mewall: Permalink: https://github.com/prism-science/sampleworks/blob/2ac07e20c23dcd42eab6fa43dfcd86c9370dbed8/src/sampleworks/core/forward_models/xray/lunus_sf.py#L203

src/sampleworks/core/forward_models/xray/lunus_sf.py
mewall: Permalink: https://github.com/prism-science/sampleworks/blob/2ac07e20c23dcd42eab6fa43dfcd86c9370dbed8/src/sampleworks/core/forward_models/xray/lunus_sf.py#L287

src/sampleworks/core/forward_models/xray/lunus_sf.py
mewall: Permalink: https://github.com/prism-science/sampleworks/blob/2ac07e20c23dcd42eab6fa43dfcd86c9370dbed8/src/sampleworks/core/forward_models/xray/lunus_sf.py#L41

src/sampleworks/core/forward_models/xray/lunus_sf.py
mewall: Permalink: https://github.com/prism-science/sampleworks/blob/2ac07e20c23dcd42eab6fa43dfcd86c9370dbed8/src/sampleworks/utils/elements.py#L72

src/sampleworks/core/forward_models/xray/lunus_sf.py
mewall: Permalink: https://github.com/prism-science/sampleworks/blob/2ac07e20c23dcd42eab6fa43dfcd86c9370dbed8/src/sampleworks/core/forward_models/xray/lunus_sf.py#L403

src/sampleworks/core/forward_models/xray/lunus_sf.py
mewall: Permalink: https://github.com/prism-science/sampleworks/blob/2ac07e20c23dcd42eab6fa43dfcd86c9370dbed8/src/sampleworks/core/forward_models/xray/lunus_sf.py#L428

src/sampleworks/core/forward_models/xray/lunus_sf.py
mewall: Permalink: https://github.com/prism-science/sampleworks/blob/2ac07e20c23dcd42eab6fa43dfcd86c9370dbed8/src/sampleworks/core/forward_models/xray/lunus_sf.py#L298

src/sampleworks/core/forward_models/xray/lunus_sf.py
mewall: Permalink: https://github.com/prism-science/sampleworks/blob/2ac07e20c23dcd42eab6fa43dfcd86c9370dbed8/src/sampleworks/core/forward_models/xray/lunus_sf.py#L417

src/sampleworks/core/forward_models/xray/lunus_sf.py
mewall: Permalink: https://github.com/prism-science/sampleworks/blob/2ac07e20c23dcd42eab6fa43dfcd86c9370dbed8/src/sampleworks/core/forward_models/xray/lunus_sf.py#L46

src/sampleworks/core/forward_models/xray/lunus_sf.py
mewall: Permalink: https://github.com/prism-science/sampleworks/blob/2ac07e20c23dcd42eab6fa43dfcd86c9370dbed8/src/sampleworks/core/forward_models/xray/lunus_sf.py#L86

src/sampleworks/synthetic/generate_synthetic_sf_lunus.py
mewall: Permalink: https://github.com/prism-science/sampleworks/blob/2ac07e20c23dcd42eab6fa43dfcd86c9370dbed8/src/sampleworks/core/forward_models/xray/lunus_sf.py#L66

src/sampleworks/synthetic/generate_synthetic_sf_lunus.py
mewall: Ignore the Claude Code response above. Here is a permalink relevant to the previous comment: https://github.com/prism-science/sampleworks/blob/2ac07e20c23dcd42eab6fa43dfcd86c9370dbed8/src/sampleworks/synthetic/generate_synthetic_sf_lunus.py#L58

src/sampleworks/synthetic/generate_synthetic_sf_lunus.py
mewall: Permalink: https://github.com/prism-science/sampleworks/blob/2ac07e20c23dcd42eab6fa43dfcd86c9370dbed8/src/sampleworks/synthetic/generate_synthetic_sf_lunus.py#L222

src/sampleworks/synthetic/generate_synthetic_sf_lunus.py
mewall: Permalink: https://github.com/prism-science/sampleworks/blob/2ac07e20c23dcd42eab6fa43dfcd86c9370dbed8/src/sampleworks/synthetic/generate_synthetic_sf_lunus.py#L205

src/sampleworks/synthetic/generate_synthetic_sf_lunus.py
mewall: Permalink: https://github.com/prism-science/sampleworks/blob/2ac07e20c23dcd42eab6fa43dfcd86c9370dbed8/src/sampleworks/synthetic/generate_synthetic_sf_lunus.py#L250

src/sampleworks/synthetic/generate_synthetic_sf_lunus.py
mewall: Permalink: https://github.com/prism-science/sampleworks/blob/2ac07e20c23dcd42eab6fa43dfcd86c9370dbed8/src/sampleworks/synthetic/generate_synthetic_sf_lunus.py#L135

src/sampleworks/synthetic/generate_synthetic_sf_lunus.py
mewall: Permalink: https://github.com/prism-science/sampleworks/blob/2ac07e20c23dcd42eab6fa43dfcd86c9370dbed8/src/sampleworks/synthetic/generate_synthetic_sf_lunus.py#L64

src/sampleworks/synthetic/generate_synthetic_sf_lunus.py
mewall: Permalink: https://github.com/prism-science/sampleworks/blob/2ac07e20c23dcd42eab6fa43dfcd86c9370dbed8/src/sampleworks/synthetic/generate_synthetic_sf_lunus.py#L378

src/sampleworks/synthetic/generate_synthetic_sf_lunus.py
mewall: Permalink: https://github.com/prism-science/sampleworks/blob/2ac07e20c23dcd42eab6fa43dfcd86c9370dbed8/src/sampleworks/synthetic/generate_synthetic_sf_lunus.py#L396

src/sampleworks/synthetic/generate_synthetic_sf_lunus.py
mewall: Permalink: https://github.com/prism-science/sampleworks/blob/2ac07e20c23dcd42eab6fa43dfcd86c9370dbed8/src/sampleworks/synthetic/generate_synthetic_sf_lunus.py#L35

src/sampleworks/synthetic/generate_synthetic_sf_lunus.py
mewall: Permalink: https://github.com/prism-science/sampleworks/blob/2ac07e20c23dcd42eab6fa43dfcd86c9370dbed8/src/sampleworks/synthetic/generate_synthetic_sf_lunus.py#L440

src/sampleworks/synthetic/generate_synthetic_sf_lunus.py
mewall: Permalink: https://github.com/prism-science/sampleworks/blob/2ac07e20c23dcd42eab6fa43dfcd86c9370dbed8/src/sampleworks/synthetic/generate_synthetic_sf_lunus.py#L673

tests/synthetic/test_generate_synthetic_sf_lunus.py
mewall: Permalink: https://github.com/prism-science/sampleworks/blob/2ac07e20c23dcd42eab6fa43dfcd86c9370dbed8/tests/synthetic/test_generate_synthetic_sf_lunus.py#L104

tests/synthetic/test_generate_synthetic_sf_lunus.py
mewall: https://github.com/prism-science/sampleworks/blob/2ac07e20c23dcd42eab6fa43dfcd86c9370dbed8/tests/synthetic/test_generate_synthetic_sf_lunus.py#L104

tests/synthetic/test_generate_synthetic_sf_lunus.py
mewall: https://github.com/prism-science/sampleworks/blob/2ac07e20c23dcd42eab6fa43dfcd86c9370dbed8/tests/synthetic/test_generate_synthetic_sf_lunus.py#L104

tests/synthetic/test_generate_synthetic_sf_lunus.py
mewall: https://github.com/prism-science/sampleworks/blob/2ac07e20c23dcd42eab6fa43dfcd86c9370dbed8/tests/synthetic/test_generate_synthetic_sf_lunus.py#L330

tests/synthetic/test_generate_synthetic_sf_lunus.py
mewall: https://github.com/prism-science/sampleworks/blob/2ac07e20c23dcd42eab6fa43dfcd86c9370dbed8/tests/synthetic/test_generate_synthetic_sf_lunus.py#L182

PR #383, opened 2026-08-18
chore(deps): declare lunus and relock

manzuoni-astera:
Unblocks the pixi issue you raised. You can declare lunus after all, and keep it workspace-level so it is available on macOS too.

## What this is

The lunus.sf declaration your `scripts/synthetic` and guidance code already import, plus `pixi.lock` regenerated on Linux. All 13 environments carry lunus at rev fb22633, `boltz-osx` included. Two files, 48 added lines.

The lock here was produced by the Relock workflow, not by hand: https://github.com/manzuoni-astera/sampleworks/actions/runs/32158645024

## Why the earlier attempts failed

Not the workspace shape. The lock going in still held lunus at `60bd44d` from the `branch = "sf"` resolve on the 15th, while the manifest had moved to `rev = fb22633` and then to a linux-only feature. When the lock disagrees with the manifest about a git rev, the affected environments go stale and get re-resolved from scratch, and a from-scratch osx-arm64 resolve needs an osx prefix the Linux runner cannot create. That is the `boltz-osx` error.

Reset the lock first and the change becomes additive, which resolves fine. I ran both shapes from a clean lock on ubuntu-latest to be sure:

| shape | result |
| --- | --- |
| lunus workspace-level | all 13 envs, 35s ([run](https://github.com/manzuoni-astera/sampleworks/actions/runs/32151923800)) |
| lunus on the linux-only boltz feature | the 3 boltz envs ([run](https://github.com/manzuoni-astera/sampleworks/actions/runs/32151908607)) |

This PR uses the first, since it also gives you lunus on your Mac. Say the word if you would rather have the scoped version.

## Two things worth correcting

The `sampleworks` editable path dep is not what forces the per-platform builds. It has no `dynamic` fields, so pixi reads its metadata statically (`using cached static metadata package=sampleworks`, 19 times in the debug log) and never builds it. The builds come from sdist-only deps: `modelcif` (boltz pins ==1.2, protenix pins ==1.4, wheels only start at 1.6), `ihm` (no wheel at any version), plus deepspeed, fairscale, antlr4-python3-runtime, assertpy, scikit-learn-extra.

And osx-arm64 is the platform your Mac can solve, not the one it cannot. The asymmetry is the other way round: locally, solving any linux-64 environment needs a linux-64 prefix, which is why local relocks die on `protenix`. Treat Relock as the only way to regenerate the lock.

You are right that `pixi lock` has no `--environment` or `--platform`, so there is no partial lock. Confirmed on 0.73.0.

## Later

lunus.sf is pure Python, so a published `py3-none-any` wheel would take it out of the source-dep set entirely. Does nothing for modelcif and ihm, but it makes your own dependency free.

## Testing

Relock green on ubuntu-latest, lock covers all 13 environments at the pinned rev. I have not run the test suites; this only adds a dependency.

## Rollout

Merge into `mew/diffuse-1` and carry on. If you change the rev later, reset `pixi.lock` to a state without lunus before re-running Relock.

### DIFF
--- a/pixi.lock
+++ b/pixi.lock
@@ -111,6 +111,7 @@ environments:
       - conda: https://conda.anaconda.org/conda-forge/osx-arm64/tk-8.6.13-h010d191_3.conda
       - conda: https://conda.anaconda.org/conda-forge/osx-arm64/zstd-1.5.7-hbf9d68e_6.conda
       - pypi: ./
+      - pypi: git+https://github.com/lanl/lunus.git?rev=fb22633f3279baf3e97f207d0bea4ce34733e6f8#fb22633f3279baf3e97f207d0bea4ce34733e6f8
       - pypi: https://files.pythonhosted.org/packages/04/4b/29cac41a4d98d144bf5f6d33995617b185d14b22401f75ca86f384e87ff1/h11-0.16.0-py3-none-any.whl
       - pypi: https://files.pythonhosted.org/packages/04/96/92447566d16df59b2a776c0fb82dbc4d9e07cd95062562af01e408583fc4/itsdangerous-2.2.0-py3-none-any.whl
       - pypi: https://files.pythonhosted.org/packages/05/7f/798705f5296a58ca505d600456748d1be48078eac8a7050d8a98bc9edb89/decorator-5.3.1-py3-none-any.whl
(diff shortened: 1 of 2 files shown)


mewall (commented):
@manzuoni-astera What do you think about the following going forward?

---
If you change the rev later, reset pixi.lock to a state without lunus before re-running Relock.
---

Should I follow the recommendation to publish an arch-independent pure python wheel instead, which, if I understand correctly, will bypass the Relock step?

PR #390, opened 2026-08-24
fix(utils): clean temporary altloc CIFs

### LINKED ISSUE #339: Improve temp file and error handling when resolving mixed hetatm altlocs.
Regarding the use of `resolve_mixed_hetatm_atom_altlocs`, from PR 336: 

Temporary CIF cleanup can be skipped if load_any(safe_reference_path) throws: resolve_mixed_hetatm_atom_altlocs() creates a delete=False tempfile, but the current unlink() only runs after a successful load_any(). Also, a cleanup unlink() failure currently aborts patching and is reported as a parse failure. Wrapping the load + subsequent parsing in a try/finally and swallowing unlink errors avoids leaking temp files and avoids false negatives.

We should fix this not in that PR but in all locations. One way to handle this would be to encapsulate all the required logic in `resolve_mixed_hetatm_atom_altlocs` directly.

manzuoni-astera:
## Summary
Fixes #339. Ensure temporary altloc-fixed CIFs are cleaned on every path.

## Impact
Parse/save failures no longer leak files; cleanup failures do not mask valid parsing.

## Changes
- `src/sampleworks/utils/cif_utils.py:242` cleans failed writes.
- `src/sampleworks/utils/guidance_script_utils.py:278` guarantees best-effort cleanup.
- Regression coverage is in `tests/utils/test_{cif_utils,guidance_script_utils}.py`.

## Testing
- Focused pixi pytest: 40 passed.
- Ruff and diff checks passed.

## Deployment
None.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved cleanup of temporary CIF files when structure conversion or parsing fails.
  - Ensured cleanup errors do not mask the original error or prevent successful structure processing.
  - Added support for consistent handling of structure paths provided as either strings or path objects.

- **Tests**
  - Added coverage for temporary-file cleanup, failure handling, and path input types.


src/sampleworks/utils/guidance_script_utils.py
```diff
@@ -275,18 +275,21 @@ def get_reward_function_and_structure(
...
-        add_missing_atoms=False,
-        ccd_mirror_path=None,
-    )
-
-    # make sure to cast paths to strings, since Path(x) != str(x) we don't want to
-    # accidentally delete the originals.
-    if str(safe_structure_path) != str(structure_path):
-        safe_structure_path.unlink()  # delete the temporary file if it was created
+    structure_path = Path(structure_path)
+    safe_structure_path = resolve_mixed_hetatm_atom_altlocs(structure_path)
```
marcuscollins, line 279: Can you check that we have a unit test that `resolve_mixed_hetatm_atom_altlocs` always returns a `Path` object? I don't want to end up with the opposite mistake than what I made before, where structure_path is now a Path, but safe_structure_path is a string.

marcuscollins (approved):
One thing to check--whether we have a unit test that ensures `resolve_mixed_hetatm_atom_altlocs` always returns a `Path`. If we have/can add that, then LGTM.

src/sampleworks/utils/guidance_script_utils.py
xraymemory: Done! Added test_always_returns_path in d6d9155: parametrized over both return branches (no-change passthrough and fixed-copy) × both input types (Path and str), asserting isinstance(result, Path). The function coerces at the top (cif_path = Path(cif_path)), so all four combinations pass

PR #336, opened 2026-07-22
fix(evals): cif patching regression to allow protenix outputs to run on all evals, and remove hydrogens in RSCC parsing

k-chrispens:
There are a few regressions that needed to be made here to get the evals running completely for the paper. These changes are archived in the zenodo record but should be applied here for consistency until we fix the Protenix JSON parsing issue.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved structure analysis by handling missing atoms and excluding hydrogens where appropriate.
  * Improved downloaded reference structure processing, including safer alternate-location handling.
  * Enhanced residue and chain remapping for more consistent CIF output.
  * Improved recognition of structure identifiers in folder names.

* **Chores**
  * Expanded synchronized checkout exclusions to include TAR archives, generated output, and repository metadata directories.

PR #395, opened 2026-08-27
fix: make sure we don't load hydrogens or impute missing atoms in RSCC script.

marcuscollins:
This change restores the `hydrogen_policy="remove", add_missing_atoms=False` arguments to `atomworks.io.parse` calls, consistent with the Stacks paper. This is necessary to get the correct results out of the RSCC script, as it fails badly or gives inaccurate results if we compute metrics using NaN coordinate inputs.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved structure parsing by consistently removing hydrogens and preventing automatic addition of missing atoms.
  * Adjusted trial-processing logs to appear before refined-structure parsing for clearer progress reporting.

### DIFF
--- a/scripts/eval/rscc_grid_search_script.py
+++ b/scripts/eval/rscc_grid_search_script.py
@@ -134,7 +134,9 @@ def process_group(
                 f"Could not find reference structure for occupancy {trials[0].altloc_occupancies}"
             )
         # parse() returns only the first altloc.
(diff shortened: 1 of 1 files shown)


k-chrispens (approved):
Approve

marcuscollins:
Closing as it was a duplicate of #336

PR #399, opened 2026-09-01
Enabling ext by default

levon-h-ast:
| Failure mode | command -v ext | ext --help |
|:---|:---|:---|
| not installed at all | catches | catches (127) |
| present, no execute bit | misses | catches (126) |
| missing shared lib / wrong arch | misses | catches (126/127) |
| segfaults on startup | misses | catches (139) |
| ext's own flag parsing broken | misses | catches (2) |
| --help ok, `ext shell` fails (auth/config) | misses | misses |
| hangs instead of exiting | misses | misses (stalls startup) |

## Summary by CodeRabbit

- **New Features**
  - The `ext` shell now starts automatically by default.
  - Set `EXT_SHELL=0` before starting the shell to opt out.
- **Bug Fixes**
  - Improved validation ensures the `ext` command is functional before handing over the shell, preventing startup issues when the command is unavailable or broken.

### DIFF
--- a/docker/astera/sampleworks-ext-shell.sh
+++ b/docker/astera/sampleworks-ext-shell.sh
@@ -1,5 +1,5 @@
-# Enter ext only when ACTL sets EXT_SHELL=1.
-[ "${EXT_SHELL:-}" = "1" ] || return 0 2>/dev/null || exit 0
+# Enter ext by default. Set EXT_SHELL=0 before shell start to opt out.
+[ "${EXT_SHELL:-1}" != "0" ] || return 0 2>/dev/null || exit 0
 
 case "$-" in
   *i*) ;;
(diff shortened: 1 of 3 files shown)


docker/astera/sampleworks-ext-shell.sh
```diff
@@ -12,7 +12,9 @@ esac
 [ -z "${EXT_SHELL_ACTIVE:-}" ] || return 0 2>/dev/null || exit 0
 [ -z "${BASH_EXECUTION_STRING:-}" ] || return 0 2>/dev/null || exit 0
 [ -z "${ZSH_EXECUTION_STRING:-}" ] || return 0 2>/dev/null || exit 0
-command -v ext >/dev/null 2>&1 || return 0 2>/dev/null || exit 0
+# Not just "is ext on PATH" but "does ext actually run" - past this point the
+# shell is handed over with exec, so a broken binary would end the session.
+ext --help >/dev/null 2>&1 || return 0 2>/dev/null || exit 0
```
denis-zaitsev, line 17: we gonna print help on every sh into container? seems really unnescessary

levon-h-ast: `ext --help` is redirecting to `/dev/null`, so no output in the console. You can see the reasoning of this change in the description of the PR.

denis-zaitsev (approved):

PR #400, opened 2026-09-02
refactor(utils): move structure_utils out of eval

### LINKED ISSUE #355: refactor: move structure_utils out of eval
`eval/structure_utils.py` now hosts core sampling-input logic (e.g. SampleworksProcessedStructure, process_structure_to_trajectory_input) used by the scalers, so its `eval/` location is misleading.

Probably should move it to `src/sampleworks/utils/`, and move `tests/eval/test_structure_utils.py` to `tests/utils/test_structure_utils.py`? This would also require updating the import lines in other files that use them.

xraymemory:
## Summary

Fixes #355. `eval/structure_utils.py` hosts core sampling-input logic (`SampleworksProcessedStructure`, `process_structure_to_trajectory_input`) used by both trajectory scalers and the Boltz/Protpardelle wrappers, so its `eval/` location was misleading. This is the move the issue asks for, nothing more.

## Changes

- `git mv src/sampleworks/eval/structure_utils.py src/sampleworks/utils/structure_utils.py` (rename; content untouched).
- `git mv tests/eval/test_structure_utils.py tests/utils/test_structure_utils.py` — its only fixtures come from the root `tests/conftest.py` (`mock_protein_config` is defined in the file itself), so no fixture plumbing moved.
- Updated the 16 importers (5 `scripts/eval`, 2 scalers, 2 model wrappers, 7 test files) and the `AGENTS.md` pointer; import blocks re-sorted where the rename changed isort order.

## Notes

- No compatibility shim at the old path: every in-repo reference is updated, and grep finds no string-based (`mock.patch`) references to the old module path.
- `structure_utils` keeps importing `ProteinConfig` from `eval.eval_dataclasses`. That dependency predates this move and creates no cycle (`eval_dataclasses` doesn't import back). Whether `ProteinConfig` should also move is a separate question.

## Testing

`boltz-osx` env: full suite `pytest tests -m 'not slow and not gpu'` → **907 passed, 31 skipped**; ruff check + format clean; `ty` clean on the moved file, both scalers, and the moved test.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reward preparation during structure sampling, including support for model-specific atom topology and reconciled coordinates and B-factors.
  - Preserved source structure data while constructing atom arrays for forward-model evaluation.

- **Refactor**
  - Updated structure utility references across evaluation scripts, model components, tests, and documentation.
  - Standardized related import organization without changing application behavior.

- **Documentation**
  - Clarified reward-input construction and two-phase reward preparation guidance.

- **Tests**
  - Added coverage for atom-array reconstruction and reward preparation across trajectory scalers.


smallfishabc:
I think this looks good. Before merge, could you rebase to the current version. And can I ask why do we need to move it to utils?

DorisMai:
> I think this looks good. Before merge, could you rebase to the current version. And can I ask why do we need to move it to utils?

Marcus suggested moving it a while ago because the functions in `structure_utils.py` are already used beyond eval and some of them likely will be refactored by me for the synthetic module.

PR #401, opened 2026-09-03
feat(eval): add option to further restrict atoms in the RSCC calculation

marcuscollins:
In the process of analyzing some data, I realized that, in the RSCC script, while we restrict the volume of comparison to be around only the selected atoms in the reference structure, we do not have the option to restrict which atoms from the generated ensemble structures are used to compute the map. This means that sometimes a side chain will be completely out of expected density, but some other side chain will partially take its place, artificially inflating the scores. While technically correct, it may still be useful to omit side chains outside of the selection to get numbers that better reflect our intuition of what is happening. The attached figure shows an example from 3HYN Lys121, where our original calculation gave an RSCC ~0.85, which doesn't reflect that the Lys121 is completely out of the expected density.
3HYN_121_v2

This PR adds an option to the RSCC script --selected-residues-only, which compares the density of the residues in the selection exclusively. With this option turned on, the above example RSCC drops to ~0.57, reflecting the poor fit.

## Summary by CodeRabbit

* **New Features**
  * Added an option to evaluate refined structures using density from only the selected residues.
  * Added support for legacy and atomworks-style selection strings.
  * Added the `--selected-residues-only` command-line flag for evaluation workflows.

* **Enhancements**
  * Correlation results can now focus specifically on selected residues instead of the entire structure.

### DIFF
--- a/scripts/eval/rscc_grid_search_script.py
+++ b/scripts/eval/rscc_grid_search_script.py
@@ -29,6 +29,7 @@ from pathlib import Path
 import numpy as np
 import pandas as pd
 import torch
+from biotite.structure import AtomArray, AtomArrayStack
 
 # Import local modules for density calculation
 from joblib import delayed, Parallel
@@ -42,7 +43,9 @@ from sampleworks.eval.structure_utils import (
     get_reference_structure_coords,
 )
 from sampleworks.utils.atom_array_utils import (
+    ATOMWORKS_COMPARISON_OPS,
     filter_to_common_atoms,
+    get_mask_from_old_selection_string,
     parse_structure,
     remove_atoms_with_any_nan_coords,
 )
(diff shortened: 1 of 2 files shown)


scripts/eval/rscc_grid_search_script.py
```diff
@@ -60,13 +63,53 @@
...
 
+def filter_to_selection(
+    atom_array: AtomArray | AtomArrayStack, selection: str
+) -> AtomArray | AtomArrayStack:
+    """Restrict an atom array to the atoms matching ``selection``.
+
+    Mirrors the masking used by
+    :func:`sampleworks.eval.structure_utils.extract_selection_coordinates` (so the
+    filtered atoms correspond to the same residues as the reference selection
+    coordinates), while preserving every model of an ``AtomArrayStack``.
```
DorisMai, line 74: maybe make an issue to refactor `extract_selection_coordinates`? The selection logic is duplicative

scripts/eval/rscc_grid_search_script.py
```diff
@@ -258,6 +309,14 @@ def process_group(
                         raise ValueError(f"Extracted base map empty for selection {selection}")
                     extracted_base_cache[selection] = extracted_base
 
+                if selected_residues_only:
+                    # Compute density from only this selection's atoms on the shared grid, so
+                    # the extracted region carries no signal from the surrounding structure.
+                    selected_atoms = filter_to_selection(atom_array, selection)
+                    selected_density = run_density_transformer(transformer, selected_atoms)
+                    computed_xmap = copy.copy(base_xmap)
+                    computed_xmap.array = selected_density.cpu().numpy()
```
DorisMai, line 318: How local is the density computation? The base_xmap is still computed from the entire protein, could other atoms nearby but outside the selection affect the density? In otherwords, do we get rscc = 1.0 even if the predicted and ref structures are identical under `selected_residues_only`?

DorisMai (approved):
lgtm, minor comments:
- please address CI test failure
- Currently this diff allows density computed from predicted_coords queried at ref_coords. Would it be useful to also look at density computed from ref_coords queried at predicted coords? or maybe rscc from a union of the two coords?

scripts/eval/rscc_grid_search_script.py
marcuscollins: Good catch, I'll refactor accordingly.

scripts/eval/rscc_grid_search_script.py
marcuscollins: Very local. The method `extract_tight` already extracts the part of the map around the target (reference structure) residues. This PR introduces the ability to additionally only calculate the predicted ensemble map from the atoms in the same residue selection (that's why we use `run_density_transformer(transformer, selected_atoms)`). The the computed_xmap is zero anywhere there aren't atoms from the predicted ensemble that match the residue selection. I've tried to clarify this in the comments.

PR #402, opened 2026-09-05
feat(rewards): prepare() takes RewardInputs, not an atom array

xraymemory:
## Summary

Follow-up to #373, from the 4 Sep rewards discussion with Marcus and Doris: `prepare` takes the
`RewardInputs` rather than an atom array, so a two-phase reward is prepared against exactly what
its `__call__` is later fed, and nothing is written back into the model atom array.

Why. `to_reward_inputs` reconciles the structure's B-factors and reference coordinates onto the
model atom order, but those values live only in the returned `RewardInputs`. The model atom array
keeps its template placeholders (Boltz: flat B = 20; RF3: noise on unresolved atoms), and nothing
writes the reconciled values back, deliberately. #373 prepared the SF reward from that array, so
the SFcalculator, and its solvent-fraction estimate, were built from placeholder coordinates.
Doris raised this on #373 on 15 Aug; the meeting settled on passing the inputs rather than
overwriting the array.

## Changes

`PreparableRewardFunctionProtocol.prepare(reward_inputs, *, device) -> None` and
`prepare_reward_if_needed(reward, reward_inputs, *, device)` in `core/rewards/protocol.py`. Both
trajectory scalers pass the `RewardInputs` they attach to the step context, right after
`to_reward_inputs` and before the denoising loop.

`RewardInputs` gains `atom_array`, the topology the tensors were built from (set by
`from_atom_array`; the first model for an `AtomArrayStack`; `None` when the dataclass is built
from bare tensors, as tests do), and `to_atom_array()`, which copies that topology and overwrites
`coord`, `b_factor` and `occupancy` from the tensors (occupancy 1.0: a topology, not an ensemble).
That is Doris's converter from `dm/reward-prepare-hook`, fed by the inputs themselves.
`StructureFactorRewardFunction.prepare` builds its gemmi structure from it. Nothing is written into
the model atom array or `SampleworksProcessedStructure`; `reward_atom_array` stays as the property
`to_reward_inputs` takes its topology from.

The two rewards queued behind this (#319, #372) disagree with the old signature anyway, so this
is the moment to settle it. #374 and #375 are updated to it (`CompositeReward.prepare`, the
build-integration test).

## Testing

`tests/rewards/test_prepare_hook.py`: the helper hands a two-phase reward the very
`RewardInputs` object and device it was given (identity, not a copy), leaves a one-phase reward
alone, and can be re-run to rebind.

`tests/rewards/test_reward_inputs.py`: `from_atom_array` keeps the source array (first model of a
stack); `to_atom_array` round-trips coordinates, B-factors and annotations, takes coordinates and
B-factors from the tensors when they differ from the template, sets occupancy to 1.0, leaves the
template untouched, and rejects a template with the wrong atom count or inputs with no topology.

`tests/integration/test_mismatch_integration.py`: on every mismatch case, the array
`to_atom_array` rebuilds has the model's annotations and the reward inputs' coordinates and
B-factors, and the model atom array is byte-for-byte what it was before.

`tests/integration/test_pipeline_integration.py`: both trajectory scalers prepare a recording
reward exactly once, before the first evaluation, with inputs whose topology is the model's
four-atom array (the structure has five).

The SF reward tests (`make_prepared_reward`, the `reward_function_1vme_sf` fixture) wrap their atom
array in single-conformer `RewardInputs`; the tests themselves are unchanged.

Local, `boltz-osx` env, on the #373 branch this was developed on: `pytest tests -m 'not slow and
not gpu'` → 929 passed, 31 skipped. The gpu-marked SF tests forced onto CPU: all 13 SF reward tests
(occupancy, every bulk-solvent mode, config) and the 16 structure-factor contract tests passed.
Ruff and ty clean on the changed files. Cherry-picked onto `main` after #373 merged; the trees
were identical, so the pick applied cleanly.

## Rollout

No new dependencies, no CLI or configuration changes, and no behavior change for any reward that
doesn't implement `prepare`. The `prepare` signature changes for #374 and #375, which already
carry it. Merge this before them.

## Summary by CodeRabbit

* **Enhancements**
  * Improved reward preparation to preserve model atom topology while using reconciled coordinates and B-factors.
  * Added support for rebuilding a reference structure from reward inputs, including consistent occupancy values.
  * Updated structure-factor rewards and sampling workflows to use complete reward inputs for improved model/structure alignment.

* **Bug Fixes**
  * Added validation for missing topology and mismatched atom counts.
  * Ensured source model atom arrays remain unchanged during reward preparation.

* **Documentation**
  * Expanded guidance on handling atom-count mismatches and reward-input alignment.

### DIFF
--- a/src/sampleworks/core/rewards/protocol.py
+++ b/src/sampleworks/core/rewards/protocol.py
@@ -26,12 +26,20 @@ class RewardInputs:
     all coordinates finite and all occupancies non-negative.  Wrappers are
     responsible for ensuring this (e.g. replacing NaN coordinates with
     noise and setting occupancy to 1.0 for model-operated atoms).
+
+    ``atom_array`` is the topology the tensors follow: the array they were built
+    from, kept for its per-atom identity (chain, residue, atom name, element,
+    altloc). It is never mutated, and its own ``coord``, ``b_factor`` and
+    ``occupancy`` are not authoritative; the tensors are. :meth:`to_atom_array`
+    combines the two for a forward model that needs an ``AtomArray`` (see
+    :class:`PreparableRewardFunctionProtocol`).
     """
 
     elements: Int[torch.Tensor, "*batch n_atoms"]
     b_factors: Float[torch.Tensor, "*batch n_atoms"]
     occupancies: Float[torch.Tensor, "*batch n_atoms"]
     input_coords: Float[torch.Tensor, "*batch n_atoms 3"]
+    atom_array: AtomArray | None = None
 
     @classmethod
     def from_atom_array(
@@ -62,7 +70,9 @@ class RewardInputs:
         Returns
         -------
         RewardInputs
-            Dataclass containing all inputs needed for reward function computation.
+            Dataclass containing all inputs needed for reward function computation,
+            with ``atom_array`` set to the topology they were built from (the first
+            model of an ``AtomArrayStack``).
         """
         # input validation: ensure atom_array has required annotations and valid values
         if not hasattr(atom_array, "element"):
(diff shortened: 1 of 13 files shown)


src/sampleworks/core/rewards/protocol.py
```diff
@@ -128,13 +138,85 @@ def from_atom_array(
         if isinstance(device, str):
             device = torch.device(device)
 
+        # Keep the topology the tensors follow. A stack shares one topology across
+        # its models, so its first model stands for it.
+        from biotite.structure import AtomArrayStack
```
DorisMai, line 143: maybe just drop the `if TYPE_CHECKING:` at the top of the file and import and module level

src/sampleworks/core/rewards/protocol.py
```diff
@@ -26,12 +26,20 @@ class RewardInputs:
...
+    ``occupancy`` are not authoritative; the tensors are. :meth:`to_atom_array`
+    combines the two for a forward model that needs an ``AtomArray`` (see
+    :class:`PreparableRewardFunctionProtocol`).
     """
 
     elements: Int[torch.Tensor, "*batch n_atoms"]
     b_factors: Float[torch.Tensor, "*batch n_atoms"]
     occupancies: Float[torch.Tensor, "*batch n_atoms"]
     input_coords: Float[torch.Tensor, "*batch n_atoms 3"]
+    atom_array: AtomArray | None = None
```
DorisMai, line 42: I would suggest naming it to `topology_atom_array` or `template_atom_array` to emphasize that this atom array is not intended for attributes like coordinates. For consistency, I would also suggest renaming accordingly the argument `template` in `to_atom_array`, as well as `structure_utils.py:55 SampleworksProcessedStructure.reward_atom_array` (currently outside the diff) which does not necessarily hold reconciled coordinates/b-factors either.

tests/integration/test_mismatch_integration.py
```diff
@@ -701,6 +701,54 @@ def test_b_factor_override(self, mismatch_case: MismatchCase):
...
+        template_b_factors = processed.reward_atom_array.b_factor.copy()
+        template_coords = processed.reward_atom_array.coord.copy()
+        assert not np.allclose(template_b_factors, reward_inputs.b_factors[0].numpy()), (
+            "setup failed to make the template and the reward inputs disagree"
+        )
+
+        prepared = reward_inputs.to_atom_array()
+
+        assert prepared.array_length() == mismatch_case.n_model
+        for category in ["atom_name", "res_id", "chain_id", "element"]:
```
DorisMai, line 739: maybe add in "res_name" and "hetero" too

DorisMai (commented):
Thanks for taking the lead on this!! I left a few minor comments. The main thing is that I thought last Friday @marcuscollins mentioned directly supplying/carrying topology information for `prepare()` rather than convert `RewardInputs` to `AtomArray`. 

I personally still find the conversion to AtomArray more intuitive because:
- The topology fields to be added (chain_id, res_id, res_name, atom_name, ins_code, hetero) are not really inputs to the reward `__call__()` function
- When converting to gemmi Structure, we will reinvent biotite’s `get_residue_starts` and maybe other utils if avoiding AtomArray?

That said, neither point is severe. I also understand that conversion to gemmi is uniquely required by SFCalculator (lunus or TorchRef does not seem to need this), which might be another point to consider for medium to long term. @marcuscollins could you clairfy how you see the `RewardInputs` to be?

DorisMai (approved):
After discussion with Marcus: we should move forward with the current implementation. Converting RewardInputs to AtomArray isn't the best long-term solution, but the interface where prepare() takes RewardInputs is the right abstraction. Approving to merge now, the best way to represent RewardInputs will be addressed in the future.

PR #408, opened 2026-09-15
fix: record the PDB id, not the run name, as protein in job metadata

### LINKED ISSUE #393: BUG:
```
root@actl-ws-marcuscollins-sampleworks-devspace-7f67d5b4b8-hv6dl:/mnt/diffuse-shared/results/sampleworks/actl-ws-marcuscollins-sampleworks-devspace-74675b59c9-v52bz/protpardelle/protpardelle/3T94_1.0occA/protpardelle_X-RAY_DIFFRACTION/pure_guidance/ens8_gw0.1# cat job_metadata.json 
{"protein": "3T94_1.0occA", "structure": "/mnt/diffuse-shared/raw/sampleworks/initial_dataset_40_occ_sweeps/processed/3T94/3T94_single_001_density_input.cif", "density": "/mnt/diffuse-shared/raw/sampleworks/initial_dataset_40_occ_sweeps/occ_sweeps/1.0occA/density_maps/3T94_1.0occA_1.00A.ccp4", "model_name": "protpardelle", "guidance_type": "pure_guidance", "log_path": "/mnt/diffuse-shared/results/sampleworks/actl-ws-marcuscollins-sampleworks-devspace-74675b59c9-v52bz/protpardelle/protpardelle/3T94_1.0occA/protpardelle_X-RAY_DIFFRACTION/pure_guidance/ens8_gw0.1/run.log", "output_dir": "/mnt/diffuse-shared/results/sampleworks/actl-ws-marcuscollins-sampleworks-devspace-74675b59c9-v52bz/protpardelle/protpardelle/3T94_1.0occA/protpardelle_X-RAY_DIFFRACTION/pure_guidance/ens8_gw0.1", "partial_diffusion_step": 300, "loss_order": 2, "resolution": 1.0, "device": "cuda:1", "gradient_normalization": true, "em": false, "guidance_start": 400, "augmentation": false, "align_to_input": true, "alignment_reverse_diffusion": null, "recycling_steps": null, "num_diffusion_steps": 500, "step_size": 0.1, "step_scaler_type": "noisespace", "model_checkpoint": "/mnt/diffuse-shared/marcus/model_params/weights/cc89_epoch415.pth", "protpardelle_config_path": null, "ensemble_size": 8, "method": null, "scaler": "pure_guidance", "gradient_weight": -1.0, "gd_steps": -1, "status": "success", "exit_code": 0, "runtime_seconds": 16.86, "started_at": "2026-07-27T22:45:29.359366", "finished_at": "2026-07-27T22:45:46.223141", "altloc_occupancies": {"A": 1.0}}
```
note that the key "protein" has value "3T94_1.0occA". It should just be the PDB id, "3T94". This causes problems later in analysis since the scripts will look for a protein config for this string, and we only create them per protein id.

Note as a temporary fix so I can analyze my data, I am going to fix all the metadata files and move the old versions to `job_metadata.json.bak`

rtviii:
- **fix(eval): add rcsb_id_from_protein_name for grid-search protein names**
- **fix(utils): record the PDB id as protein in job_metadata.json**
- **fix(eval): resolve the PDB id from protein names in existing job metadata**

This should fix #393.

`job_metadata.json` records `protein` as the grid-search protein name rather than the PDB id, e.g. `"protein": "3T94_1.0occA"`. The name is the `name` column of the `--proteins` CSV and passes unchanged through `JobConfig`, `GuidanceConfig` and `JobResult` into `_write_job_metadata`. Since #321, `scan_grid_search_results` prefers the metadata value over the one parsed from the directory name. Protein configurations are keyed by PDB id, so the RSCC, LDDT and RMSD scripts log "Skipping protein with no configuration" and drop every occupancy-sweep trial.

## Change

- `RCSB_ID_PATTERN` in `utils/cif_utils.py` and `rcsb_id_from_protein_name` in `eval/occupancy_utils.py`. The function returns the RCSB id (legacy `3T94` or extended `pdb_00003t94`) when the whole name is an id followed only by occupancy tokens, and `None` otherwise; the id's case is kept. The occupancy-token regex that `extract_protein_and_occupancy` had inline becomes the shared `OCCUPANCY_TOKEN_PATTERN`, with no change in behavior.
- `_write_job_metadata` writes the id as `protein` and the full name as `protein_name`. Names that are not of that form, such as `lysozyme_1.0occA`, are written unchanged.
- `scan_grid_search_results` applies the same resolution to metadata already on disk, so existing runs evaluate without rewriting their files.

The name is not changed where it enters the pipeline: `run_grid_search.py` uses it for output directories and in the `results.json` merge key, and occupancies are parsed from it.

this is the first of several small PRs that will replace what i tried in #301.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved recognition of protein identifiers in grid-search results, including standard and extended RCSB ID formats.
  - Preserved complete protein names and occupancy details while consistently recording the associated protein identifier.
  - Improved handling of valid occupancy annotations, including case variations.
  - Invalid or ambiguous protein names are safely retained without being incorrectly converted.


src/sampleworks/utils/cif_utils.py
```diff
@@ -18,6 +19,11 @@
 )
 
 
+# Extended PDB IDs use `pdb_` plus eight alphanumerics. Legacy IDs have four
+# alphanumerics and start with a digit, which excludes folder names such as `TEST`.
+RCSB_ID_PATTERN = re.compile(r"pdb_[A-Za-z0-9]{8}|[0-9][A-Za-z0-9]{3}")
```
marcuscollins, line 24: This same regex currently appears in `patch_output_cif_files.py`, as _VALID_RCSB_ID. Can you replace that version and import this one there, so we don't have duplicates.

src/sampleworks/eval/occupancy_utils.py
```diff
@@ -1,5 +1,15 @@
 import re
 
+from sampleworks.utils.cif_utils import RCSB_ID_PATTERN
+
+
+# `{value}occ{label}` altloc occupancy token, e.g. `0.25occA`, as written by occupancy_to_str.
+OCCUPANCY_TOKEN_PATTERN = re.compile(r"(\d+\.?\d*)(?i:occ)([A-Za-z])")
```
marcuscollins, line 7: are we sure only the "occ" part of the string needs to be case insensitive?

src/sampleworks/eval/occupancy_utils.py
```diff
@@ -34,13 +44,45 @@ def extract_protein_and_occupancy(dir_name: str) -> tuple[str, dict[str, float]]
     protein = dir_name.split("_")[0].lower()
 
     altloc_occupancies: dict[str, float] = {}
-    for match in re.finditer(r"(\d+\.?\d*)occ([A-Za-z])", dir_name, re.IGNORECASE):
+    for match in OCCUPANCY_TOKEN_PATTERN.finditer(dir_name):
```
marcuscollins, line 47: Why get rid of `re.IGNORECASE`?

marcuscollins (approved):
A few things to double check/add to the PR, otherwise approved.

Issue #68, opened 2026-01-29
Fix CIF output files


marcuscollins:
We need to have, at least, entity ids and we need to follow the requirement that _every_ atom in the atom_site loop has a unique `id` field. It does not, surprisingly, have to be sequential or numeric. It can be any unique string (this may be useful for very large proteins or large numbers of models, and therefore we may want to use alphanumeric strings)

The easiest solution to making sure all header information is present would be, for now, to copy it from the original input cif file.

manzuoni-astera:
Wrote this up in the consolidated design doc — PR #303 (`docs/design/cif-output-correctness.md`).

**Scope correction worth noting:** the "every `atom_site` row needs a unique `id`" requirement is **already satisfied** by our native writers. I verified on the pinned **biotite 1.4.0** that a multi-model ensemble stack writes `id = 1..N` across all rows (unique, including across models). The patch script overwrites `id` (`patch_output_cif_files.py:245`) only because it edits a *template* CIF whose original id column no longer matches after res-id remap / NaN drop — not because our generated output is invalid.

So the real remaining native gaps are:
- **entity / `entity_poly`** — never constructed by the writers; only Protenix even carries `label_entity_id`, so this is cross-model bookkeeping.
- **header** (cell, symmetry, struct, entity_poly, pdbx_poly_seq_scheme) — dropped at `parse()` time. The doc's Phase 1 proposes copying a whitelist of header categories from the input CIF at write time (the "copy from the original input cif" approach you suggested), with the atom-alignment caveats spelled out.

Phased plan + open questions are in §6–7. No code change yet — this is the catalog/plan step so we don't hack at it piecemeal.

PR #413, opened 2026-09-16
fix: write tortoize-readable multi-model CIFs

rtviii:
Further replacing #301. Refs #68.

- `set_structure()` copies the `atom_id` annotation into `_atom_site.id` and tiles every column per model, so a multi-model file has duplicate values in the category key. 
- Main's writer emits only `atom_site`, `cell` and `struct_conn`. With one model libcifpp rebuilds the missing `entity_poly_seq` from `atom_site` and tortoize works; with several models it rebuilds it from every model's rows and hits a duplicate key violation.

The 4 check on a carved 1vme:

```
ids duplicated, no entities  -> "Will drop category atom_site", no atoms loaded
ids unique,     no entities  -> Duplicate Key violation, cat: entity_poly_seq
ids duplicated, entities     -> "Will drop category atom_site", no atoms loaded
ids unique,     entities     -> 2 models, 400 residues each, real z-scores
```

Evaluation runs tortoize on `refined-patched.cif` while `scripts/patch_output_cif_files.py` fixes both properties before the validators see the file so this might not have come up...

## Change

- `renumber_atom_site_ids` in `utils/cif_utils.py` renumbers `_atom_site.id` once across all models. It touches only the `id` column; every other column and category is byte-identical, and clashscore is unchanged.
- `carry_polymer_entity_categories` in `utils/cif_utils.py` carries real `_entity`, `_entity_poly`, `_entity_poly_seq`, `_struct_asym` and deposited `_chem_comp` rows from the deposited entry, selecting the reference entity by sequence rather than by chain id so model chain relabelling does not matter. No synthetic entity stubs: when the deposit cannot be fetched, or no unambiguous alignment exists, the file is written with a warning and no entity categories.

- The alignment treats `_atom_site.label_seq_id` as authoritative and only shifts the deposited numbering to match it. Deposited residues may be missing from the model; only MSE/MET is treated as equivalent, and ambiguous mappings fail explicitly.

- `_write_structure_cif` in `utils/guidance_script_utils.py` applies both to `refined.cif` and to every trajectory CIF, with one deposited-entry load per save.

## Summary by CodeRabbit

- **New Features**
  - Structural CIF outputs now include sequential, unique atom identifiers and model metadata.
  - Polymer entity information, including sequences and chemical component details, is carried into generated CIF files when a matching deposited structure is available.
  - Reference structures can be resolved and cached automatically for output enrichment.
  - Trajectory and final structure exports support optional reference CIF data.

- **Bug Fixes**
  - Invalid or ambiguous polymer matches are detected without producing partially modified output.
  - CIF coordinates are still written when polymer metadata cannot be retrieved or matched.


src/sampleworks/utils/cif_utils.py
```diff
@@ -257,6 +258,194 @@ def resolve_mixed_hetatm_atom_altlocs(cif_path: Path | str) -> Path:
...
+    Parameters
+    ----------
+    cif_file : CIFFile
+        Single-block CIF file whose ``atom_site`` category is modified in place.
+    """
+    category = cif_file.block["atom_site"]
+    category["id"] = np.arange(1, category.row_count + 1)
+
+
+_POLYMER_ENTITY_CATEGORIES = ("entity", "entity_poly", "entity_poly_seq")
```
marcuscollins, line 277: Nit: move constants to the top of the file.

src/sampleworks/utils/cif_utils.py
```diff
@@ -257,6 +258,194 @@ def resolve_mixed_hetatm_atom_altlocs(cif_path: Path | str) -> Path:
...
+    ----------
+    name : str
+        Three-letter monomer name.
+
+    Returns
+    -------
+    str
+        Canonical monomer name, with selenomethionine normalized to methionine.
+    """
+    return "MET" if name == "MSE" else name
```
marcuscollins, line 317: What about CYS/CSO?

marcuscollins: Also, please look through our code, I think we have utility functions elsewhere that do this. As the package grows I'm trying to make sure we don't risk doing the same things in different ways in different places, so re-use should be high priority.

src/sampleworks/utils/cif_utils.py
```diff
@@ -257,6 +258,194 @@ def resolve_mixed_hetatm_atom_altlocs(cif_path: Path | str) -> Path:
...
+        """Keep minimum-cost paths and mark equally optimal alternatives.
+
+        Parameters
+        ----------
+        target : dict
+            Dynamic-programming states for the next reference residue.
+        key : tuple[int, bool]
+            Modeled position and whether the path is deleting reference residues.
+        cost : int
+            Number of deletion runs in the candidate path.
```
marcuscollins, line 357: deletion runs, or total deletions?

src/sampleworks/utils/cif_utils.py
```diff
@@ -257,6 +258,194 @@ def resolve_mixed_hetatm_atom_altlocs(cif_path: Path | str) -> Path:
...
+        Modeled monomer names.
+
+    Returns
+    -------
+    list[int] | None
+        Reference indices for the unique optimal embedding, or ``None`` when absent or ambiguous.
+    """
+    reference = [_canonical_monomer(name) for name in reference]
+    modeled = [_canonical_monomer(name) for name in modeled]
+    states: dict[tuple[int, bool], tuple[int, tuple[int, ...] | None, int]] = {
```
marcuscollins, line 337: Might be worth creating a type for this, to simplify and document what exactly this is.

src/sampleworks/utils/cif_utils.py
```diff
@@ -257,6 +258,194 @@ def resolve_mixed_hetatm_atom_altlocs(cif_path: Path | str) -> Path:
...
+        count : int
+            Number of equally optimal paths, capped at two.
+        """
+        current = target.get(key)
+        if current is None or cost < current[0]:
+            target[key] = (cost, path, count)
+        elif cost == current[0]:
+            target[key] = (cost, None, min(2, current[2] + count))
+
+    for reference_index, reference_name in enumerate(reference):
```
marcuscollins, line 369: It would help to have a comment explaining what exactly this is doing.

src/sampleworks/utils/cif_utils.py
marcuscollins: Yeah, look up _closest_canonical_amino_acid, it's in eval/structure_utils.py, we could move it somewhere more sensible, but I think that does what you want.

src/sampleworks/utils/cif_utils.py
```diff
@@ -257,6 +258,194 @@ def resolve_mixed_hetatm_atom_altlocs(cif_path: Path | str) -> Path:
...
+
+    Returns
+    -------
+    str
+        Canonical monomer name, with selenomethionine normalized to methionine.
+    """
+    return "MET" if name == "MSE" else name
+
+
+def _unique_subsequence_indices(reference: list[str], modeled: list[str]) -> list[int] | None:
```
marcuscollins, line 320: I don't see tests for this--it probably could use some, especially to help readers understand what it does, or should do.

src/sampleworks/utils/cif_utils.py
```diff
@@ -257,6 +258,194 @@ def resolve_mixed_hetatm_atom_altlocs(cif_path: Path | str) -> Path:
...
+        states = next_states
+
+    candidates = [value for (index, _), value in states.items() if index == len(modeled)]
+    if not candidates:
+        return None
+    minimum_cost = min(value[0] for value in candidates)
+    optimal = [value for value in candidates if value[0] == minimum_cost]
+    if sum(value[2] for value in optimal) != 1 or optimal[0][1] is None:
+        return None
+    return list(optimal[0][1])
```
marcuscollins, line 385: Not sure where to put this comment, but there's a weakness/edge case in this algorithm that I think we need to address. Suppose you have a run of three alanines in the reference sequence but only two appear in the modeled sequence. The position of the deletion is ambiguous with 3 possibilities--all have the same cost. That will yield a path `None`. There may be a small chance of this actually happening, so we should find out whether it ever happens in our examples, and at least when the path is set to None we should warn the user what may have happened and exactly where so that they can investigate. Otherwise this fails silently.

src/sampleworks/utils/cif_utils.py
```diff
@@ -257,6 +258,194 @@ def resolve_mixed_hetatm_atom_altlocs(cif_path: Path | str) -> Path:
...
+    sequence_ids = np.asarray(sequence["entity_id"].as_array(str))
+    sequence_names = np.asarray(sequence["mon_id"].as_array(str))
+    matches: list[tuple[str, list[int]]] = []
+    for entity_id in reference_block["entity_poly"]["entity_id"].as_array(str):
+        reference_names = list(sequence_names[sequence_ids == entity_id])
+        indices = _unique_subsequence_indices(reference_names, residue_names)
+        if indices is not None:
+            matches.append((str(entity_id), indices))
+    if len(matches) != 1:
+        raise ValueError(f"Expected one reference entity sequence match, found {len(matches)}")
```
marcuscollins, line 420: Would this happen if you had two sequences that were similar but not identical?

src/sampleworks/utils/cif_utils.py
```diff
@@ -340,3 +529,207 @@ def _normalize_nulls(value: Any) -> Any:
...
+
+    Raises
+    ------
+    ValueError
+        If no entity has polymer residues, or an entity has conflicting residue names.
+    """
+    atom_site = output_block["atom_site"]
+    entity_ids = atom_site["label_entity_id"].as_array(str)
+    chain_ids = atom_site["label_asym_id"].as_array(str)
+    sequence_ids = atom_site["label_seq_id"].as_array(str)
```
marcuscollins: Aren't these all ints? Is it just convenient to keep them as strings?

src/sampleworks/utils/cif_utils.py
```diff
@@ -340,3 +529,207 @@ def _normalize_nulls(value: Any) -> Any:
...
+        Row dictionaries sharing the same columns.
+
+    Returns
+    -------
+    dict[str, list[str]]
+        Concatenated category columns.
+    """
+    columns = {name: [] for name in rows[0]}
+    for row in rows:
+        for name, values in row.items():
```
marcuscollins, line 633: You should probably raise a KeyError if either there's an extra column name (which would be caught, but you could provide a more useful error message) OR if there's a missing column. Presumably we want to use explicit missing values (e.g. "?")

src/sampleworks/utils/cif_utils.py
```diff
@@ -340,3 +529,207 @@ def _normalize_nulls(value: Any) -> Any:
...
+
+
+def carry_polymer_entity_categories(
+    output: CIFFile,
+    reference: str | Path | CIFFile,
+) -> tuple[str, ...]:
+    """Carry consistent deposited polymer entity categories into an output CIF.
+
+    The output's residue sequence is matched to the deposited sequence for each
+    entity. Deposited numbering is shifted only when needed to match the output's
```
marcuscollins, line 645: Can we shift label_seq_id to match the original deposited numbering? The issue is that we often want to compare back to the original CIF input for metrics, and models like Protenix renumber the sequence starting at 1. It looks like in general this will make the entity ids, residue ids, etc... match the Sampleworks output, and I think we actually want the Sampleworks output to match in the input ids. LMK if you want to discuss.

marcuscollins: Maybe we can keep both numberings?

src/sampleworks/utils/cif_utils.py
```diff
@@ -340,3 +529,207 @@ def _normalize_nulls(value: Any) -> Any:
...
+        if "pdbx_strand_id" in entity_poly_row:
+            entity_poly_row["pdbx_strand_id"] = [",".join(sorted(chains))]
+        modeled_sequence = _one_letter_sequence(names)
+        for column in ("pdbx_seq_one_letter_code", "pdbx_seq_one_letter_code_can"):
+            if column in entity_poly_row:
+                entity_poly_row[column] = [modeled_sequence]
+        entity_poly_rows.append(entity_poly_row)
+
+        sequence_row = _select_category_rows(
+            reference_block["entity_poly_seq"], "entity_id", reference_id
```
marcuscollins, line 696: General Nit: I would make things like "entity_poly_seq" constants just to avoid misspellings. We might already have some of these CIF reserved names as constants somewhere in the package.

src/sampleworks/utils/guidance_script_utils.py
```diff
@@ -76,19 +84,77 @@
 from sampleworks.utils.torch_utils import try_gpu
 
 
+_RCSB_ID_PATTERN = re.compile(r"^(pdb_[A-Za-z0-9]{8}|[0-9][A-Za-z0-9]{3})(?:_|$)")
```
marcuscollins, line 87: This got used in your previous PR as well. Make sure to rebase, and make sure this constant is defined in just one location and re-used.

src/sampleworks/utils/guidance_script_utils.py
```diff
@@ -76,19 +84,77 @@
...
+_RCSB_CACHE = Path("~/.sampleworks/rcsb").expanduser()
+
+
+def _resolve_rcsb_id(protein: str) -> str:
```
marcuscollins, line 91: I think this method was also included in the previous PR, or something very similar. We don't want to duplicate functionality like this. FWIW I've noticed that agents like Claude are not respecting our AGENTS.md in this very well, and tend to recreate functionality at the drop of a hat, so something to watch out for.

tests/utils/test_cif_entity_categories.py
```diff
@@ -0,0 +1,284 @@
...
+from atomworks.io.utils.io_utils import load_any
+from biotite.structure import AtomArrayStack, stack
+from biotite.structure.io.pdbx import set_structure
+from biotite.structure.io.pdbx.cif import CIFCategory, CIFFile
+from sampleworks.utils.cif_utils import carry_polymer_entity_categories
+from sampleworks.utils.guidance_script_arguments import GuidanceConfig
+from sampleworks.utils.guidance_script_utils import save_everything
+
+
+def _multi_model_cif(structure_path: Path) -> CIFFile:
```
marcuscollins, line 17: This should be made a pytest fixture, which would just re-use it at test time, instead of reloading the CIF file each time it is called. Makes for faster testing.

tests/utils/test_cif_entity_categories.py
```diff
@@ -0,0 +1,284 @@
...
+    output.write(str(output_path))
+    reloaded = CIFFile.read(str(output_path)).block
+
+    assert categories == ("entity", "entity_poly", "entity_poly_seq", "struct_asym")
+    atom_site = reloaded["atom_site"]
+    assert set(atom_site["pdbx_PDB_model_num"].as_array(str)) == {"1", "2"}
+    assert set(atom_site["label_entity_id"].as_array(str)) == set(
+        reloaded["entity"]["id"].as_array(str)
+    )
+    assert set(reloaded["entity"]["type"].as_array(str)) == {"polymer"}
```
marcuscollins, line 60: Is it worth also checking that the sequences are correct?

tests/utils/test_cif_entity_categories.py
marcuscollins: You would need to specify the structure path, but it looks like you usually use the same one?

tests/utils/test_cif_entity_categories.py
```diff
@@ -0,0 +1,284 @@
...
+    )
+
+    with pytest.raises(ValueError, match="reference entity sequence match"):
+        carry_polymer_entity_categories(output, reference)
+
+    for category in ("entity", "entity_poly", "entity_poly_seq", "struct_asym"):
+        assert category not in output.block
+
+
+def test_carry_polymer_entity_categories_normalizes_selenomethionine(resources_dir):
```
marcuscollins, line 146: Should also check CSO as noted previously.

tests/utils/test_guidance_script_utils.py
```diff
@@ -93,6 +107,148 @@ def test_save_everything_uses_model_atom_array_for_mismatch(tmp_path: Path):
     )
 
     assert (tmp_path / "refined.cif").exists()
+    assert "without polymer entity categories" in caplog.text
+
+
+@pytest.mark.parametrize(
```
marcuscollins, line 113: See comment above: I think the method tested here is a duplicate of something in your previous PR, so these new tests shouldn't be needed if you remove this function from the present PR, and use the other one.

marcuscollins (commented):
Please have a look at these comments and let me know if you have questions. I think this looks basically fine but I want to double check some things and clean up just a bit.

PR #414, opened 2026-09-16
fix(utils): write output CIFs with the input's chains and numbering

rtviii:
Further replacing #301...

Output CIFs carry the generative model's chain ids and residue numbers rather than the input's. it is instead desirable that chain "B" in the original should match chain "B" in the final output etc, with the same numbering and sequence. 

## Change

- `AtomReconciler.model_to_struct` in `utils/atom_reconciler.py`,  plus a `mapped_struct_mask` property so callers cannot serialize an unmapped template coordinate by omission.
- `from_arrays` treats MSE `SE` and MET `SD` as the same atom for matching, and no longer reports an identity mapping when the two arrays agree on atom count and order but disagree on chain ids, residue numbers, residue names, atom names or elements.
- Both trajectory scalers put the canonical CPU reconciler and the filtered input atom array into the run metadata alongside the existing model atom array, for mismatch runs only.
- `save_everything` and both trajectory writers map every coordinate set through the reconciler before writing, so `refined.cif` and each trajectory CIF carry the input's identities.

Non-mismatch runs are unchanged: the reconciler is an identity mapping, no metadata is moved, and the save path is the one on main.

## Overlap with #413

This PR and #413 rewrite the same save path in `utils/guidance_script_utils.py` (`save_everything` and both trajectory writers), so whichever merges second needs a real rebase rather than a textual one. They compose: this PR decides which atoms and identities are written, #413 decides how the CIF is written (unique `_atom_site.id`, carried entity categories). After both, the prepared array from here should go through #413's writer. The `atom_id` regeneration here overlaps #413's `_atom_site.id` renumbering at a different layer; it is needed only because this branch starts from main, where nothing renumbers.

## Follow-ups

- Retiring `scripts/patch_output_cif_files.py` needs this PR plus #413; the script does renumbering and metadata grafting, and those are now both native.
- Chemical equivalence beyond MSE/MET (modified residues generally) is not handled.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved saving of trajectories and final structures when model and input atom layouts differ.
  - Preserved input atom identities, coordinates, numbering, and annotations during output generation.
  - Correctly handles model-only atoms and filtered structures.
  - Added support for matching equivalent methionine sulfur atom names.

- **Tests**
  - Added coverage for reconciliation, coordinate mapping, mismatch handling, and CPU/GPU sampling scenarios.


src/sampleworks/utils/atom_reconciler.py
```diff
@@ -60,6 +85,19 @@ class AtomReconciler:
     n_struct: int
     n_common: int
 
+    @property
+    def mapped_struct_mask(self) -> torch.Tensor:
```
DorisMai: doesn't seem used anywhere except in tests?

src/sampleworks/utils/guidance_script_utils.py
```diff
@@ -113,7 +133,126 @@ def _write_coords_into_array(
...
+            str(struct_atom_array.chain_id[index]),
+            int(struct_atom_array.res_id[index]),
+            str(struct_atom_array.ins_code[index]),
+        ): index
+        for index in mapped_struct_indices
+    }
+    arrays: list[AtomArray] = []
+    coord_chunks: list[torch.Tensor] = []
+    model_categories = set(model_atom_array.get_annotation_categories())
+    atom_annotations = ("atom_name", "element", "atomic_number", "charge", "b_factor", "occupancy")
```
DorisMai: should we use the input structure's occupancy? why not the 1/batch?

src/sampleworks/utils/guidance_script_utils.py
```diff
@@ -113,7 +133,126 @@ def _write_coords_into_array(
...
+        residue_map[key] = struct_index
+
+    extras_by_struct_residue: dict[tuple[str, int, str], list[int]] = {}
+    for model_index in np.flatnonzero(model_only_mask):
+        model_key = (
+            str(model_atom_array.chain_id[model_index]),
+            int(model_atom_array.res_id[model_index]),
+            str(model_atom_array.ins_code[model_index]),
+        )
+        struct_index = residue_map.get(model_key)
```
DorisMai: what if the structure atom_array has an entire residue missing that the model_atom_array predicts? it would just be silently dropped here? right now this problem is unlikely to happen because the sequence used by model to predict coordinates is inferred from the structure atom_array, though not impossible to happen -- e.g. a whole residue has occupancy=0 which would be filtered out from structure atom_array. This would be a bigger problem when PR #415 comes in where the model can explicitly predict missing residues.

src/sampleworks/utils/guidance_script_utils.py
```diff
@@ -113,7 +133,126 @@ def _write_coords_into_array(
     array_copy.coord = coords
 
 
-def _save_trajectory(trajectory, atom_array, output_dir, subdir_name, save_every):
+def _prepare_output_array(
```
DorisMai: this is a very long function, and I would strongly encourage trying to use index arrays over explicit for loops which can be both quite slow and have poor readability. If you really cannot avoid the for loops, I would appreciate a brief comment here and there explaining what's going on -- for example, it took me a while to keep track of the dict (e.g. residue_map, last_mapped_by_residue, extras_by_struct_residue), like is it mapping from model space or not, how many elements are there etc.

src/sampleworks/utils/guidance_script_utils.py
```diff
@@ -127,16 +266,22 @@ def _save_trajectory(trajectory, atom_array, output_dir, subdir_name, save_every
...
 
     for i, coords in enumerate(trajectory):
-        ensemble_size = coords.shape[0]
         if i % save_every != 0:
             continue
-        array_copy = atom_array.copy()
-        array_copy = stack([array_copy] * ensemble_size)
-        _write_coords_into_array(array_copy, coords.detach().numpy())
-        save_structure(str(output_dir / f"trajectory_{i}.cif"), array_copy)
+        output = _prepare_output_array(coords, atom_array, model_atom_array, reconciler)
```
DorisMai: over the trajectory, only the coordinates changed, the atom matching and renumbering between model space and structure atom array should be the same. might be better to refactor the preparation out of the for loop so the renumbering only needs to be figured out once.

tests/utils/test_save_reconciliation.py
```diff
@@ -0,0 +1,194 @@
...
+
+    paths_and_expected = [
+        (tmp_path / "refined.cif", 2.0),
+        (tmp_path / "trajectory" / "denoised" / "trajectory_0.cif", 4.0),
+        (tmp_path / "trajectory" / "next_step" / "trajectory_0.cif", 6.0),
+    ]
+    for path, expected_x in paths_and_expected:
+        output = _load_first_model(path)
+        ca = output[(output.chain_id == "B") & (output.res_id == 365) & (output.atom_name == "CA")]
+        np.testing.assert_allclose(ca.coord, [[expected_x, 0.0, 0.0]])
```
DorisMai: nitpick: again, a bit weaker test here, why not just assert on all coordinates

tests/utils/test_save_reconciliation.py
```diff
@@ -0,0 +1,194 @@
...
+    cif = CIFFile.read(str(output_path))
+    block = cif[next(iter(cif))]
+    atom_ids = block["atom_site"]["id"].as_array().tolist()
+    assert output.chain_id.tolist() == ["X", "X", "X", "Y", "Y"]
+    assert output.res_id.tolist() == [365, 365, 365, 20, 20]
+    assert set(output.atom_name[output.chain_id == "X"]) == {"N", "CA", "OXT"}
+    assert len(set(atom_ids)) == len(atom_ids) == len(output)
+    x_ca = output[(output.chain_id == "X") & (output.atom_name == "CA")]
+    np.testing.assert_allclose(x_ca.coord, [[2.0, 0.0, 0.0]])
+    np.testing.assert_allclose(output.coord[output.atom_name == "OXT"], [[3.0, 0.0, 0.0]])
```
DorisMai: nitpick: stronger test should check all coords rather than just pick one or two.

tests/utils/test_save_reconciliation.py
```diff
@@ -0,0 +1,194 @@
...
+    )
+
+    output_path = tmp_path / "refined.cif"
+    output = _load_first_model(output_path)
+    cif = CIFFile.read(str(output_path))
+    block = cif[next(iter(cif))]
+    atom_ids = block["atom_site"]["id"].as_array().tolist()
+    assert output.chain_id.tolist() == ["X", "X", "X", "Y", "Y"]
+    assert output.res_id.tolist() == [365, 365, 365, 20, 20]
+    assert set(output.atom_name[output.chain_id == "X"]) == {"N", "CA", "OXT"}
```
DorisMai: I would ideally want to see the ordering of atoms is expected, that is a model-only atom is inserted to the residue right after the matched atoms, so you never break a residue into non-contiguous entries of atoms. in this example you have it's a bit hard to test that since you only have one residue per chain, so the wrong behavior of simply appending model only atoms at the end of chain would still pass.

DorisMai (changes requested):
My main / blocking comment is regarding the `_prepare_output_array` function, including how we handle missing entire residues, which atom array's annotation to use. 

I am also not sure if we should always write out MSE if the input sequence has MSE. Some predictor (like boltz) just doesn't handle MSE, and so the forward model (e.g. the scattering factor that depends on the element type of the atom) in reward/guidance uses SD. If the ensemble is guided by reward that assumes SD, should we still write out MSE in the output refined.cif?

